PR #4, opened 2019-10-28
Fix alias command on readme

joaogui1:
Add quotes to the alias command on readme, withuot them most shells would simply alias stack to dex, and not the full command.


dougalm:
No longer needed (made the change earlier)

PR #5, opened 2019-10-28
Adds response to C benchmark

joaogui1:
Before the C version wouldn't print anything about how long it took to run, now it's easier to run the benchmarks as it outputs the time taken.


dougalm:
@googlebot rescan

PR #9, opened 2019-10-30
Travis CI configuration

peterbecich:
Seems to work: https://travis-ci.org/peterbecich/dex-lang/builds/604830911

Anything else to cover than `make` and `make tests`?


dougalm:
Great idea. Thanks for putting this together. It looks like there are some hoops I need to jump through to get Travis set up on our end. Shouldn't take more than a few days.

PR #10, opened 2019-11-01
Fix Travis CI caching

peterbecich:
Wrong environment variable was used.  Caches `.stack-work` correctly now.


dougalm:
Excellent! That should speed things up. Thank you.

Issue #8, opened 2019-10-28
Unable to print transposed matrix when elements are not constant?


jyh1:
The following example seems to work correctly:
```
mat1 = for i::4 j::4. 0
mat2.i.j = mat1.j.i
:p mat2
> [[0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0]]
```

However, weird error happened when the matrix contains different elements:
```
mat3 = for i::4 j::4. asint i
mat4.i.j = mat3.j.i
:p mat4
Compiler bug!Lookup failed:i
context:
[4=>4=>Int]
(tab::4=>Int = (for  j::4 . 
    mat3.j.i);
(tab_1::4=>4=>Int = (for  i::4 . 
    tab);
tab_1))
```

The type seems to be correctly inferred:
```
:t mat4
(4=>(4=>Int))
```

dougalm:
Yep, that's a bug! Thanks for the report. These kinds of bugs can be a bit tricky to track down so it might have to wait till later in the week.

dougalm:
Btw, to explain the error message a bit: that's a pretty-printing of the normalized IR which failed type checking because `i` in `mat3.j.i` is unbound. You can inspect the different IRs produced with `:typed`, `:normalized`, `:simp`, `:imp`, `:llvm`, and `:asm`.

jyh1:
Not sure if this warrants a separate issue, another unbounded variable bug:

```
mat = for i::4 j::4 . asint i
:p for i. mat.i.i
[0, 1, 2, 3]

:p (tmp = for i. mat.i.i; tmp)
Compiler bug!Lookup failed:i
context:
[4=>Int]
mat.i
```
Feel like this is because the eta-reduction of AtomaticFor [here](https://github.com/google-research/dex-lang/blob/92a916859befc746fa050e55fb71b733d04d21ea/src/lib/Normalize.hs#L303) might be problematic... Should make sure `i` is not referenced in `e` before carrying out the reduction?

dougalm:
Thanks for tracking that down! Even if it isn't the cause of this bug, you're right that the eta reduction is broken. Feel free to submit a PR. Otherwise I'll get to it soon.

jyh1:
I would love to contribute! What are you thinking of the fix? Obviously we can simply turn off the eta reduction for now. To make it right requires information about free variables. I am still getting familiar with the codebase though I think currently there is no such information in the simplify pass?

[...]

PR #12, opened 2019-11-01
Fix eta-reduction for NAtomicFor

jyh1:
Fix the broken eta-reduction mentioned [here](https://github.com/google-research/dex-lang/issues/8#issuecomment-547697495).

This will incur additional traverses of the AST, however [similar operations](https://github.com/google-research/dex-lang/blob/92a916859befc746fa050e55fb71b733d04d21ea/src/lib/Normalize.hs#L589) seems to have already been used in the pass.

### DIFF
--- a/examples/eval-tests.dx
+++ b/examples/eval-tests.dx
@@ -188,3 +188,6 @@ litArr = [10, 5, 3]
 
 :p fold (for i::3. 0.0) $ for i::2. lam c. (for j. c.j + real (asint j))
 > [0.0, 2.0, 4.0]
+
+:p (mat = for i::4 j::4 . asint i; tmp = for i. mat.i.i; tmp)
+> [0, 1, 2, 3]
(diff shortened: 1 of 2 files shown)


jyh1:
@googlebot I signed it!

dougalm:
Very nice! Sorry I've been slow to respond to the bug. Thanks for the fix, and great work figuring out how to query free vars without any help from me.

Do you mind adding your two failing examples as test cases to the [end of one of the quine tests](https://github.com/google-research/dex-lang/blob/master/examples/eval-tests.dx)? Then I'll be very happy to merge.

Don't worry about the extra AST traversals. Eventually it'd be nice to guarantee linear compile times, but it'll probably require making things more complicated and hard to change. We can focus on simplicity and correctness for now.

jyh1:
Thanks! I only added the second failing example that this patch fixed. The first one seems to be not caused by eta-reduction.

dougalm:
Actually, it looks like you might have to rename `mat` to `mat'` or something because it's shadowed in the test file (if you run `make tests` or `make run-eval-tests` it should show up). We're probably being overly strict about shadowing -- it would only be a warning in Haskell/GHC.

jyh1:
Got it. Seems `mat2` works..

dougalm:
Merged!

PR #14, opened 2019-11-03
Fix materializeAtom in simplify pass

jyh1:
The reason for issue #8 might be due to the broken `materializeAtom` function. Currently, it basically eliminates the `NAtomFor` by creating a declaration of `NScan` and returning the bound variable. This is problematic in that it makes the `body` expression of `NAtomFor` out of current context during recursive calls. E.g. `afor i. afor j. x.j.i` becomes `tab = for j. x.j.i; tab_1 = for i. tab; tab_1`, i.e. `for j. x.j.i` is taken outside of context of index variable `i`.

This patch fixes this issue by trying to maintain the nested `NAtomicFor` structure in the generated declaration. It introduces a helper function to recursively transform the atom expression to an `NExpr` and then creates a single declaration for it.

### DIFF
--- a/src/lib/Normalize.hs
+++ b/src/lib/Normalize.hs
@@ -319,17 +319,23 @@ materializeAtoms xs = do
 
 materializeAtom :: NAtom -> SimplifyM ((NAtom, [NDecl]), SimpEnv)
 materializeAtom atom = case atom of
-  NAtomicFor b body -> do
+  NAtomicFor _ _ -> do
     scope <- asks snd
     let ty = getAtomType scope atom
-    refreshBindersRSimp [b] $ \[b'] -> do
-      ((body', decls), env) <- materializeAtom body
-      scope' <- asks snd
-      let v = rename (rawName "tab") (scope' <> snd env)
-      let env' = (mempty, v @> L ty)
-      let decl = NLet [v:>ty] (NScan b' [] [] (NAtoms [body']))
-      return ((NVar v, decls ++ [decl]), env <> env')
+    atomExpr <- atomToNExpr atom
+    let v = rename (rawName "tab") scope
+    let env' = (mempty, v @> L ty)
+    let decl = NLet [v:>ty] atomExpr
+    return ((NVar v, [decl]), env')
   _ -> return ((atom, []), mempty)
+  where
+    atomToNExpr :: NAtom -> SimplifyM NExpr
+    atomToNExpr atom = case atom of
+      NAtomicFor b body -> 
+        refreshBindersRSimp [b] $ \[b'] -> do
+          bodyExpr <- atomToNExpr body
+          return (NScan b' [] [] bodyExpr)
+      _ -> return (NAtoms [atom])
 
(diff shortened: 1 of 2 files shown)


dougalm:
Wow, very nice work finding the bug! Your analysis is exactly right. The problem is that the decls need to remain in the scope of the binder `b'`. Your fix works, but it might be simpler to just add the decls into the body of the new scan. Basically this patch:

    -      let decl = NLet [v:>ty] (NScan b' [] [] (NAtoms [body']))
    -      return ((NVar v, decls ++ [decl]), env <> env')
    +      let decl = NLet [v:>ty] (NScan b' [] [] (wrapNDecls decls (NAtoms [body'])))
    +      return ((NVar v, [decl]), env <> env')

How does that look to you?

dougalm:
Actually, I take that back. Your solution has the benefit that it creates fewer unnecessary let bindings. That might be worth the extra lines of code. I'll happily merge either version. I'll leave it up to you.

src/lib/Normalize.hs
```diff
@@ -319,17 +319,23 @@ materializeAtoms xs = do
...
+    atomExpr <- atomToNExpr atom
+    let v = rename (rawName "tab") scope
+    let env' = (mempty, v @> L ty)
+    let decl = NLet [v:>ty] atomExpr
+    return ((NVar v, [decl]), env')
   _ -> return ((atom, []), mempty)
+  where
+    atomToNExpr :: NAtom -> SimplifyM NExpr
+    atomToNExpr atom = case atom of
+      NAtomicFor b body ->
```
dougalm: Trailing space (maybe configure your editor to show those?).

src/lib/Normalize.hs
```diff
@@ -319,17 +319,23 @@ materializeAtoms xs = do
...
-      return ((NVar v, decls ++ [decl]), env <> env')
+    atomExpr <- atomToNExpr atom
+    let v = rename (rawName "tab") scope
+    let env' = (mempty, v @> L ty)
+    let decl = NLet [v:>ty] atomExpr
+    return ((NVar v, [decl]), env')
   _ -> return ((atom, []), mempty)
+  where
+    atomToNExpr :: NAtom -> SimplifyM NExpr
+    atomToNExpr atom = case atom of
```
dougalm: GHC warns that `atom` here shadows the binding above. You could rename to `atom'` or move this function out to the top level. It's nice to try to keep the compilation warning-free because the warnings often catch real bugs.

jyh1: Yes, always better to keep warning free. I moved it to top level.

jyh1:
> Actually, I take that back. Your solution has the benefit that it creates fewer unnecessary let bindings. That might be worth the extra lines of code. I'll happily merge either version. I'll leave it up to you.

I think it's better to keep the current version for the reason you mentioned, it is supposed to perform inlining after all:). Besides I feel like the behavior of the function is more clear in the current patch.

PR #18, opened 2019-11-23
Make prompt green

oxinabox:
A very small thing,
but I find it is a surprising enhancement to usability.

image shows green prompt

It also makes me feel like we have made some progress as a society since the 80s.

### DIFF
--- a/src/dex.hs
+++ b/src/dex.hs
@@ -94,7 +94,8 @@ evalPrelude fname pass = do
 
 replLoop :: Monoid env => FullPass env-> InputT (StateT env IO) ()
 replLoop pass = do
-  sourceBlock <- readMultiline ">=> " parseTopDeclRepl
+  -- ANSI escape codes make prompt green
+  sourceBlock <- readMultiline "\ESC[32m\STX>=> \ESC[m\STX" parseTopDeclRepl
   result <- lift $ evalDecl pass sourceBlock
   liftIO $ putStrLn $ pprint result
 


dougalm:
I agree that a colored prompt can help visually parse a sequence of commands. I don't want to hard-code the color because it's not a great choice for a white-background terminal (which I use) and it would be invisible on a green-background one. So I plumbed it through in 4f6411aec6 as a command line argument, `--prompt`, to `dex repl`. Quoting the ANSI escape sequences is a little tricky. The following works for me in zsh and bash.

    dexprompt=$(printf '\x1B[32m\x02>=> \x1B[m\x02')
    alias dex-repl='stack exec dex -- repl --prompt "$dexprompt"'

Speaking of ergonomics, have you tried out the notebook mode yet?

oxinabox:
> Speaking of ergonomics, have you tried out the notebook mode yet?

Not yet, I have been repeatedly rebuild dex.
Trying (and failing) to find a way to make it nice.

I think I'll leave that to someone who knows build systems and linkers better than me now 
though.

oxinabox:
>  Speaking of ergonomics, have you tried out the notebook mode yet?

Oh, that is _different_
Took me a couple of minutes to workout how to use it.
That It wasn't like JuPyTeR,
but was actually more like a LaTeX editor with preview on save.

This might be a solid idea at having literate programming, without having the downsides around state/order and doing terrible things to version control.
I guess there is some weave/rmarkdown thing that is similar?

dougalm:
Glad you like it! It's a toy for now but I'd like to develop it further.

I've always struggled with jupyter-style notebooks because of the state problem, which you point out. You have to mentally track the state of the interpreter to choose which pieces to evaluate and it's easy to get tangled up.

Of course, you *want* some state, because you don't want to be recomputing everything from scratch all the time. But I think it's cleaner to just treat it as a caching problem, where semantically everything is pure and stateless, with state secretly used in the implementation as a (very effective) optimization.

In the current notebook setup every top-level declaration runs in its own logical thread, which means that in principle you also get concurrency (and maybe even parallelism) for free. So you can have part of your program churning on an expensive calculation and still be editing other parts of the file and seeing results.

PR #21, opened 2019-11-23
Remove color flag from diff as it breaks mac

oxinabox:
Annoyingly there is no `diff --color` for mac.
(What a terrible world).

I assume diff works fine without it.


dougalm:
Didn't know --color wasn't supported on Mac. Removing it is totally fine by me.

PR #24, opened 2019-11-27
Starting to implement property based testing

jyh1:


dougalm:
I'm very excited about this! Looking forward to the avalanche of bugs we'll uncover.

tests/TestPass.hs
```diff
@@ -0,0 +1,68 @@
...
+            return res
+    return eval
+
+
+loadTypeEnv :: String -> IO [(SigmaType, Name)]
+loadTypeEnv fname =
+    case sourcePass >+> deShadowPass >+> typePass of
+        TopPass pass -> do
+            envs <- loadFile fname pass
+            let env = (snd (unsafeCoerce envs)) :: TypeEnv
```
jyh1: I am trying to pre-load the definitions from Prelude to the generator environment so it can make use of them in random generation, thus obviating the need to write generators for primitives. The current approach is to evaluate the prelude up to the `typePass` and then extract the [type environment](https://github.com/google-research/dex-lang/blob/674c4df6b212b83b74fdc34885c5b4f8b8d7720b/src/lib/Type.hs#L28). However it is tricky to do the extraction (perhaps need dependent type?). I resorted to the `unsafeCoerce` which is obviously not very satisfying. Is there a better way?

dougalm: We could make versions of the passes that expose their environments rather than hide them. Or we could extract types directly from the parsed program, since all of the names in the prelude have explicit type annotations. But we can do one of those things later. For now, I'm happy with `unsafeCoerce` used here for generating test examples.

PR #29, opened 2019-12-05
Update table constructor generator to respect Data constraint

jyh1:


dougalm:
Looks great. Thank you!

PR #49, opened 2019-12-28
Add compare equal operators <= and >=

oxinabox:
It surprised me that `1 <= 2` etc did not work.

This seemed a good chance to dip my toes into the Haskell some.


dougalm:
Nice! Thank you.

PR #64, opened 2020-04-10
Generalize AsIdx to arbitrary index set types

apaszke:
Previously the `AsIdx` constructor has taken an `Int` that was implicitly
interpreter as the upper bound for the `IdxSetLit` type. This patch
makes the index set argument explicit, allowing generalization of
`AsIdx` to e.g. `Range`.

### DIFF
--- a/src/lib/Imp.hs
+++ b/src/lib/Imp.hs
@@ -94,7 +94,7 @@ toImpCExpr dests op = case op of
     n' <- typeToSize (varAnn b)
     emitLoop n' $ \i -> do
       let ithDest = tabGetIAtom dests i
-      extendR (b @> asIdx n' i) $ toImpExpr ithDest body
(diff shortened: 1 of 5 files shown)


src/lib/Imp.hs
```diff
@@ -94,7 +94,7 @@ toImpCExpr dests op = case op of
     n' <- typeToSize (varAnn b)
     emitLoop n' $ \i -> do
       let ithDest = tabGetIAtom dests i
-      extendR (b @> asIdx n' i) $ toImpExpr ithDest body
+      extendR (b @> asIdx (varAnn b) i) $ toImpExpr ithDest body
```
apaszke, line 97: **Important note:** This assumes that the `AsIdx` constructor always expects expressions in the range [0, n) where n is the size of the index space (even though the real index space might be e.g. `Range 5 10`).

PR #68, opened 2020-04-21
Remove IdxSetLit in favor of IntRange

apaszke:


src/lib/Type.hs
```diff
@@ -636,8 +634,8 @@ flattenType :: Type -> [(BaseType, [Int])]
 flattenType ty = case ty of
   BaseType b  -> [(b, [])]
   RecType r -> concat $ map flattenType $ toList r
-  TabType (IdxSetLit n) a -> [(b, n:shape) | (b, shape) <- flattenType a]
-  IdxSetLit _ -> [(IntType, [])]
+  TabType (FixedIntRange low high) a -> [(b, (high - low):shape) | (b, shape) <- flattenType a]
+  FixedIntRange _ _ -> [(IntType, [])]
```
apaszke, line 638: I'm not sure if this is correct, so please check this carefully. IIUC `flattenType` is meant to give you something that can hold a runtime representation of a value of that type, and for indices (no matter what's the index set) it's just a single Int. If we were interested in serializing the type itself, then we would need at least two integers, but I don't think that this is what's happening here. Are the types saved in plain text?

src/lib/Imp.hs
```diff
@@ -275,7 +274,7 @@ lookupVar v = do
     Just v' -> v'
 
 indexSetSize :: Type -> ImpM IExpr
-indexSetSize (IdxSetLit n) = return $ ILit (IntLit n)
+indexSetSize (FixedIntRange low high) = return $ ILit $ IntLit $ high - low
```
apaszke, line 277: What's interesting is that this case seems unnecessary, but in fact programs as simple as `[1, 2, 3]` fail to compile without it! This is because if it's skipped, then each time we query this function (e.g. to generate the shape of an array), we allocate a fresh variable and use that in the shape expression. This causes some type mismatches later, because even though those variables originate from the same expression, we use different variables and our type-level equality is purely syntactic. To better illustrate this, lowering a program like `[1, 2, 3]` without this extra rule produces this:
```
> Compiler bug!
>
> Type mismatch in copy: (Int, [v_2]) != (Int, [v])(, copy v_1 v_3)
> === context ===
> Checking after imp pass:
> imports: ()
> [v_1:Ptr (Int[v])]
> v:Int = %isub 3 0
> v_2:Int = %isub 3 0
> v_3:Ptr (Int[v_2]) = alloc (Int, [v_2])
> v_4:Ptr (Int[]) = v_3.0
> store v_4 1
> v_5:Ptr (Int[]) = v_3.1
> store v_5 2
> v_6:Ptr (Int[]) = v_3.2
> store v_6 3
> copy v_1 v_3
> free v_3
> (*ans* @> L (afor *:3 . aget v_1))
> exports: (*ans* @> L (3=>Int))
```

Note that what I did in here is not a robust fix, and the same problem will bite us in the future when we will be interested in compiling truly dependent ranges, because then we have no choice other than doing the subtraction at runtime. One way to fix this that I thought about would be to memoize answers of `indexSetSize` inside `ImpM` so that each type will get assigned a unique size variable that will be used throughout the lowered program. WDYT?

apaszke: Turns out that another way to solve this problem is to add a simplification case to `impSub`. I'll do that in a followup PR that I'm preparing now.

src/lib/Syntax.hs
```diff
@@ -75,6 +75,9 @@ data Type = TypeVar TVar
           | NoAnn
             deriving (Show, Eq, Generic)
 
+pattern FixedIntRange :: Int -> Int -> Type
+pattern FixedIntRange low high = IntRange (DepLit low) (DepLit high)
+
```
dougalm, line 80: I didn't know about this extension! Very cool. There are lots of places that could use it.

src/lib/Type.hs
dougalm: Yes, flattening is used in the binary serialization format, and the (unflattened) type itself is stored explicitly, using its concrete syntax. (e.g. you can see that we parse it [here](https://github.com/google-research/dex-lang/blob/master/src/lib/Serialize.hs#L109)). So it's fine that we lose the type information when we flatten.

src/lib/Imp.hs
dougalm: That sounds like a good fix, at least until we think about it more. Basic constant folding seems like a very reasonable thing to do short of full canonicalization.

PR #70, opened 2020-04-21
Allow index ranges to skip one of their bounds

apaszke:
One thing we might want to reconsider is the syntax for skipped bounds, because right now this code is valid:
```
for i:4. for j:i.... 1.0
```
and it says that the inner loop goes from `i` to the upper bound. The first three dots after `i` are the index range operator, while the fourth one terminates the `for`. An alternative would be to e.g. say that bounds can be replaced with a `*` giving:
```
for i:4. for j:i...*. 1.0
```

Of course it is legal to put a space between `...` and `.`, but it's optional and might make the code a little weird (especially that we use 4-character range operators for `Int` restrictions). On the other hand the current syntax is quite nice, as it will allow us to write the type of lower-triangular arrays as `(i:n)=>(...i)=>Real` in the future.

### DIFF
--- a/src/lib/Syntax.hs
+++ b/src/lib/Syntax.hs
@@ -33,7 +33,8 @@ module Syntax (
     strToName, nameToStr, unzipExpr, declAsModule, exprAsModule, lbind, tbind,
     noEffect, isPure, EffectName (..), EffectRow, Vars,
     traverseType, monMapSingle, monMapLookup, PiType (..), makePi, applyPi,
-    isDependentType, newEnv, newLEnv, newTEnv, pattern FixedIntRange)
+    isDependentType, newEnv, newLEnv, newTEnv, pattern FixedIntRange,
+    indexRangeBound)
   where
 
 import Data.Tuple (swap)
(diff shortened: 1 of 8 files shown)


src/lib/Syntax.hs
```diff
@@ -745,7 +751,10 @@ traverseType :: Applicative m => (Kind -> Type -> m Type) -> Type -> m Type
 traverseType f ty = case ty of
   BaseType _           -> pure ty
   IntRange a b         -> liftA2 IntRange (f depIntType a) (f depIntType b)
-  IndexRange a ac b bc -> liftA4 IndexRange (f depIntType a) (pure ac) (f depIntType b) (pure bc)
+  -- TODO: We really shouldn't be passing depIntType into f in this case
+  IndexRange a b       -> liftA2 IndexRange
+    (traverse (\(t, i) -> liftA2 (,) (f depIntType t) (pure i)) a)
+    (traverse (\(t, i) -> liftA2 (,) (f depIntType t) (pure i)) b)
```
apaszke, line 757: This unfortunately ended up looking pretty awful, but it's the first time I've even attempted to write a traversal (I still don't have a good mental model of them), so if you have any ideas as to how to make this code nicer then please do let me know!

dougalm: Yes, these applicative traversals often end up looking ugly. If we really wanted we could try "applicative do" or even just specialize to monads (since we only ever traverse with monads). But this looks fine to me.

PR #74, opened 2020-05-05
Fix freeVars to ignore DeBruijn indices

apaszke:
DeBruijn names cannot be free by definition. It would be a bit nicer to
discard them at the binders, but this approach is significantly simpler.

This also incldues a minor cleanup in how `Var`s are handled in
freeVars. The previous instance was binder-specific, so I've removed it
and replaced it with explicit functions that can be used depending on
the semantics of the call site.

### DIFF
--- a/src/lib/Syntax.hs
+++ b/src/lib/Syntax.hs
@@ -527,7 +527,7 @@ class HasVars a where
 instance HasVars FExpr where
   freeVars expr = case expr of
     FDecl decl body -> freeVars decl <> (freeVars body `envDiff` fDeclBoundVars decl)
-    FVar v@(_:>ty) -> v@>L ty <> freeVars ty
+    FVar v       -> freeVarVars v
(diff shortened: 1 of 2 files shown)


src/lib/Syntax.hs
```diff
@@ -555,28 +555,30 @@ instance HasVars Type where
...
   freeVars (Pi a b) = freeVars a <> freeVars b
 
-instance HasVars b => HasVars (VarP b) where
-  freeVars (_ :> b) = freeVars b
+-- NOTE: We don't have an instance for VarP, because it's used to represent
+--       both binders and regular variables, but each requires different treatment
+freeBinderVars :: Var -> Vars
+freeBinderVars (_ :> t) = freeVars t
+
+freeVarVars :: Var -> Vars
```
apaszke, line 572: This name is a bit awkward, so I'm open to suggestions if you have any

PR #75, opened 2020-05-08
Add a builtin Either type and case expression

apaszke:
This includes a few tests for type inference, but note that the
functions cannot yet be used anywhere, because the normalization and
lowering is not implemented properly just yet.


dougalm:
Beautiful!

PR #78, opened 2020-05-12
Desugar sum types into product types

apaszke:
`Either a b` becomes `(Bool, a, b)`, where the bool tag stores the
used constructor. The value of type that doesn't correspond to the right
case is left unspecified (arbitrary value will be selected during
compilation). `SumCase` operators are desugared into applications of
both branches and then a `Select` operator.

Note that this approach has the benefit of being very simple, but it has
some pretty significant downsides:

- A type obtained by taking an n-time sum of a type `T` is supposed to
  have `|T|*n` inhabitants, not `|T|^n * 2`.
- We keep doing unnecessary compute on the unspecified value only to
  keep throwing it away.
- Cases might generate hardware traps or aborts, which we would have to
  mask to obtain desired semantics.
- It's impossible to always generate a value of a type that is stored in
  one `Either` constructor, because it might not have any inhabitants.
  `0....0` (i.e. `IntRange 0 0`) is a good and simple example.

### DIFF
--- a/src/lib/Normalize.hs
+++ b/src/lib/Normalize.hs
@@ -12,6 +12,7 @@ module Normalize (normalizeModule) where
 import Control.Monad
 import Control.Monad.Reader
 import Data.Foldable
+import Data.Bifunctor
 
 import Env
 import Syntax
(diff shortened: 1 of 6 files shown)


src/lib/Normalize.hs
```diff
@@ -96,6 +125,20 @@ normalizeTLam (FTLam tvs qs body) =
...
+desugarType :: Type -> Type
+desugarType (SumTy l r) = RecTy $ Tup $ [l, r, TC $ BaseType BoolType]
+desugarType tv@(TypeVar _) = tv
+desugarType (ArrowType l (Pi a (eff, b))) = ArrowType l $ Pi (desugarType a) (eff, desugarType b)
+desugarType (TabType (Pi a b)) = TabType $ Pi (desugarType a) (desugarType b)
+desugarType (Forall vs qs t)   = Forall vs qs $ desugarType t
+desugarType (TypeAlias vs t)   = TypeAlias vs $ desugarType t
+desugarType (TC con)           = TC $ fmapTyCon con desugarType id
+desugarType eff@(Effect _ _)   = eff
+desugarType NoAnn = NoAnn
```
apaszke, line 137: I really don't like the fact that it's a poor man's traversal, but making `Type` a functor is annoying, because you'd need to take a fix-point and that introduces more constructors to unpack 🤮

dougalm: Without changing `Type`, we could make a type-blind traversal like `traverseType :: Applicative f => Type -> (Type -> f Type) -> f Type`. But this looks fine to me. Most of the cases are still handled by `fmapTyCon`.

PR #79, opened 2020-05-12
Make sum types valid index sets

apaszke:
Also, make `Bool` an index set. Now, the following runs correctly:

```haskell
type Weights n m = n=>m=>Real

type Biases n = n=>Real

type Params n m = (Either (n,m) n)=>Real

flatten : (Weights n m, Biases n) -> Params n m
flatten (w, b) = for idx. case idx
                Left (i,j) -> w.i.j
                Right i    -> b.i

w = for i:2. for j:2. (real (asint i)) * 2.0 + (real (asint j))
b = for j:2. neg (real (asint j))

restructure : Params n m -> (Weights n m, Biases n)
restructure params = ( for i. for j. params.(%left((i,j)))
                     , for i.        params.(%right(i))    )

(w, b)
restructure (flatten (w, b))
```

with the last two expressions returning the same values!

### DIFF
--- a/src/lib/Normalize.hs
+++ b/src/lib/Normalize.hs
@@ -45,7 +45,7 @@ normalize expr = case expr of
     extendR env $ normalize body
   FVar v -> lookupVar v
   FPrimExpr (OpExpr (SumCase sumVal lBody rBody)) -> do
-    ~(Tup [lInput, rInput, side]) <- unpackRec =<< normalize sumVal
+    ~(Tup [side, lInput, rInput]) <- unpackRec =<< normalize sumVal
     lOutput <- app lBody lInput
     rOutput <- app rBody rInput
     emit $ Select (getType lOutput) side lOutput rOutput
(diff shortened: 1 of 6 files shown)


src/lib/Normalize.hs
```diff
@@ -42,19 +44,48 @@ normalize expr = case expr of
...
+arbitraryValue (TabTy a b)                = Con $ AFor a $ arbitraryValue b
+arbitraryValue (TC (BaseType RealType))   = Con $ Lit $ RealLit 1.0
+arbitraryValue (TC (BaseType IntType))    = Con $ Lit $ IntLit 1
+arbitraryValue (TC (BaseType BoolType))   = Con $ Lit $ BoolLit False
+arbitraryValue (TC (BaseType StrType))    = Con $ Lit $ StrLit ""
+arbitraryValue (TC (RecType r))           = Con $ RecCon $ fmap arbitraryValue r
+-- XXX: This might not be strictly legal, because those types might have no inhabitants!
+arbitraryValue t@(TC (IndexRange _ _ _))  = Con $ AsIdx t $ Con $ Lit $ IntLit 0
+arbitraryValue t@(TC (IntRange _ _))      = Con $ AsIdx t $ Con $ Lit $ IntLit 0
+arbitraryValue t                          = Con $ AsIdx t $ Con $ Lit $ IntLit 0
```
apaszke, line 79: This is a hack, because inside `restructure` we have constructors like `Left @n (i, j) : Either (n, m) n`, and the only thing we know about `n` is that it's an index set. In this case it is valid to generate an expression like this one (of course under the assumption that the index set will end up being non-empty, but that's also a problem in the previous PR...), instead of failing with an error. I'm not sure what's the best way to get my hands on the class environment at this point. Any hints?

PR #80, opened 2020-05-13
Implement a better lowering for sum types

apaszke:
Instead of normalizing them out while translating to the core IR,
they're kept there until the Imp backend, which allows us to e.g. shrink
the index sets they represent (by excluding the inhabitants of
`(Bool, l, r)` which are inaccessible). In the end any values are still
lowered to products though, with case statements desugared into masked
function application in the simplification pass.

Also, add a proper parser for Left/Right so that we can avoid the
awkward builtin syntax.

### DIFF
--- a/src/lib/Imp.hs
+++ b/src/lib/Imp.hs
@@ -87,10 +87,18 @@ toImpCExpr' op = case op of
       copyAtom ithDest ans
     return dest
   TabGet x i -> impTabGet x i
+  SumGet x getLeft ->
+    case x of
+      SumVal _ l r -> return $ if getLeft then l else r
+      val -> error $ "Expected a sum type, got: " ++ pprint val
+  SumTag x ->
+    case x of
+      SumVal t _ _ -> return t
+      val -> error $ "Expected a sum type, got: " ++ pprint val
   RecGet x i -> do
     case x of
       RecVal r -> return $ recGet r i
-      val -> error $ "Expected a record, got: " ++ show val
+      val -> error $ "Expected a record, got: " ++ pprint val
   RunReader r (LamExpr ref body, env) -> do
     toImpExpr (env <> ref @> L r) body
   RunWriter (LamExpr ref body, env) -> do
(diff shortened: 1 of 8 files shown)


src/lib/Imp.hs
```diff
@@ -212,40 +234,60 @@ intToIndex ty@(TC con) i = case con of
...
-      f :: (Atom, IExpr) -> IExpr -> ImpM IExpr
-      f (i, stride) cumIdx = do
-        i' <- indexToInt i
-        iDelta  <- impMul i' stride
-        impAdd cumIdx iDelta
-  _ -> fromScalarAtom idx
-
-getStrides :: Traversable f => f (a, Type) -> ImpM (f (a, IExpr))
+indexToInt :: Type -> Atom -> ImpM IExpr
+indexToInt ty idx = case ty of
```
apaszke, line 255: I have refactored this function to also get the type as an argument, because it's always readily available at the call sites and otherwise we would have to retrieve it ourselves in here...

src/lib/PPrint.hs
```diff
@@ -168,21 +168,31 @@ instance (Pretty ty, Pretty e, PrettyLam lam) => Pretty (PrimOp ty e lam) where
...
+prettyPrimCon (Lam _ _ lam) = parens $ prettyL lam
+prettyPrimCon (RecCon r)    = p r
+prettyPrimCon (AFor n body) = parens $ "afor *:" <> p n <+> "." <+> p body
+prettyPrimCon (AGet e)      = "aget" <+> p e
+prettyPrimCon (AsIdx n i)   = p i <> "@" <> p n
+prettyPrimCon (AnyValue t)  = parens $ "AnyValue @" <> p t
+prettyPrimCon (SumCon c l r) = parens $ "SumCon" <+> p c <+> p l <+> p r
+prettyPrimCon con = prettyExprDefault (ConExpr con)
+
+instance {-# OVERLAPPABLE #-} (Pretty ty, Pretty e, PrettyLam lam) => Pretty (PrimCon ty e lam) where
```
apaszke, line 189: This overlap is not great, but there's no way to pattern match for `Left`/`Right` pretty printing in the general instance 😕

PR #82, opened 2020-05-14
Minor cleanup and reorg of Type.hs

apaszke:
As I was reading the file to better understand the type checking
machinery, I thought it might be a nice occasion to do a small cleanup
too. Most of it is just shuffling existing code around, to better group
it with relevant functions. I also got rid of all the `check*` functions
that were not part of `HasType` typeclass, because I thought it's a bit
weird that they follow the same naming, while each has its own slightly
different meaning.

### DIFF
--- a/src/lib/Type.hs
+++ b/src/lib/Type.hs
@@ -154,7 +154,7 @@ checkKind ty = case ty of
   TC con -> do
     let (conKind, resultKind) = tyConKind con
     void $ traverseTyCon conKind (\(t,k) -> checkKindIs k t)
-                                 (\(e,t) -> checkType e >>= checkConstraint t)
+                                 (\(e,t) -> checkType e >>= checkTypeEq t)
     return resultKind
 
 checkKindIs :: Kind -> Type -> TypeM ()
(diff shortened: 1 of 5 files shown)


src/lib/Type.hs
```diff
@@ -186,116 +167,101 @@ checkKindEq k1 k2 | k1 == k2  = return ()
...
+        eTy <- traverseExpr con return checkPureType checkPiType
+        liftM pureType $ checkConType eTy
+      OpExpr op -> do
+        eTy <- traverseExpr op return checkTypeAndAtom checkPiType
+        checkOpType eTy
+      where checkTypeAndAtom x = liftM (fromAtomicFExpr x,) (checkPureType x)
     SrcAnnot e pos -> addSrcContext (Just pos) $ checkEffType e
     Annot e ty -> do
-      ty' <- checkPureType e
+      (eff, ty') <- checkEffType e
```
apaszke, line 237: By the way this was a bug, but was currently getting masked by the fact that inference drops all annotations!

dougalm: good catch!

PR #86, opened 2020-05-15
Use type-driven casting instead of from/toScalarAtom

apaszke:
Silly? Definitely! More readable? (Arguably) yes! Too much for this codebase? Who knows?!


apaszke:
(@dougalm note that if you think that this doesn't go too far, then we might be able to apply the same trick to `Imp`)

PR #91, opened 2020-05-21
Use a single type environment for type and kind checking

apaszke:


src/lib/Inference.hs
```diff
@@ -26,22 +26,23 @@ import PPrint
...
+type InferM = ReaderT JointTypeEnv (SolverT (Either Err))
 
 inferModule :: TopInfEnv -> FModule -> Except (FModule, TopInfEnv)
 inferModule (tyEnv, tySubEnv) m = do
   let (Module bid (imports, exports) body) = m
   checkImports tyEnv m
   (body', tySubEnvOut) <- expandTyAliases tySubEnv body
-  (body'', tyEnvOut) <- runInferM tyEnv $ catTraverse (inferDecl noEffect) body'
+  (body'', tyEnvOut) <- runInferM tyEnv $
+    catTraverse (inferDecl noEffect) fromNamedEnv body' tyEnv
```
apaszke, line 37: This change to `catTraverse` is a little disappointing. I think we might be able to run the outer inference with `ReaderT TypeEnv`, and switch to `ReaderT JointTypeEnv` once we enter `check` in `inferDecl`. Then the environment around here would truly be monoidal and those changes wouldn't have been necessary. I played around with this idea for a while but I decided that it's better to leave it alone for now.

PR #94, opened 2020-06-05
Avoid assumptions about dimension sizes being statically known

apaszke:


src/lib/Serialize.hs
```diff
@@ -224,14 +232,68 @@ liftExceptIO :: Except a -> IO a
...
+          (IntLit n) <- scalarFromArray arr
+          Just n
+        loadBound _ = Nothing
+
+-- Similar to flattenType, except it only returns the missing arrays
+flattenVal :: Val -> [(Var, ArrayType)]
+flattenVal = flattenVal' 1
+
+flattenVal' :: Int -> Val -> [(Var, ArrayType)]
+-- TODO: Make the JAX backend return atoms with only ArrayTy
```
apaszke, line 272: TODO: JAX backend currently doesn't use `ArrayTy`, because I haven't implemented the type recovery based on the AFor statements. This is left as a future cleanup.

src/lib/TopLevel.hs
```diff
@@ -193,8 +195,17 @@ evalBackend expr = do
...
-      return $ reStructureArrays (getType expr) $ map Var outVars
+      let exprType = getType expr
+      lift $ modifyMVar engine $ \env -> do
+        let inPtrs = fmap (env !) inVars
+        outPtrs <- callLLVM logger llvmFunc inPtrs
+        let outTys = flattenType exprType
+        -- TODO: Assert length of outPtrs equal to outTys
+        let (outNames, env') = nameItems (Name ArrayName "arr" 0) env outPtrs
+        let outVars = zipWith (\v (_, b) -> v :> ArrayTy b) outNames outTys
+        -- XXX: This assumes that the implementation of flattenType and makeDest in Imp yield the same order!
```
apaszke, line 206: TODO: It should be the responsibility of the backend to return the atom with the appropriate type. Right now we assume that the flattening implementations from `Imp` and `JAX` can always match exactly with the unflattening implemented in `Serialize`

PR #100, opened 2020-06-17
Hardcode support for triangular arrays

apaszke:
This is just to make sure that we are actually well prepared to handle
tables with dependent dimensions, because Imp and Serialization have
been made aware of the necessary changes. Now we can start implementing
support for more general (structured) raggedness.


oxinabox:
Does something need ot change in the [Cholesky example](https://github.com/google-research/dex-lang/blob/main/examples/chol.dx) to take full advantage of this, or can that remain the same because this change is just an implementation detail?

apaszke:
@oxinabox Good point! This PR makes it possible to update the cholesky example to construct a triangular array instead of padding the upper diagonals with zeros.

apaszke:
This is superseded by #109.

PR #101, opened 2020-06-17
Erase the distinction between terms and types

dougalm:
We've been gradually adding dependent features on top of our Milner-ish type system and it's become an awkward hybrid. This PR rebuilds the type system from scratch, embracing dependent types and removing the type-term distinction completely. The result is simpler and more expressive. (There must be a catch, and I'm sure we'll learn what it is soon enough!)

Moving to dependent types was Adam Paszke's first suggestion on looking closely at Dex. It also seems to be the direction the typed functional programming community is heading generally. Our design borrows ideas from many places, especially the work of Edwin Brady and Conor McBride.

### No more special type-level constructs

One surprising (to me) result is that the dependent system is much simpler to describe and implement than the original Milner version. We've been able to get rid of several bespoke language features for doing things at the type level since we can now just use the ordinary (and more general) term-level versions:

```
Milner-style types          | dependent types
----------------------------+--------------------------
type abstraction            | (implicit) lambda
forall                      | dependent arrow
(implicit) type application | (implicit) application
type alias                  | let binding
parametric type alias       | let binding with lambda
```

### Checking type equality

We might be concerned that it's now expensive to compare types for equality, since types can be arbitrary terms. For example, checking that this program is well-typed would require showing that `expensive_plus 2 2` is equal to `4`.

``` 
xs : (Fin 4)=>Real = ... 
n = expensive_plus 2 2

for i:(Fin n). xs.i 
```

The obvious pragmatic solution is to only attempt a limited amount of reduction and then fall back to syntactic equality. Currently we inline let bindings and do beta reduction but we don't evaluate any primitive operations like `+`. In the future we might want to go further. Perhaps we could evaluate any expression that doesn't require a loop, with the assumption that in numerical programming all the non-trivial work is in loops.

### Type inference

As Conor McBride points out, there are two different things that we often lump under "type inference". There's implicit *application*, where we turn `id 1` into `id Int 1` and then there's implicit *abstraction*, where we turn `\x. x` into `\a:Type. \x:a. x`. By a happy coincidence, implicit application is very useful and also easy to implement, whereas implicit abstraction, which is much harder to implement, is not actually very useful! I don't think we use it anywhere in the Haskell program that implements Dex, for example. As programmers, we're actively eager to provide explicit annotations for top-level polymorphic functions. It's what lets us reason locally about types.

Roughly following [Practical type inference for arbitrary-rank types](https://www.microsoft.com/en-us/research/wp-content/uploads/2016/02/putting.pdf), we implement a bidirectional type inference system, in which we're either *checking* that a term has a particular type or *inferring* the type of a term. In either case, we produce an explicitly typed term in the core language as a result. We do implicit *application* everywhere but we only do implicit *abstraction* when we're in checking mode, e.g. checking a user-supplied annotation. So we'll wrap `\a:Type. ...` around `\x. x` in `id : (a:Type -> a -> a) = \x. x` but not in `id = \x. x`. Importantly, we're in checking mode when we look at the argument of a function application, so we can infer the right polymorphic type when we use our runST-like effect dischargers like `withAccum : (h:Type ?-> Ref h w -> {Accum h|eff} a) -> {|eff} (a & w)`

Thanks to Adam Paszke and Dimitrios Vytiniotis for helping figure out how to make this work.

### Implicit arguments and holes

We now allow functions with Idris-style implicit arguments. These replace and generalize forall/type-abstraction, where the arguments can be ordinary terms, not just types. We write the type of such a function as `a ?-> b`, with `?->` a variant on the arrow type to go along with our other three (and counting):
```
a  -> {eff} b  Function arrow (with effects)
a  => b        Table arrow
a ?-> b        Implicit function arrow
a --o b        Linear arrow (for AD)
```

Also following Idris, free lowercase variables in top-level type annotations are lifted as implicit arguments of type `Type` by the parser, so we can still write `id : a -> a`, which is syntactic sugar for `id : (a:Type) ?-> a -> a`.

Arguments to implicit functions are filled in during type inference in exactly the same way we used to fill in implicit type arguments. We can also use the same mechanism to fill in arguments for *explicit* functions, by providing a hole `_` instead of an ordinary argument. At some point we'll add the converse mechanism, for explicitly providing arguments for implicit functions.

### Lexical syntax changes

Since we've lost the syntactic distinction between terms and types, we can't "pun" type constructors and term constructors, because we can't (easily) distinguish based on the context. Previously, we used `(,)` both for the pair constructor, like `(1,2.3)`, and the type constructor of the type of pairs, `(Int, Real)`. Now we use `&` (evoking Curry-Howard) for the type constructor, `(Int & Real)`. Similarly, we used literal integers, like `3`, for literal index sets. Now we write `Fin 4` for the index set `{0, 1, 2, 3}`.

We also introduce Coq-style syntactic sugar for let-bound lambdas that lets us share a binder between the type and the term:

```
def id (a:Type) (x:a) : a = x
```

It's still valid to use the Haskell-style "unzipped" form:
```
id : a:Type -> a -> a =
  \a x. x
```

I can't decide which one I prefer.

### Region types

References now carry an ST-style "region" parameter. References with the same region paramater may alias each other, which is important for efficient AD (we want aliased indexed reads to become aliased indexed accumulations).

### Small changes

  * Parser improvements (especially distinguishing lexeme parsers from the rest)
  * Switch to ordinary names instead of de Bruijn indices for pi binders, handling alpha renaming with skolems when we check equality.
  * Define more things in the prelude rather than creating special built-in syntactic forms. This is easier to do now that our type system is more expressive. 
  * Like Haskell, treat parenthesized operators `(+)` as names, allowing them to be defined like ordinary variables and used in prefix position, like `(+) 1.0 2.0`.

### Still to do: typeclasses, linearity, AD of effects

We're hoping to handle typeclasses like Idris does, as a special sort of implicit argument representing a dictionary that can be synthesized by the compiler from available definitions. Linearity and AD shouldn't cause any special problems, but I think we should merge with main even though they're currently not ready.


makefile
```diff
@@ -45,15 +45,18 @@ libdex: cbits/libdex.so
...
+                regression brownian_motion particle-swarm-optimizer \
                 ode-integrator
+
+# example-names = linear-tests ad-tests include-test chol
 quine-test-targets = $(example-names:%=run-%)
 
 doc-names = $(example-names:%=doc/%.html)
 
-tests: test-prep quine-tests repl-test
+# tests: test-prep uexpr-tests quine-tests repl-test
```
apaszke, line 58: I guess this and the duplicate `example-names` are no longer necessary?

src/lib/Embed.hs
```diff
@@ -4,238 +4,241 @@
...
 
-type EmbedT m = CatT EmbedEnv m  -- TODO: consider a full newtype wrapper
-type Embed = Cat EmbedEnv
+newtype EmbedT m a = EmbedT (ReaderT EmbedEnvR (CatT EmbedEnvC m) a)
+  deriving (Functor, Applicative, Monad, MonadIO, MonadFail)
+
+type Embed = EmbedT Identity
+type EmbedEnv = (EmbedEnvR, EmbedEnvC)
+type EmbedEnvC = (Scope, [Decl])
+type EmbedEnvR = (Name, EffectRow)
```
apaszke, line 46: Can you please add some comments explaining what those things are (`EmbedEnvR` in particular)?

src/lib/Env.hs
```diff
@@ -29,9 +29,11 @@ infixr 7 :>
 newtype Env a = Env (M.Map Name a)  deriving (Show, Eq, Ord, Functor, Foldable, Traversable)
 
 -- TODO: consider parameterizing by namespace, for type-level namespace checks.
-data Name = Name NameSpace Tag Int | NoName | DeBruijn Int
+-- `NoName` is used in binders (e.g. `_ = <expr>`) but never in occurrences.
+-- TODO: Consider putting it under a separate `Binder` type instead.
```
apaszke, line 33: Yeah why not make binders `Maybe Name`s (doesn't have to happen now)?

src/lib/Imp.hs
```diff
@@ -27,76 +26,69 @@ import Env
...
       copyAtom outDest ans
       return vsOut
 
 runImpM :: [ScalarTableVar] -> ImpM a -> a
 runImpM inVars m = fst $ runCat m (mempty, (inVarScope, mempty))
   where
     inVarScope :: Scope
-    inVarScope = foldMap varAsEnv $ fmap (fmap $ const ()) inVars
+    inVarScope = fmap (const Nothing) $
+      foldMap varAsEnv $ fmap (fmap $ const ()) inVars
```
apaszke, line 51: Can't you replace `const ()` with `const Nothing` and skip the outer `fmap` here?

src/lib/Imp.hs
```diff
@@ -285,40 +285,37 @@ intToIndex ty _ = error $ "Unexpected type " ++ pprint ty
...
-  liftM getReverse $ flip evalStateT (IIntVal 1) $
-    forM (Reverse xs) $ \(x, ty) -> do
-      stride  <- get
-      size    <- lift $ indexSetSize ty
-      stride' <- lift $ impMul stride size
-      put stride'
-      return (x, ty, stride)
+fromILitInt :: IExpr -> Int
+fromILitInt (ILit (IntLit x)) = x
+fromILitInt expr = error $ "Not an int: " ++ pprint expr
```
apaszke, line 309: This is unused I think

src/lib/Imp.hs
```diff
@@ -285,40 +285,37 @@ intToIndex ty _ = error $ "Unexpected type " ++ pprint ty
...
+
+toImpBaseType :: Type -> BaseType
+toImpBaseType (TabTy _ a) = toImpBaseType a
+toImpBaseType (TC con) = case con of
+  BaseType b       -> b
+  IntRange _ _     -> IntType
+  IndexRange _ _ _ -> IntType
+  _ -> error $ "Unexpected type: " ++ pprint con
+toImpBaseType ty = error $ "Unexpected type: " ++ pprint ty
```
apaszke, line 318: This is unused too (I removed both of those in a recent commit)

src/lib/Imp.hs
```diff
@@ -336,26 +333,23 @@ indexSetSize (TC con) = case con of
       ExclusiveLim x -> indexToInt n x
       Unlimited      -> indexSetSize n
     impSub high' low'
-  RecType r -> do
-    sizes <- traverse indexSetSize r
-    impProd $ toList sizes
+  PairType a b -> do
+    a' <- indexSetSize a
+    b' <- indexSetSize b
+    impMul a' b'
```
apaszke, line 339: nit: `bindM2` could simplify this a little

src/lib/Imp.hs
```diff
@@ -336,26 +333,23 @@ indexSetSize (TC con) = case con of
...
   BaseType BoolType -> return $ IIntVal 2
-  SumType (l, r) -> bindM2 impAdd (indexSetSize l) (indexSetSize r)
+  SumType l r -> bindM2 impAdd (indexSetSize l) (indexSetSize r)
   _ -> error $ "Not implemented " ++ pprint con
 indexSetSize ty = error $ "Not implemented " ++ pprint ty
 
 elemCount :: ScalarTableType -> ImpM IExpr
 elemCount t = case t of
-  TabType pi | isDependentType pi ->
-    error $ "Tables with sizes of dimensions dependent on previous dimensions are not supported: " ++ pprint t
```
apaszke, line 350: Nooooo, this is super important! We cannot handle tables with dimensions dependent on other dimensions, and this is our primary safeguard

src/lib/Imp.hs
```diff
@@ -336,26 +333,23 @@ indexSetSize (TC con) = case con of
...
   TabTy n b -> bindM2 impMul (indexSetSize n) (elemCount b)
   BaseTy _  -> return IOne
   _ -> error $ "Not a scalar table type: " ++ pprint t
 
 offsetTo :: ScalarTableType -> IExpr -> ImpM IExpr
 offsetTo t i = case t of
-  TabType pi | isDependentType pi ->
```
apaszke, line 357: As above

src/lib/TopLevel.hs
```diff
@@ -241,39 +226,20 @@ requestArrays backend vs = case backend of
...
-  where
-    -- Substitutes all arrays appearing in the type annotations inside x
-    substAtomTypes :: Atom -> IO Atom
-    substAtomTypes a = case a of
-      Var v -> Var <$> traverse substInType v
-      -- No substitutions under lambdas. This might cause some issues if we wanted
-      -- to print their body, but it's questionable whether we even want to do that.
-      Con c -> Con <$> traverseExpr c substInType substAtomTypes return
-      _     -> error $ "Unexpected atom: " ++ pprint x
+  x' <- substArrayLiterals' backend (arrayVars (getType x)) x
```
apaszke, line 231: Might be worth adding a comment mentioning the short correctness proof we've arrived at yesterday

src/lib/Syntax.hs
```diff
@@ -146,119 +121,123 @@ scalarTableBaseType t = case t of
...
+      | ConExpr (PrimCon e)
+      | OpExpr  (PrimOp  e)
+      | HofExpr (PrimHof e)
+        deriving (Show, Eq, Generic, Functor, Foldable, Traversable)
+
+data PrimTC e =
+        BaseType BaseType
+      | IntRange e e
+      | IndexRange e (Limit e) (Limit e)
+      | ArrayType ArrayType
```
apaszke, line 177: I think this one should no longer be used anywhere and can be removed

src/lib/Syntax.hs
```diff
@@ -146,119 +121,123 @@ scalarTableBaseType t = case t of
...
+type SrcPos = (Int, Int)
 
-data Atom = Var Var
-          | TLam [TVar] [TyQual] Expr
-          | Con Con
-            deriving (Show, Eq, Generic)
+type PatP  a = WithSrc (PatP' a)
+data PatP' a = PatBind a
+             | PatPair (PatP a) (PatP a)
+             | PatUnit  deriving (Show, Eq, Functor, Foldable, Traversable)
```
apaszke, line 156: Why is this necessary? Patterns should always be non-empty, so I don't think you actually need this?

src/lib/Syntax.hs
```diff
@@ -146,119 +121,123 @@ scalarTableBaseType t = case t of
...
-           | LetPoly Var FTLam
-           | TyDef TVar Type
-             deriving (Show, Eq, Generic)
-
-type Var  = VarP Type
-data FTLam = FTLam [TVar] [TyQual] FExpr  deriving (Show, Eq, Generic)
-type FModule = ModuleP [FDecl]
-
-data RuleAnn = LinearizationDef Name    deriving (Show, Eq, Generic)
+type UExpr = WithSrc UExpr'
```
apaszke, line 126: Might be nice to preserve the `F*` naming convention instead of `U` (which I suppose comes from unified and made sense when there were two frontend syntaxes, but now there is only one).

src/lib/TopLevel.hs
```diff
@@ -187,20 +172,20 @@ initializeBackend backend = case backend of
   _ -> error "Not implemented"
 
 arrayVars :: HasVars a => a -> [Var]
-arrayVars x = map (\(v@(Name ArrayName _ _), L ty) -> v :> ty) $ envPairs (freeVars x)
+arrayVars x = [v:>ty | (v@(Name ArrayName _ _), ty) <- envPairs (freeVars x)]
```
apaszke, line 175: I prefer the old implementation because it made the existence of any non-array free variable a crash, whereas this one will happy skip such vars.

src/lib/Simplify.hs
```diff
@@ -30,183 +27,204 @@ data SimpOpts = SimpOpts { preserveDerivRules :: Bool }
...
 -- in practice it also has to happen sooner or later, because the Imp pass assumes that only array
 -- variables are imported.
-type SimplifyM a = ReaderT SimpEnv (ReaderT ((SubstEnv, RuleEnv), SimpOpts) Embed) a
+type SimplifyM a = ReaderT SimpEnv (ReaderT (SubstEnv, SimpOpts) Embed) a
 
-simplifyModule :: SubstEnv -> RuleEnv -> Module -> (Module, SubstEnv)
-simplifyModule substEnv rulesEnv m = (mOut', subst (envOut', mempty) envOut)
-  where (mOut , envOut ) = simplifyModuleOpts (SimpOpts True ) (substEnv, rulesEnv) m
-        (mOut', envOut') = simplifyModuleOpts (SimpOpts False) (substEnv, rulesEnv) mOut
+simplifyModule :: SubstEnv -> Module -> (Module, SubstEnv)
```
apaszke, line 32: Do any simplification contracts change after this PR? IIRC you said that your change will make it optional to inline the atoms instead of the previous behavior which has forced it. I don't remember what it was exactly, but does this change anything?

makefile
dougalm: Yes, though it's good to keep track of tests that we've temporarily disabled. Added TODOs.

src/lib/Embed.hs
dougalm: Good idea.

src/lib/Imp.hs
dougalm: Yep, good one

src/lib/Imp.hs
dougalm: :+1:

src/lib/Imp.hs
dougalm: Sorry, my bad. This was from the comment-stuff-out-until-it-compiles phase of my merge.

src/lib/Syntax.hs
dougalm: Unit patterns can by useful. `\(). <expr>` can be clearer than `\_. <expr>`.

src/lib/Syntax.hs
dougalm: Actually the `U` is for "untyped"! We used to only have `UExpr`, then `FExpr` came in alongside it. Now we're back to `UExpr`. I don't have strong feelings either way.

src/lib/TopLevel.hs
dougalm: Oh good point. In the merge I read them as identical and picked an arbitrary one.

src/lib/Simplify.hs
dougalm: Yes, it's now syntactically possible to have atoms on the rhs of a decl, but it shouldn't happen after this pass. As we discussed, I'll write a checker for the simplified IR that documents and checks these rules.

PR #102, opened 2020-06-19
Get rid of some compilation warnings

apaszke:


src/lib/Syntax.hs
```diff
@@ -728,6 +727,9 @@ instance Ord SourceBlock where
 
 -- === Synonyms ===
 
+varType :: Var -> Type
+varType = varAnn
```
dougalm, line 731: Any reason to have `varType` instead of just using `varAnn`? Maybe the more specific type conveys intent?

apaszke: Yeah, that was my idea. `varAnn` does not indicate that it returns a type (and in some cases it doesn't!), but this is how we _usually_ use it in Dex

PR #103, opened 2020-06-19
Various small syntax tweaks

hexahedria:
Changes a few things related to syntax and syntactic sugar:

- ~~There is now a syntax for implicit lambda functions. For instance, you could write~~
  ```
  asidx2 : a:Type ?-> Int -> a = \a? \i. fromOrdinal a i
  ```
  ~~I'm using `\a?` since I like that syntax, but I'm happy to change it to something else.~~ [edit: going to revisit this later]
- Functions `asidx` and `(@)` have been implemented as syntactic sugar for `fromOrdinal`. This makes it possible to round-trip indices:
  ```
  >=> fromOrdinal (Fin 10) 3
  3@(Fin 10)
  >=> 3@(Fin 10)
  3@(Fin 10)
  ```
  I picked a precedence fairly arbitrarily. Not sure if there's a better choice.
- When you try to print something that doesn't have `prettyVal` it prints a warning instead of crashing the repl:
  ```
  >=> Int
  -- Unexpected prettyVal argument, falling back to Haskell pretty-printer:
  Int
  ```
- Signed literals are now parsed differently. Before this change, the expression `1.+1.` was parsed as `(1.) (+1.)` = `(1.) (1.)`, i.e. as a function application, leading to a confusing type error. Now, it parses as `(+) 1. 1.`. The downside is that, when you actually want to do an application, you need extra parens: `f (-1.)`. This matches Haskell's behavior, as far as I can tell, and it seems like a reasonable tradeoff to me. (Implementation-wise, the parser does not consider signed literals to be part of the leaf expression, but instead as a prefix operator only. The prefix parser is then responsible for negating the value when its argument is a literal.) Note that this change also removes the unary plus for numeric literals, but there's no unary plus operator anyway so I don't feel like that's a major loss.
- ~~When reading some of the dex tests, I was confused by uses of the implicit arrow in the `def` syntax. For instance, `def ixadd (n:Type) ?-> (i:n) (x:Int) : n = ...` has an arrow between the first two arguments but nothing between the second two. I've implemented a possible alternative syntax `def ixadd (n:Type)? (i:n) (x:Int) : n = ...` which feels more natural to me. I'm not 100% sold on this, though, so happy to remove it.~~ [edit: going to revisit this later]

I made these changes mostly to start getting familiar with the codebase. I'd be happy to split out these changes into individual PRs, if that would be better (or undo some of them entirely).

Also, I'm not sure if these things deserve tests, and if so, where those tests should live.


oxinabox:
> When you try to print something that doesn't have prettyVal it prints a warning instead of crashing the repl: 

🎉 

> Signed literals are now parsed differently. Before this change, the expression 1.+1. was parsed as (1.) (+1.) = (1.) (1.), i.e. as a function application, leading to a confusing type error. Now, it parses as (+) 1. 1..

I am not sure that i wouldn't find that *more* confusing.

Right now error message is:
```
>=> (1.) (+1)
Type error:
Expected: (a->b)
  Actual: Real
(Solving for: [a:Type, b:Type])

(1.) (+1)
 ^^
```
Seems clear enough: expected something that was callable, but you gave me a `Real` instead.

It is inline with Python's error:
```python
>>> (1.)(+1)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: 'float' object is not callable
```
and R's error
```R
> (1.)(+1)
Error: attempt to apply non-function
```

and OCaml's error:
```ocaml
> (1.)(+1)
Error: This expression has type float
       This is not a function; it cannot be applied.
```


(Fun aside: in Julia `(1.)(+1)==1.` because julia jhas juxtaposition multiplication)

> The downside is that, when you actually want to do an application, you need extra parens: f (-1.). This matches Haskell's behavior, as far as I can tell, and it seems like a reasonable tradeoff to me. (Implementation-wise, the parser does not consider signed literals to be part of the leaf expression, but instead as a prefix operator only. The prefix parser is then responsible for negating the value when its argument is a literal.) Note that this change also removes the unary plus for numeric literals, but there's no unary plus operator anyway so I don't feel like that's a major loss.

This seems undesirable to me.
It seems best to leave the identification of literals (negative or otherwise) to the lexer, rather than parsing it as a function application that is then special cased to optimize out.
I think it is reasonable to remove support for numeric literals prefixed with `+` if that is useful which this PR does do,
but having to decorate  negative literals when passed as arguments is annoying

oxinabox:
> Functions asidx and (@) have been implemented as syntactic sugar for fromOrdinal. This makes it possible to round-trip indices:

🎉  this used to work i think when  https://github.com/google-research/dex-lang/commit/1c709c42f7f2092fd6489ca49ba899ca8aee2b6a
I didn't notice when it stopped working
but it was useful and now i really miss it.

hexahedria:
Sorry, maybe I wasn't clear what the issue I was trying to address was? On `main`:
```
>=> 1.+1.
Type error:
Expected: (a->b)
  Actual: Real
(Solving for: [a:Type, b:Type])

1.+1.
^^
```
vs Haskell
```
> 1.0+1.0
2.0
```
vs Python
```
>>> 1.+1.
2.0
```
Is your opinion that the error in this case is the right behavior?

Also, for comparison, in Haskell:
```
> negate -1

<interactive>:8:1: error:
    • Non type-variable argument in the constraint: Num (a -> a)
      (Use FlexibleContexts to permit this)
    • When checking the inferred type
        it :: forall a. (Num a, Num (a -> a)) => a -> a
```
vs
```
> negate (-1)
1
```
so Haskell seems to want parentheses too.

In terms of the implementation, I agree it's a bit messy. But since application has higher precedence than addition/subtraction, I couldn't figure out a good way of preventing it from eagerly folding the `+` or `-` into a leaf expression for the literal even when it appears in the `1.+1.` context. Perhaps there's something obvious here I'm missing.

oxinabox:
>```
 >=> 1.+1.
Type error:
```

Oh that is not so good. Now I follow.

I do maintain that losing the ability to write `linspace (Fin 300) -2.0 1.0`,
and having to write `linspace (Fin 300) (-2.0) 1.0`,
is an unreasonable price to pay.

Looking ath the haskell error message there
Are we just swapping one confusing error message for another possibly even worse one?
I think Haskell is wrong here.
I guess Haskell users are less focused on numerics that I would expect Dex users to be.

Infix functions in curried languages always feel weird to me
In general people should put spaces between functions and their arguments.
We insist on that for prefix functions.
Insisting on it for infix would make `1.+1.` a parse error.
It would still allow `1. +1.0` but perhaps that is more obvious as being a function call?

**Comment by hexahedria** [comment]
The more I think about this, the more complicated it seems. Some cases to consider:
- `3@(Fin 10)` uses `@` as an infix operator (or, at least, that's how I'm doing it in this PR). Insisting on spaces around infix would force this to be `3 @ (Fin 10)`, right?
- Even if spaces were required for infix operators, `1.+1.` would still parse, right? Because it sees `1.` as a leaf expression, then looks for an application, and parses `+1.` as a leaf expression. I think this means that prefix function applications don't actually require spaces before their arguments, currently? Consistent with that hypothesis, `neg(1.0)` parses fine. Should it?
- What should happen with `x -(1.0)`? Right now on `main` this parses differently than `x -1.0`, in that the first one is `(-) x 1.0` but the second is `x (-1.0)`. As far as I can tell, the first one is parsed as: `x` is a leaf expr, try to parse application since it has high precedence, prefix `-` has lower precedence than application so we get stuck, fall through to infix `-` and continue. Whereas the second one is parsed as: `x` is a leaf expr, try to parse application since it has high precedence, succeed because `-1.0` is a leaf expression. Maybe this is fine, I guess? But it's surprising.
- For that matter, what should happen with `f -x`? Is that `f (-x)`? Or `(-) f x`? If it's the former then maybe we want prefix `-` to have higher precedence than function application?
- What about `(- x)`? Is that a syntax error or is that valid?

The only thing I really feel strongly about here is that `1.+1.` and `1.-1.` should NOT be a type error, that should either be a parse error or be equivalent to `(+) 1. 1.` or `(-) 1. 1.`.

**Comment by dougalm** [comment]
Thanks for the PR! Comments below.

> Functions asidx and (@) have been implemented as syntactic sugar for fromOrdinal

Great! I love that we can now do this in userspace (except for the fixity declaration).

> When you try to print something that doesn't have prettyVal it prints a warning instead of crashing the repl

Excellent!

> Also, I'm not sure if these things deserve tests, and if so, where those tests should live.

Yes, I would add simple examples to test each new feature. You can always put something in `eval-tests.dx` but maybe we should make a new quine test, `parser-test.dx`, for testing lexical syntax specifically. We can even start to test parser errors there, since they need some work.

### Syntax for function arrows

Thanks for thinking about alternative lexical syntax for implicit lambda and implicit lambda via `def`. With both of these, we should keep in mind that implicit functions are just one of five (!) different sorts of arrows we have. I think we should aim for a syntax that treats them uniformly, with a special case for the plain arrow because it's so common. Our five arrows are

```
->    plain
--o   linear
=>    table arrow
?->   implicit function
?=>   implicit typeclass dictionary (WIP on dictionary-synthesis branch)
```

> There is now a syntax for implicit lambda functions. ... `\a? \i. fromOrdinal a i`

I agree that we should have syntax for lambda with a specific arrow. But it doesn't have to be beautiful, because we expect `def ...` and type annotations to do the work in most cases. Generalizing your proposal to the other arrows, perhaps we could have the arrow *as* the terminator, like `\x -> ...` and `\x ?-> ...`. I like that the `->` is the same as Haskell. Perhaps the unannotated binders in `\x y z ?->` (i.e. `x` and `y`) default to `->`? Or Perhaps they all share the `?->`.

> I've implemented a possible alternative syntax `def ixadd (n:Type)? (i:n) (x:Int) : n = ...`

Again, my worry is that this works for the implicit arrow but not for the others. I don't love the current `def` syntax, and I think we should iterate on it. But the idea is that each binder has an arrow associated with it, so you can write `def id (a:Type)?-> (x:a)-> : a = ...`, but because the `->` arrow is so common, you can optionally leave it off, just writing `def id (a:Type)?-> (x:a) : a = ...`

### Unary negation

This is a tricky one! Prefix negation is one of those math notation quirks that we're stuck with. Your change is definitely an improvement over what we have currently (removing prefix `+` and treating negative literals and negated expressions uniformly) so let's merge it.

To go further, and allow `f -x` to parse as `f (-x)` we'd need to commit to whitespace-sensitive parsing. Looking at Haskell, there seems to be a [GHC proposal in the works](https://github.com/ghc-proposals/ghc-proposals/pull/229) to do exactly that for prefix operators. I suspect we'll end up going down that route too, but for now I think we should err on the side of keeping the implementation simple and focusing our efforts on the more interesting and risky parts of the language.

hexahedria:
All right, for now I'll roll back the changes for implicit lambda, but I'll keep thinking about it. It does make sense to have a consistent syntax for all the arrows. Maybe the same solution would work for lambdas and `def`s? In either case, that probably should go in a separate PR.

I'll keep the `asidx`/`(@)`, `prettyVal`, and unary negation and add some tests.

dougalm:
Looks good! Mind rebasing/merging onto main and then I'll merge it?

PR #105, opened 2020-06-20
remove return type from dex_free

oxinabox:
This fixes the warning during compilation
```
cbits/libdex.c:148:1: warning: control reaches end of non-void function [-Wreturn-type]
```

I am not sure what that `char *` was intended for but it was not being returned I guess it could have been the same as the input, but i fail to see why we would want an extra reference to a pointer freed  memory.

### DIFF
--- a/cbits/libdex.c
+++ b/cbits/libdex.c
@@ -14,7 +14,7 @@ char* malloc_dex(long nbytes) {
   return malloc(nbytes);
 }
 
-char* free_dex(char* ptr) {
+void free_dex(char* ptr) {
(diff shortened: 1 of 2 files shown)


dougalm:
> I am not sure what that char * was intended for

Nothing at all! I just tinkered with the LLVM codegen and the C shim until something worked. I suspect we should be able to get rid of the `free_dex` and `malloc_dex` completely and just call into stdlib directly. They're only there because I could't figure out how to get linking to work. Thanks for the cleanup!

PR #108, opened 2020-06-22
Replace AFors with table lambdas

apaszke:
The recent major refactor has removed DeBruijn variables, meaning that
`AFor`s were no longer considered binders. However, if we want to support
arrays with dimensions that can depend on each other, there needs to be
a way to refer to a variable bound by a previous views.

This could be fixed by making AFor carry around a binder (i.e. wrap an
`Abs Atom`), but at this point it seemed like there was really no excuse
not to try out representing the array-of-structures view as a table
lambda. Once I did that, it turned out that there was a bunch of other
changes that paired with it nicely, which unfortunately made for a
relatively large change in the end. Here's a summary of the main updates:
* The structure-of-arrays conversion is not done using table lambdas,
  instead of the dedicated `AFor` and `AGet` constructors.
* The lack of `AGet` atom also means that we no longer can have scalar
  array references embedded in the types, which is actually nice.
* `AFor`s were used to represent destinations (i.e. writeable atoms) in
  the Imp translation, but they did not enforce the guarantee that their
  terms terminate with variables to references. The new `Dest` structure
  guarantees that and makes the code quite a bit clearer.
* I've added back an `ArrayType`, which is meant to be embedded in table
  lambdas. The only two operations that can be applied to values of this
  type are `ArrayOffset` and `ArrayLoad`. Note that `ArrayOffset`
  requires a linear offset, so all the details of linear index
  calculation have to be embedded in the lambda by each respective
  backend. This has the upside of allowing backends to use different
  representations for their arrays by e.g. padding any raggedness.

This work should allow me to rebase my previous attempts of adding
support for ragged arrays (see #100).

Unfortunately, all this does not come with any downsides. The biggest
problem right now is that the use of table lambdas allows enough
flexibility that the only reliable way to take the data out of the
system (e.g. for plotting) requires a reevaluation of the lambda over
the whole index set which usually boils down to making a copy of the
array in a very inefficient manner. We'll have to think more about this
problem in the future, but for now the plotting performance is not a
big concern.

### DIFF
--- a/src/lib/Imp.hs
+++ b/src/lib/Imp.hs
@@ -132,7 +132,10 @@ toImpOp (maybeDest, op) = case op of
              FFICall "int_to_index_set" IntType [i', n']
     returnVal $ toScalarAtom resultTy ans
   IdxSetSize n -> returnVal . toScalarAtom resultTy =<< indexSetSize n
-  IndexAsInt i -> returnVal . toScalarAtom IntTy    =<< indexToInt (getType i) i
+  IndexAsInt idx -> case idx of
+    Con (AsIdx _ i)  -> returnVal $ i
+    Con (AnyValue t) -> returnVal $ anyValue t
+    _                -> returnVal . toScalarAtom IntTy =<< indexToInt (getType idx) idx
   Inject e -> do
     let rt@(TC (IndexRange t low _)) = getType e
     offset <- case low of
@@ -150,11 +153,6 @@ toImpOp (maybeDest, op) = case op of
     returnVal $ Var subrefVar
   ArrayOffset arr off -> returnVal . toArrayAtom =<< impOffset (fromArrayAtom arr) (fromScalarAtom off)
   ArrayLoad arr       -> returnVal . toScalarAtom resultTy =<< load (fromArrayAtom arr)
-  UnwrapIndex idx -> case idx of
-    (Con (AsIdx _ e))  -> returnVal e
-    -- TODO: Handle the AnyValue instance for index sets in simplification instead of here.
-    (Con (AnyValue t)) -> returnVal $ toScalarAtom resultTy (anyValue t)
-    _                  -> error $ "Unsupported UnwrapIndex argument: " ++ pprint idx
   Cmp _ _ _ -> error $ "All instances of Cmp should get resolved in simplification"
   _ -> do
     returnVal . toScalarAtom resultTy =<< emitInstr (IPrimOp $ fmap fromScalarAtom op)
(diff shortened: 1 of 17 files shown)


src/lib/Imp.hs
```diff
@@ -143,73 +177,161 @@ toImpHof env (maybeDest, hof) = do
...
+    _  -> error "Unexpected decls"
+
+destToAtom' :: (IVar -> ImpM Atom) -> [Var] -> Dest -> EmbedT ImpM Atom
+destToAtom' fScalar forVars dest = case dest of
+  DFor (Abs v d) -> do
+    scope <- looks $ fst . snd
+    v' <- lift $ case v of
+      (NoName :> t) -> freshVar ("idx" :> t)
+      _             -> freshVar v
+    -- FIXME: This does not propagate the scope back!
```
dougalm, line 278: I think it's already correct. The scope due to the decls isn't relevant any more, now that they're wrapped up in a block. Rather than `runEmbedT`, you could just use `BuildScoped`, which scopes the decls and builds them into a block. (Also above.)

src/lib/Imp.hs
```diff
@@ -259,179 +387,126 @@ toScalarAtom ty ie = case ie of
...
+anyValue (TC (BaseType BoolType))   = ILit $ BoolLit False
+anyValue (TC (BaseType StrType))    = ILit $ StrLit ""
+-- XXX: This is not strictly correct, because those types might not have any
+--      inhabitants. We might want to consider emitting some run-time code that
+--      aborts the program if this really ends up being the case.
+anyValue (TC (IntRange _ _))        = ILit $ IntLit 0
+anyValue (TC (IndexRange _ _ _))    = ILit $ IntLit 0
+anyValue t = error $ "Expected a scalar type in anyValue, got: " ++ pprint t
 
 intToIndex :: Type -> IExpr -> ImpM Atom
```
dougalm, line 414: It's so great that these are no longer imp-specific.

src/lib/Serialize.hs
```diff
@@ -166,102 +154,51 @@ valFromPtrs ty ptrs = do
...
-traverseVal f val = case val of
-  Con con -> do
-    ans <- f con
-    liftM Con $ case ans of
-      Just con' -> return con'
-      Nothing   -> mapM (traverseVal f) con
-  atom -> return atom
+pprintVal val = asStr $ prettyVal val
+
+prettyVal :: Val -> Doc ann
```
dougalm, line 185: Much better!

src/lib/Imp.hs
apaszke: Ohhh nice, I didn't know about `BuildScoped`! I think I agree that this is fine, as the code I generate here should have no free variables, and so no substitution should ever have any effect on its body (except for the subst from beta-reduction of course, but that's ok).

PR #109, opened 2020-06-23
Hardcode support for triangular matrices

apaszke:
Based on the generalization of AFors, we're finally able to support
(structured) ragged arrays, and this is an initial exploration that
hardcodes formulas necessary to handle triangular matrices. An
interesting side effect of having us keep the scalar table type as part
of the array type is that we are now forced to supply to arguments to
both `IOffset` and `ArrayOffset`: one indicating the index value, that
is used to determine the type of the output, while the other one
supplying the linear offset, to be used for evaluation and code
generation.

I think there is an easy option to get rid of the double index in the case of
`ArrayOffset`, although we would have to give up some safety in that the
arrays we would be retrieving from the backends would have no known
length. If we want to keep their length around, we would have to start
returning some extra information e.g. from the LLVM programs, which is
possible as well.

### DIFF
--- a/src/lib/Imp.hs
+++ b/src/lib/Imp.hs
@@ -151,8 +151,10 @@ toImpOp (maybeDest, op) = case op of
     subrefVar  <- freshVar (varName refVar :> RefTy h (snd $ applyAbs a i))
     extend ((subrefVar @> subrefDest, mempty), mempty)
     returnVal $ Var subrefVar
-  ArrayOffset arr off -> returnVal . toArrayAtom =<< impOffset (fromArrayAtom arr) (fromScalarAtom off)
-  ArrayLoad arr       -> returnVal . toScalarAtom resultTy =<< load (fromArrayAtom arr)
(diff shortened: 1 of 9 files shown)


apaszke:
Supersedes #100.

src/lib/Imp.hs
```diff
@@ -267,11 +270,11 @@ destToAtomScalarAction fScalar dest = do
 
 destToAtom' :: (IVar -> ImpM Atom) -> [Var] -> Dest -> EmbedT ImpM Atom
 destToAtom' fScalar forVars dest = case dest of
-  DFor (Abs v d) -> do
+  DFor a@(Abs v _) -> do
     v' <- lift $ case v of
       (NoName :> t) -> freshVar ("idx" :> t)
```
dougalm, line 275: As discussed, I'm worried that the embed monad doesn't know that these variables are in scope. Using `buildLam` should take care of it.

PR #111, opened 2020-06-23
Implement an algebraic simplification engine for index calculation

apaszke:
Turns out that once we start getting ragged dimensions, the index
calculations are no longer quite as simple, and might become polynomial
in certain indices. In a previous PR I have encoded one such formula for
the case of triangular arrays, but of course we should do better than
that and this PR lays the groundwork for this.

The general idea here is that linear offset derivation involves finding
closed forms for a bunch of sum expressions. The new Algebra.hs file
implements an algebraic simplification engine that is able to do some
basic math on polynomials, including summation. It does not support all
possible range expressions just yet (this requires an extension of the
polynomial language to support a `max(..., 0)` operation) and cannot
simplify sums of high degree polynomials (there is a closed formula for
those, but it's not implemented just yet).

### DIFF
--- a/src/lib/Imp.hs
+++ b/src/lib/Imp.hs
@@ -29,7 +29,7 @@ import Env
 import Type
 import PPrint
 import Cat
-import Util (bindM2)
+import qualified Algebra as A
 
 -- Note [Valid Imp atoms]
 --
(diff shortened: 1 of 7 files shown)


src/lib/Imp.hs
```diff
@@ -420,30 +420,17 @@ offsetTo ty i     = fromScalarAtom <$> fromEmbed (offsetToE ty (toScalarAtom Int
...
 -- TODO: Accept an index instead of an ordinal?
 offsetToE :: MonadEmbed m => ScalarTableType -> Atom -> m Atom
 offsetToE ty i = case ty of
-  TabTy (NoName:>_) b -> imul i =<< elemCountE b
-  TabTy v (TabTy (NoName :> TC (IndexRange _ Unlimited (InclusiveLim (Var v')))) (BaseTy _)) | v == v' ->
-    triangularSum i
-  TabTy _ _ ->
-    error $ "Tables with sizes of dimensions dependent on previous dimensions are not supported: " ++ pprint ty
   BaseTy _  -> error "Indexing into a scalar!"
+  TabTy (NoName :> _) b -> imul i =<< elemCountE b
```
apaszke, line 430: This is only necessary, because the algebra backend cannot handle exclusive bounds in index ranges. It should be removed soon!

PR #117, opened 2020-06-25
Fix build badge branch name

lancelet:
- The build badge was pointing to a `master` branch instead of the `main` GitHub branch.


dougalm:
Good catch, thanks! Hey, it's even passing. Who knew? :)

PR #124, opened 2020-06-28
Fix ixsub, which was previously doing addition

lancelet:
`ixsub` was previously adding indices.

I'm not 100% sure how to test this though. Some examples would be really useful.

Initially, I tried this as a test:
```
def identityIx (x:n=>Real) : n=>Real = for i. x.(ixadd (ixsub i 1) 1)

identityIx [0.,1.,2.,3.]
> [0.,1.,2.,3.]  -- what I'd hoped for, not what I got
```
However, because the index value is checked when it is created (ie. when `ixsub` is evaluated), this happens:
```
Index out of bounds. -1 not in [0, 4)
```

Then I thought I had to avoid creating an invalid index (even if it would later become valid before being used), and `select` seemed a likely choice:
```
def shiftRZeroFill (x:n=>Real) : n=>Real = for i. select (ordinal i == 0) 0.0 (x.(ixsub i 1))

:p shiftRZeroFill [5.,6.,7.,8.]
> [0.,5.,6.,7.]  -- what I'd hoped for, not what I got
```
But again:
```
Index out of bounds. -1 not in [0, 4)
```
I guess `select` is evaluating both branches?

I'm a bit stuck thinking about how `ixadd` or `ixsub` could be evaluated at this point, but I'm probably missing something obvious.

### DIFF
--- a/prelude.dx
+++ b/prelude.dx
@@ -202,7 +202,7 @@ def fromOrdinal (n:Type) (i:Int) : n = %asidx n i
 def asidx (n:Type) ?-> (i:Int) : n = fromOrdinal n i
 def (@) (i:Int) (n:Type) : n = fromOrdinal n i
 def ixadd (n:Type) ?-> (i:n) (x:Int) : n = fromOrdinal n $ ordinal i + x
-def ixsub (n:Type) ?-> (i:n) (x:Int) : n = fromOrdinal n $ ordinal i + x
+def ixsub (n:Type) ?-> (i:n) (x:Int) : n = fromOrdinal n $ ordinal i - x
 def iota (n:Type) : n=>Int = for i. ordinal i
 
 -- TODO: we want Eq and Ord for all index sets, not just `Fin n`


dougalm:
> ixsub was previously adding indices.

Oops!

 > I guess select is evaluating both branches?

Yes, that's right. I'd like to add a genuine `if` that only evaluates one side.

> I'm a bit stuck thinking about how ixadd or ixsub could be evaluated at this point, but I'm probably missing something obvious.

For testing purposes, you can just create some indices from integer literals, like this

```
:p
  i = 3 @ (Fin 10)
  (ixadd i 1, ixsub i 1)
```

You could also use the *unsafe* version of int-to-index conversion that doesn't check the range, taking on the proof burden yourself. Right now that's not exposed in the prelude, but it probably should be. (Confusingly, the internal names for the safe/unsafe versions are `IntAsIndex`/`AsIdx` and the exposed names are `%asidx`/`<not exposed>`. We should probably be more consistent and explicit. Perhaps `%asIdxChecked` and `%asIdxUnsafe` for the exposed names, at least.)

Index arithmetic raises all sorts of problems for typed indices, and we're still working through them, as you can see. I think we might two functions: `asIdx : Idx n ?=> Int -> Maybe n` (where the caller can ignore the `Nothing` side and crash if they want to)
and `asIdxClamped : Idx n ?=> Int -> n` which hits the rails and doesn't go past them.

Issue #114, opened 2020-06-24
"Compiler bug" message when printing a lambda


hexahedria:
```
>=> :p \x. x + 1.0
Compiler bug!
Please report this at github.com/google-research/dex-lang/issues

src/lib/TopLevel.hs:177:20-59: Non-exhaustive patterns in lambda
```
Admittedly, printing a lambda is a weird thing to do. But it seems like there could be a better error message here, and it seems like it's failing even before getting to the pretty printer.

dougalm:
Fixed by #133.

PR #133, opened 2020-07-01
Don't simplify away NewtypeLet, add type checks for NewtypeApp.

hexahedria:
Before this change, defining type synonyms for newtypes produced incorrect behavior; the type synonym would be simplified into the contents of the newtype instead of the NewtypeApp type. For instance:

```
-- Works:
@newtype
NewPair : Type -> Type -> Type = \a b. (a & b)

mkNewPair   : (a & b) -> NewPair a b = unsafeNewtypeCon
fromNewPair : NewPair a b -> (a & b) = unsafeFromNewtypeCon

newPair : NewPair Int Real = mkNewPair (1, 2.0)

-- Fails:
NewPairIntReal = NewPair Int Real

:p NewPairIntReal
> (Int & Real)

newPair2 : NewPairIntReal =  mkNewPair (1, 2.0)
> Type error:
> Expected: (Int & Real)
>   Actual: NewPair Int Real
```

This PR does the same reduction at simplification time as is done at reduction time, so that NewtypeLet bindings are treated consistently. In addition, it adds a bit more typechecking for newtypes, to hopefully make it more robust (which involves saving the type of the let binding into the NewtypeApp type constructor).

```
NewPairIntReal = NewPair Int Real

:p NewPairIntReal
> NewPair Int Real

newPair2 : NewPairIntReal = mkNewPair (1, 2.0)

:p fst $ fromNewPair newPair
> 1
```

Also fixes issue #114 by only extracting array names at the top level, since that issue was manifesting here too when trying to print newtype-wrapped types.

(Side comment: I wonder if it's necessary to duplicate the logic for reduceAtom and simplifyAtom? Perhaps some of these steps could be combined in one place instead. Another thing that would allow us to do is to reduce expressions like `fst (Int & Real)`. Maybe for a separate PR, though.)


dougalm:
Looks great!

>  I wonder if it's necessary to duplicate the logic for reduceAtom and simplifyAtom?
Yes, we might want to de-dup at some point.

src/lib/Type.hs
```diff
@@ -309,20 +309,27 @@ withAllowedEff eff m = updateAllowedEff (const eff) m
...
+  TypeKind         -> return TyKind
+  EffectRowKind    -> return TyKind
   JArrayType _ _   -> undefined
-  NewtypeApp _ _ -> return () -- TODO!
+  NewtypeApp _ wrapTy xs -> foldM checkApp wrapTy xs where
+    checkApp :: Atom -> Atom -> TypeM Type
+    checkApp (Pi piTy) x = do
+      x |: absArgType piTy
+      -- Newtype arrows should be pure.
+      let (PlainArrow Pure, resultTy) = applyAbs piTy x
```
apaszke, line 331: It's better to perform checks like this one through code like
```hs
(PlainArrow Pure, resultTy) <- return $ applyAbs piTy x
```
because that will use the `MonadFail` instance of `TypeM` to print a nice error message, instead of trapping the program due to the pattern match failure in pure code.

src/lib/TopLevel.hs
```diff
@@ -169,7 +169,10 @@ initializeBackend backend = case backend of
   _ -> error "Not implemented"
 
 arrayVars :: HasVars a => a -> [Var]
-arrayVars x = map (\(v@(Name ArrayName _ _), (ty, _)) -> v :> ty) $ envPairs (freeVars x)
+arrayVars x = foldMap go $ envPairs (freeVars x)
+  where go :: (Name, (Type, BinderInfo)) -> [Var]
+        go (v@(Name ArrayName _ _), (ty, _)) = [v :> ty]
+        go _ = []
```
apaszke, line 175: Um, that looks incorrect to me? We should not have any free non-array variables here!

src/lib/Type.hs
hexahedria: Oh, is that why that shows up everywhere? I didn't know that about Haskell!

src/lib/TopLevel.hs
hexahedria: Hm, really? Maybe there's a better fix then. In practice it seems like every global is treated as free, so if the expression has any global symbols in it (like `Ord`, or `(&)`) it traps on a pattern match failure.

hexahedria: Should global variables like that not be treated as free? Or is it ok for `GlobalName`s to show up here, and any other name cause an error? (Or is it bad that global names are showing up here?)

apaszke: Hmmm I mean I think you might be right in that we might start seeing global non-array vars due to newtype binders, but in general we should still check that we don't have anything else, because simplification should have inlined them away!

PR #144, opened 2020-07-07
add nix

### LINKED ISSUE #123: Question: Let cabal handle cbits compilation?
Q: Is there a reason to compile `libdex.c` as a separate step in the `makefile`?

I'm keen to try letting Cabal compile `libdex.c`, in the pattern that is often used elsewhere (eg. see this example in [the bytestring package](https://github.com/haskell/bytestring/blob/e6cb01e2ec0bfdd19298418c85f220925a9fa307/bytestring.cabal#L131-L132)).

However, since it might well be set up the way it is for good reasons, I thought I'd check first before I give it a go. Is this a good idea? :-)

btw: my motivation is to allow `stack install` to work properly. Currently, if I `stack install` and then try to run `dex` from a different directory, I see this:
```
$ dex
dyld: Library not loaded: cbits/libdex.so
  Referenced from: /Users/jsm/.local/bin/dex
  Reason: image not found
[2]    48521 abort      dex
```
Since it's trying to load `libdex.so` dynamically rather than statically linking that code.

chessai:
Adds nix, so you can just enter a nix-shell to build with cabal, and also nix-build to build.

Additionally resolves #123 since that made things less annoying and I figured I might as well

### DIFF
--- a/.nix/compiler.nix
+++ b/.nix/compiler.nix
@@ -1,0 +1,7 @@
+rec {
+  ghc865 = "ghc865";
+  ghc883 = "ghc883";
+  ghc8101 = "ghc8101";
+
+  default = ghc865;
+}
--- a/.nix/nixpkgs.nix
+++ b/.nix/nixpkgs.nix
@@ -1,0 +1,23 @@
+{ compiler ? (import ./compiler.nix).default
+}:
+
+let
+  llvmOverlay = self: super: {
+    haskell = super.haskell // {
+      packages = super.haskell.packages // {
+        "${compiler}" = super.haskell.packages."${compiler}".override {
+          overrides = hself: hsuper: with super.haskell.lib; {
+            llvm-hs = unmarkBroken (hsuper.llvm-hs_9_0_1.override {
+              llvm-config = self.llvm_9;
+            });
+
+            llvm-hs-pure = hsuper.llvm-hs-pure_9_0_0;
+          };
+        };
+      };
+    };
+  };
+in import ./pinned-nixpkgs.nix {
+  config = {};
+  overlays = [ llvmOverlay ];
+}
--- a/.nix/pinned-nixpkgs.nix
+++ b/.nix/pinned-nixpkgs.nix
@@ -1,0 +1,15 @@
+{ config ? {}
+, overlays ? []
+}:
+
+with {
+  owner = "NixOS";
+  repo = "nixpkgs-channels";
+  rev = "41aa2272feb21b1300b9f99de3f8371a6ddd8986";
+  sha256 = "0qj3x0jcsz3lbia93a06dzvp2psayxxqxly62kjmsywvsm0picby";
+};
+
+import (builtins.fetchTarball {
+  url = "https://github.com/${owner}/${repo}/archive/${rev}.tar.gz";
+  inherit sha256;
+}) { inherit config overlays; }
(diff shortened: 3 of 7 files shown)


chessai:
@googlebot I signed it!

chessai:
For some reason, all commands segfault. I wonder if this is an LLVM issue?

chessai:
The gdb output is not very helpful:

```gdb
(gdb) file "result/bin/dex"
Reading symbols from result/bin/dex...
(No debugging symbols found in result/bin/dex)
(gdb) r repl
Starting program: /nix/store/0vldjdzm31i83glhxvnkc0kcwaixvkr5-dex-0.1.0.0/bin/dex repl
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/nix/store/xg6ilb9g9zhi2zg1dpi4zcp288rhnvns-glibc-2.30/lib/libthread_db.so.1".
[New Thread 0x7fffe56f3700 (LWP 12335)]
[New Thread 0x7fffe4ef2700 (LWP 12336)]
[New Thread 0x7fffdffff700 (LWP 12337)]
[New Thread 0x7fffdf7fe700 (LWP 12338)]

Thread 1 "dex" received signal SIGSEGV, Segmentation fault.
0x0000000000000000 in ?? ()
```

chessai:
I am going to leave LLVM with debug output compiling overnight. Then maybe gdb will be more helpful.

apaszke:
Thanks @chessai! Maybe it's the `foreign import ccall "threefry2x32" linking_hack` from `LLVMExec` that's causing the trouble? You've changed how we link `cbits` into the Haskell part, so it might be due to that. There might be some useful discussion around that in #123.

chessai:
@apaszke I confirmed that the segfault at startup goes away when changing the ExternFunSpecs mallocFun and freeFun from "malloc_dex" and "free_dex" to "malloc" and "free". It's not immediately clear to me how to resolve this without the hack, but also not clear to me why we can't just use malloc/free here. Though, trying to run any nontrivial program segfaults very quickly as well. Going to debug more later.

lancelet:
I've seen this problem (in the comments I left on #123), but don't yet have a solution. My vague (and possibly wrong) understanding of the situation is as follows...

First, it's not a "normal cbits" situation. AFAIU, in typical uses of static compilation with C code via Cabal, the linker is going to see that symbols referenced from the Haskell code are defined in the object compiled from the C code. Because they're referenced, it will link them into the final binary. However, with dex (again, AFAIU), we're not doing that. Instead, _at run time_, LLVM is creating code that tries to reference the symbols. So the linker never sees that most of the functions are required (except in the case of the single `threefry2x32` function). It's my current hypothesis that because static compilation is taking place and the function aren't used from Haskell code, they're simply not included.

If you look at the symbols defined in the final binary (with the "linking hack" in place, and letting Cabal do static linking), you can see that `threefry2x32` is present, but the other symbols like `malloc_dex` and `free_dex` are not. Because that one function (`threefry2x32`) has a hook in the Haskell code which is calling it.

One solution might be to use the "linking hack" to provide hooks into _all_ the functions (instead of removing it from that one function). That might ensure that they're all present in the final binary. I don't know for sure whether that will work, but it's what I was planning to try next.

Another potential hiccup is that there might be a difference between LLVM's ability to find functions in a dynamically-loaded library (which is what's currently there) vs being able to find them in a statically-compiled binary (which is what we'll get with Cabal doing the linking). I don't know how most of that works. :-)

dougalm:
I think this is superseded by #386

PR #159, opened 2020-07-13
Use the large code model to avoid linking issues on macOS

apaszke:
We need to use the large code model for macOS, because the libC functions
are loaded very far away from the JITed code. This does not prevent the
runtime linker from attempting to shove their offsets into 32-bit values
which cannot represent them, leading to segfaults that are very fun to debug.
It would be good to find a better solution, because larger code models
might hurt performance if we were to end up doing a lot of function calls.


duvenaud:
Yes, that fixed it!  That was fast.

apaszke:
@duvenaud awesome, thanks for checking!

PR #164, opened 2020-07-16
Add user-defined nonrecursive algebraic data types

dougalm:
We can now define our own ADTs in userspace:

```
  data Maybe a:Type =
    Nothing
    Just a
```

And use them with pattern-matching case analysis:

```
case ... of
  Nothing -> ...
  Just x  -> ...
```

The ADTs subsume several language features we'd implemented one-by one: product types (replacing our built-in `Pair`), sum types (replacing our built-in `Either`) and named types (replacing our `newtype` mechanism). ADTs usually enable recursive types too, but we're not allowing that for now, just as we don't allow recursive let bindings.


src/lib/Imp.hs
```diff
@@ -181,6 +200,21 @@ toImpOp (maybeDest, op) = case op of
...
+    return $ Con $ AnyValue ty
+  CoerceOp destTy@PreludeBoolTy x -> case getType x of
+    BoolTy -> do
+      tag <- emitInstr $ IPrimOp $ ScalarUnOp BoolToInt $ fromScalarAtom x
+      returnVal $ Con $ SumAsProd destTy (toScalarAtom IntTy tag) [[], []]
+    ty -> error $ "unexpected type: " ++ pprint ty
+  CoerceOp BoolTy x -> case x of
+    DataCon _ [] 0 [] -> returnVal $ Con $ Lit $ BoolLit False
+    DataCon _ [] 1 [] -> returnVal $ Con $ Lit $ BoolLit True
+    Con (SumAsProd _ tag [[],[]]) -> do
```
apaszke, line 214: What's the difference between `DataCon` and `SumAsProd`? Do we use the first when we know the constructor statically, while the later is for the cases when it might be dynamic?

src/lib/Imp.hs
```diff
@@ -51,10 +51,10 @@ import qualified Algebra as A
 -- TODO: use `Env ()` instead of `Scope`. The problem is that we're mixing up
 -- Imp and Core types in there.
 type EmbedEnv = ((Env Dest, [IVar]), (Scope, ImpProg))
-type ImpM = Cat EmbedEnv
+type ImpM = ReaderT Bindings (Cat EmbedEnv)
 
-toImpFunction :: ([ScalarTableVar], Block) -> (ImpFunction, Atom)
-toImpFunction (vsIn, block) = runImpM vsIn $ do
+toImpFunction :: Bindings -> ([ScalarTableVar], Block) -> (ImpFunction, Atom)
```
apaszke, line 56: What are the bindings here? Do they describe the data declarations? How about using a type alias to make it clear what they represent?

src/lib/Inference.hs
```diff
@@ -33,12 +33,17 @@ data RequiredTy a = Check a | Infer
...
 inferModule scope (UModule decls) = do
-  ((), (_, decls')) <- runUInferM (mapM_ (inferUDecl True) decls) mempty scope
-  return $ Module Typed decls' mempty
+  ((), (bindings, decls')) <- runUInferM mempty scope $
+                                mapM_ (inferUDecl True) decls
+  let bindings' = envFilter bindings $ \(_, b) -> case b of
+                    DataBoundTypeCon _   -> True
+                    DataBoundDataCon _ _ -> True
+                    _ -> False
+  return $ Module Typed decls' bindings'
```
apaszke, line 42: Do I understand correctly that `bindings` contains the local scope, which is where the ADT declarations end up? Why don't we declare them as globals and put them together with `decls`?

src/lib/Imp.hs
dougalm: Yes, that's right. But `SumAsProd` is a wart. We'll be able to get rid of it once we let ourselves use arbitrary (cheapish) expressions to reconstruct atoms from pointers, rather than having to pretend that everything is an atom.

src/lib/Inference.hs
dougalm: This way the core IR doesn't need syntax for data definitions at all. They don't really represent "computation", so that seems appropriate to me.

src/lib/Imp.hs
dougalm: Yes, they describe the data declarations. I don't think they're used any more, now that the data declarations are cached inline.

PR #167, opened 2020-07-22
Add evalpoly() to prelude and tests for any() and all().

duvenaud:
It feels a bit silly to add all these one-liners to the prelude, but they exist in Numpy, Jax, and Julia, and are possible to get wrong.  Also, I needed evalpoly() for a Runge-Kutta solver.

The equivalent function and its gradient also was benchmarked for speed in Julia:
https://twitter.com/sethaxen/status/1259973355869925376
they report 17 microseconds for a 10,000 degree polynomial, I got 25 microseconds on my laptop in Dex.

It looks like they took some care to allow evaluation on things like complex numbers and matrices.  My implementation operates on arbitrary vector spaces, so I hope it works on those cases too out of the box but I haven't tests that yet.

Julia also has a custom rule for the gradient:
https://github.com/JuliaDiff/ChainRules.jl/pull/190/files

Jax has fairly short implementation without a custom rule:
https://github.com/google/jax/pull/3076/files
and there's some discussion of tuning the amount of loop unrolling for different architectures:
https://github.com/google/jax/pull/3076

### DIFF
--- a/examples/eval-tests.dx
+++ b/examples/eval-tests.dx
@@ -718,6 +718,22 @@ arr2d.(1@_)
     r1 := y
 > ((), (2, 1))
 
+
+:p any [True, False]
+> True
+:p any [False, False]
+> False
+:p any []:(Fin 0)=>Bool
+> False
+
+:p all [True, True]
+> True
+:p all [False, True]
+> False
+:p all []:(Fin 0)=>Bool
+> True
+
+
 'Tests for numerical utilies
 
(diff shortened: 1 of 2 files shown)


dougalm:
> It feels a bit silly to add all these one-liners to the prelude
Not at all. This is canonical stuff. We could start making separate module files for more specialized functions though.

> evaluation on things like complex numbers
That reminds me: we should add a complex data type now that we have ADTs.

PR #171, opened 2020-07-24
Start allowing dependent data constructors

dougalm:
This lets us package together a table and its size, to make a list whose length is *not* exposed in its type.

data List a:Type = MakeList n:Int xs:(Fin n => a)

Getting it working was mostly about being careful with the nested scopes of binders in patterns and data constructor definitions. I strengthened the types of our IRs to make the book-keeping a bit easier.

It's still a WIP, because we can't build tables of such dynamic-sized lists, like `Fin 3 => List Int`, because our codegen assumes it can allocate the memory for a table before it starts to populate it. I'll try to fix that in a separate PR.


src/lib/Algebra.hs
```diff
@@ -53,12 +53,13 @@ data SumClampPolynomial = SumClampPolynomial ClampPolynomial Var deriving (Show,
...
   BaseTy _  -> liftC $ poly [(1, mono [])]
-  TabTy v _ -> (offsets t) `psubstSumVar` (indexSetSize $ varType v)
+  TabTy b _ -> (offsets t) `psubstSumVar` (indexSetSize $ binderType b)
   _ -> error $ "Not a ScalarTableType: " ++ pprint t
 
 offsets :: ScalarTableType -> SumClampPolynomial
 offsets t = case t of
-  TabTy v body -> sumC v $ elemCount body
+  -- TODO: not sure about `fromBind` here`
+  TabTy b body -> sumC (fromBind "_" b) $ elemCount body
```
apaszke, line 62: Well yeah... We should only do `sumC` if the binder is a regular variable (as it always used to be) and otherwise we should throw a hard error for now. I think that in the future a valid thing to do would be to iterate over all variables bound in the pattern and run `sumC` for every one of them, since summing over a product type is the same as doing multiple sums over its components.

src/lib/Env.hs
```diff
@@ -5,83 +5,92 @@
...
 infixr 7 :>
 
 newtype Env a = Env (M.Map Name a)
                 deriving (Show, Eq, Ord, Generic, Functor, Foldable, Traversable)
 
 -- TODO: consider parameterizing by namespace, for type-level namespace checks.
--- `NoName` is used in binders (e.g. `_ = <expr>`) but never in occurrences.
--- TODO: Consider putting it under a separate `Binder` type instead.
-data Name = Name NameSpace Tag Int | GlobalName Tag | GlobalArrayName Int | NoName
+data Name = Name NameSpace Tag Int | GlobalName Tag | GlobalArrayName Int
```
apaszke, line 32: Yay, no more `NoName`s!

src/lib/MDImp.hs
```diff
@@ -36,9 +34,13 @@ progToMD (ImpProg stmts) = MDImpProg $ evalCat $ forM stmts (\s@(d, _) -> (d,) <
 
 -- TODO: Collapse loops, hoist allocations, etc.
 stmtToMDInstr :: ImpStatement -> Translate (MDImpInstr ImpKernel)
-stmtToMDInstr (~(Just d), instr) = case traceShowId instr of
+stmtToMDInstr (d, instr) = case traceShowId instr of
```
apaszke, line 37: Oh wow I must have left some debugging statements 🤦

PR #179, opened 2020-07-29
[WIP] Remove syntactic ANF restriction on core IR

dougalm:
Our core IR makes a syntactic distinction between expressions (`Expr`) and atoms (`Atom`). Atoms include things like variables, lambdas, literals and constructor applications. They are guaranteed to be fully-reduced (even after substitution) and effect-free. This lets us substitute variables with atoms without ever worrying about duplicating runtime work.

The problem is that this syntactic definition of what's runtime-free is too conservative and we've found ourselves having to add new cases to `Atom` to represent little computations that we want to treat as cheap. We'd rather just be able to use the whole expression language. For example, we have the `SumCon` constructor to represent a value of type `Either a b`, given a boolean tag and values for the left and right cases: `SumCon tag x y`. We'd rather just use an expression like `Case tag of True -> Left x; False -> Right y`. That's cheap enough that we're happy to inline it anywhere. But it's hard to capture that cheapness syntactically.

It's a bigger problem for complicated tables backed by raw buffers, which is why we're making the change now. A table of lists (length-hidden tables) will be backed by a buffer containing the lengths of the lists, and the lists themselves. We can easily write a small expression that constructs such an object from the raw buffers using our ordinary expression language. But it's a stretch to do it purely using our language of atoms.

The plan is to just merge `Atom` into `Expr`, and work with an `isCheap :: Expr -> Bool` predicate that can be defined arbitrarily. Without Haskell's type system helping us, we're going to have to be extra careful to not duplicate work or reorder effects.

### DIFF
--- a/src/lib/Parser.hs
+++ b/src/lib/Parser.hs
@@ -162,7 +162,10 @@ exprAsModule :: UExpr -> (Name, UModule)
 exprAsModule e = (asGlobal v, UModule (toNest [d]))
   where
     v = mkName "_ans_"
-    d = ULet PlainLet (WithSrc (srcPos e) (UPatBinder (Bind (v:>()))), Nothing) e
+    d = sourceLet (WithSrc (srcPos e) (UPatBinder (Bind (v:>()))), Nothing) e
+
+sourceLet :: UPatAnn -> UExpr -> UDecl
+sourceLet = ULet (PlainLet SomeEffects)
 
 -- === uexpr ===
 
@@ -225,7 +228,8 @@ topDecl = dataDef <|> topLet
 
 topLet :: Parser UDecl
 topLet = do
-  lAnn <- (char '@' >> letAnnStr <* (void eol <|> sc)) <|> return PlainLet
+  lAnn <-  (char '@' >> letAnnStr <* (void eol <|> sc))
+       <|> return (PlainLet SomeEffects)
   ~(ULet _ (p, ann) rhs, pos) <- withPos decl
   let ann' = fmap (addImplicitImplicitArgs pos) ann
   return $ ULet lAnn (p, ann') rhs
@@ -274,7 +278,7 @@ simpleLet :: Parser (UExpr -> UDecl)
 simpleLet = label "let binding" $ do
   p <- try $ (letPat <|> parens pat) <* lookAhead (sym "=" <|> sym ":")
   ann <- optional $ annot uType
-  return $ ULet PlainLet (p, ann)
+  return $ sourceLet (p, ann)
 
 letPat :: Parser UPat
 letPat = nameAsPat $ upperName <|> lowerName <|> symName
@@ -289,7 +293,7 @@ funDefLet = label "function definition" $ mayBreak $ do
   let funTy = buildPiType piBinders eff ty
   let letBinder = (v, Just funTy)
   let lamBinders = flip map bs $ \(p,_, arr) -> ((p,Nothing), arr)
-  return $ \body -> ULet PlainLet letBinder (buildLam lamBinders body)
+  return $ \body -> sourceLet letBinder (buildLam lamBinders body)
   where
     arg :: Parser (UPat, UType, UArrow)
     arg = label "def arg" $ do
@@ -380,7 +384,7 @@ wrapUStatements statements = case statements of
   (s, pos):rest -> WithSrc pos $ case s of
     Left  d -> UDecl d $ wrapUStatements rest
     Right e -> UDecl d $ wrapUStatements rest
-      where d = ULet PlainLet (WithSrc pos (UPatBinder (Ignore ())), Nothing) e
+      where d = sourceLet (WithSrc pos (UPatBinder (Ignore ())), Nothing) e
   [] -> error "Shouldn't be reachable"
 
 uStatement :: Parser UStatement
--- a/src/lib/Type.hs
+++ b/src/lib/Type.hs
@@ -35,7 +35,10 @@ type TypeCheckEnv = OptionalEnv (TypeEnv, EffectRow)
 type TypeM = ReaderT TypeCheckEnv Except
 
 class Pretty a => HasType a where
-  typeCheck :: a -> TypeM Type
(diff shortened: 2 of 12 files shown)


duvenaud:
You've probably thought of this already, but if you have an ```isCheap :: Expr -> Bool``` function, why not a ```timeCost :: Expr -> Real```, ```memCost :: Expr -> Real```, cache-friendliness, etc.?  Such functions could also depend on the architecture.

src/lib/Parser.hs
```diff
@@ -162,7 +162,10 @@ exprAsModule :: UExpr -> (Name, UModule)
 exprAsModule e = (asGlobal v, UModule (toNest [d]))
   where
     v = mkName "_ans_"
-    d = ULet PlainLet (WithSrc (srcPos e) (UPatBinder (Bind (v:>()))), Nothing) e
+    d = sourceLet (WithSrc (srcPos e) (UPatBinder (Bind (v:>()))), Nothing) e
+
+sourceLet :: UPatAnn -> UExpr -> UDecl
+sourceLet = ULet (PlainLet SomeEffects)
```
apaszke, line 168: Will this `SomeEffects` annotation get later refined to `NoEffects` if the expr doesn't have them?

src/lib/Type.hs
```diff
@@ -162,57 +161,61 @@ isPure expr = case exprEffs expr of
...
     Linearize _ -> NoEffects
     Transpose _ -> NoEffects
-    -- These are more convservative than necessary.
-    -- TODO: make them more precise by de-duping with type checker
     RunReader _ _ -> SomeEffects
     RunWriter _   -> SomeEffects
     RunState _ _  -> SomeEffects
-    Tile _ _ _ -> error "not implemented"
-  Case _ _ _ -> error "not implemented"
+    Tile _ _ _ -> SomeEffects
```
apaszke, line 183:
```suggestion
    Tile _ t s -> functionEffs t <> functionEffs s
```

dougalm:
I'm abandoning this change because it's trickier than expected and I think there are simpler ways to achieve the same goals. It turns out we rely a lot on the ANF restriction during defunctionalization and imp lowering. This PR passes tests but only because *we're still observing the ANF restriction in practice*. I think it'll become a bug farm when that changes.

### Here's what went wrong

1. We'd been taking for granted an invariant that GHC calls the "trivial-constructor-argument invariant" (described in "Secrets of the GHC inliner"): constructor arguments should be cheap enough to evaluate that we don't mind inlining them. This was enforced syntactically, because constructor arguments were `Atom`s. It's especially important in our simplification/defunctionalization pass. We want to inline constructors so that we can match them with destructors. In GHC it's an optimization, but for us it's mandatory! If we have a term like `((), \x. x)` we *must* inline it so that we can find the call site of the lambda. But if we have non-cheap arguments, like `(sum xs, \x. x)` then we can't inline it without duplicating work.

2. We can no longer assume that an atom of a certain type is either a variable or the data constructor for that type. We rely on that invariant when we apply `fromScalarAtom`, and when we do beta reduction during simplification.

3. Without the ANF constraint we've lost the syntactic guarantee that subexpressions can't have effects, so we have to be much more careful about inlining.

4. Something that I expected to be a benefit actually turned out not to be. Without the ANF constraint, I can just write `Add x (Add y z)` instead of `do { tmp <- emit (Add y z); emit (Add x tmp) }`. But the old style of making a let binding for everything was actually less taxing, because I never had to reason about the risk of duplication.

### Here's what we'd been hoping to achieve with this change, and how we might do it within ANF

1. We want to carry arbitrary expressions (especially case expressions) in substitutions.

In defunctionalization and Imp lowering, we separate complex expressions into a simple-but-expensive part, representing the numerical work, and a complex-but-cheap part (like a lambda) that we inline downstream. Until now, complex-but-cheap things have been represented as atoms. But, now that we have case expressions, that's no longer possible. A case expression like this

```
f = case x of
      True  ->
        y = ...
        \x. x + y
      False ->
        z = ...
        w = ...
        \x. x * z + w
```

Will become this


```
fData = case x of
  True  ->
    y = ...
    Left y
  False ->
    z = ...
    w = ...
    Right (z, w)
```

And we'll carry a substitution that substitutes `f` with:

```
case fData of
  Left y       -> \x. x + y
  Right (z, w) -> \x. x * z + w
```

But we don't need ANF to do this sort of thing. We can just add a `case` case to atoms. Adam points out that this can be something like `MiniCase Var [Alt]`, with a variable occurrence for a scrutinee, which means that it will never be able to appear in closed expressions (values). We'll select the appropriate case during substitution if we have a concrete constructor for the scrutinee variable.

2 . Make certain simplifications easier

For example, if we had:

```
xs = for i. <e1>
...  xs.i ...
```

We want to be able to inline the definition of `xs` and then beta reduce. With the ANF restriction, we can't inline `xs = ...` because it's not an atom, so we have to do the beta reduction more indirectly, via a lookup. It's annoying but not that hard.


3. Express (lambda-bound) type applications in types

A function like Haskell's `fmap` would have a type

`fmap : Functor f ?=> (a -> b) -> f a -> f b`

We currently can't express this because we can't have function application (which is non-atomic) in types. I don't have a good solution to this one. We don't expect to be substituting `f` with an arbitrary lambda term. It's just going to be a type constructor. We need to think harder about restrictions on functions that produce types.


4. A more readable IR

Normalization makes curried application very verbose, with one line per application. Before this PR, normalization turns something as simple as `1 + 1` into a five-liner!

```
  tmp:((Add Int) ?=> Int -> Int -> Int) = (+) Int
  tmp1:(Int -> Int -> Int) = tmp _
  tmp2:(Int -> Int) = tmp1 1
  tmp3:Int = tmp2 2
  _ans_:Int = tmp3
```

With this PR, it's just:

```
  _ans_:Int = (+) Int _ 1 2
```

It's also faster to compile because it's just less code (the types for those partially applied functions take up a lot of space!). *But* we could achieve the same thing if we just had an n-ary function application construct, `App Atom [Atom]`, in `Expr`.

PR #180, opened 2020-07-30
Non-extensible records and variants

hexahedria:
This PR adds support for non-extensible records and variants, with optionally repeated labels.

Syntax-wise, records look like
```
x : {a: Int & b: Real} = {a = 3, b = 4.0}
```
and variants look like
```
x : {a: Int | b: Real} = {| a = 3 |}
```

I used `&` and `|` for parity with tuples and Either. Empty records are denoted as `x : {&} = {}`. Technically there's an empty variant type `{|}` but it is uninhabited.

One interesting property of the implementation is that you are allowed to repeat labels. This was originally designed to support easy extensible records (see Daan Leijen's "Extensible records with scoped labels") but I don't think it causes any harm to include them in the non-extensible case. Perhaps this will make it easier to add extensibility in the future.

Records can be unpacked using pattern-matching let bindings, and variants can be unpacked using a case statement. For now, variants need to be explicitly annotated with their type unless it is sufficiently obvious (to the typechecker) from the preceding code what the type is. You can freely put either records or variants into tables, and hopefully can combine them in arbitrary ways.

I've implemented some preliminary logic for treating records and variants as index sets, since records have a nice interpretation as named axes, and variants as named slices/concatenations. However, so far I haven't figured out how to plumb everything through the low level internals. I got far enough that you can create tables indexed by variants, but once you have one of those tables, you can't print it or get anything out of it without getting a bunch of errors. Hopefully this is easy to fix for someone with a better mental model of how Imp and Interpreter work.

Some things I wasn't sure about:

- At some points I needed a binder, but I didn't have one, so I used `Ignore ty` even though I wasn't necessarily going to ignore it. There's probably a better way of doing that, though.
- I sort of hacked together the `Imp` translation rules, so it's very possible that I missed something.
- There were some shenanigans I had to do to get the index set logic working, in particular generating case statements and casting things to `PreludeBoolTy` to enable pattern matching on them. I feel like there's probably something simpler than that? Although I suppose it's not too terrible.

Eventually, we may want to make the records and variants extensible, so that you could write something like `{a=x, b=y, ...c}` to add fields `a` and `b` to a record `c`. This could enable some cool polymorphic helper functions or other interesting use cases, and would also make it so variants don't have to be annotated with a type (since we could infer some extensible type and then let unification figure out what the real type was).

### DIFF
--- a/src/lib/Embed.hs
+++ b/src/lib/Embed.hs
@@ -97,8 +97,15 @@ emitOp op = emit $ Op op
 
 emitUnpack :: MonadEmbed m => Expr -> m [Atom]
 emitUnpack expr = do
-  let (TypeCon def params) = getType expr
-  let [DataConDef _ bs] = applyDataDefParams def params
+  bs <- case getType expr of
+    TypeCon def params -> do
+      let [DataConDef _ bs] = applyDataDefParams def params
+      return bs
+    RecordTy types -> do
+      -- TODO: is using Ignore here appropriate? We don't have any existing
+      -- binders to bind, but we still plan to use the results.
+      let bs = toNest $ map Ignore $ toList types
+      return bs
   expr' <- deShadow expr <$> getScope
   vs <- freshNestedBinders bs
   embedExtend $ asSnd $ Nest (Unpack (fmap Bind vs) expr') Empty
--- a/src/lib/Embed.hs
+++ b/src/lib/Embed.hs
@@ -581,6 +581,11 @@ traverseAtom def@(_, fAtom) atom = case atom of
   DataCon dataDef params con args -> DataCon dataDef <$>
     traverse fAtom params <*> pure con <*> traverse fAtom args
   TypeCon dataDef params -> TypeCon dataDef <$> traverse fAtom params
+  Record items -> Record <$> traverse fAtom items
+  RecordTy items -> RecordTy <$> traverse fAtom items
+  Variant types label i value -> Variant <$>
+    traverse fAtom types <*> pure label <*> pure i <*> fAtom value
+  VariantTy items -> VariantTy <$> traverse fAtom items
 
 -- === DCE ===
 
--- a/src/lib/Embed.hs
+++ b/src/lib/Embed.hs
@@ -35,6 +35,7 @@ import Control.Monad.Reader
 import Control.Monad.Writer hiding (Alt)
 import Control.Monad.Identity
 import Data.Foldable (toList)
+import qualified Data.Map.Strict as M
 import Data.Maybe
 import GHC.Stack
 
@@ -43,7 +44,7 @@ import Syntax
 import Cat
 import Type
 import PPrint
-import Util (bindM2)
+import Util (bindM2, traverseFunM, restructure)
 
 newtype EmbedT m a = EmbedT (ReaderT EmbedEnvR (CatT EmbedEnvC m) a)
   deriving (Functor, Applicative, Monad, MonadIO, MonadFail, Alternative)
@@ -263,6 +264,10 @@ getSnd :: MonadEmbed m => Atom -> m Atom
 getSnd (PairVal _ y) = return y
 getSnd p = emitOp $ Snd p
 
+getUnpacked :: MonadEmbed m => Atom -> m [Atom]
+getUnpacked (Record items) = return $ toList items
+getUnpacked a = emitUnpack (Atom a)
+
 app :: MonadEmbed m => Atom -> Atom -> m Atom
 app x i = emit $ App x i
 
@@ -405,6 +410,9 @@ unsafeIntToBool b = emitOp $ ScalarUnOp UnsafeIntToBool b
 
 indexAsInt :: MonadEmbed m => Atom -> m Atom
 indexAsInt idx = emitOp $ IndexAsInt idx
+
+unsafeCoerce :: MonadEmbed m => Type -> Atom -> m Atom
+unsafeCoerce ty x = emitOp $ CoerceOp ty x
 
 instance MonadTrans EmbedT where
   lift m = EmbedT $ lift $ lift m
@@ -698,6 +706,12 @@ indexSetSizeE (TC con) = case con of
   SumType l r -> bindM2 iadd (indexSetSizeE l) (indexSetSizeE r)
   _ -> error $ "Not implemented " ++ pprint con
   where
+indexSetSizeE (RecordTy types) = do
+  sizes <- traverse indexSetSizeE types
+  foldM imul (IntVal 1) sizes
+indexSetSizeE (VariantTy types) = do
+  sizes <- traverse indexSetSizeE types
+  foldM iadd (IntVal 0) sizes
 indexSetSizeE ty = error $ "Not implemented " ++ pprint ty
 
 clampPositive :: MonadEmbed m => Atom -> m Atom
@@ -727,6 +741,25 @@ indexToIntE ty idx = case ty of
     imul rSize lIdx >>= iadd rIdx
   TC (IntRange _ _)     -> indexAsInt idx
   TC (IndexRange _ _ _) -> indexAsInt idx
+  RecordTy types -> do
+    sizes <- traverse indexSetSizeE types
+    (strides, _) <- traverseFunM
+      (\sz prev -> do {v <- imul sz prev; return (prev, v)}) sizes (IntVal 1)
+    -- Unpack and sum the strided contributions
+    subindices <- getUnpacked idx
+    subints <- traverse (uncurry indexToIntE) (zip (toList types) subindices)
+    scaled <- mapM (uncurry imul) $ zip (toList strides) subints
+    foldM iadd (IntVal 0) scaled
+  VariantTy types -> do
+    sizes <- traverse indexSetSizeE types
+    (offsets, _) <- traverseFunM
+      (\sz prev -> do {v <- iadd sz prev; return (prev, v)}) sizes (IntVal 0)
+    -- Build and apply a case expression
+    alts <- flip mapM (zip (toList offsets) (toList types)) $
+      \(offset, subty) -> buildNAbs (toNest [Ignore subty]) $ \[subix] -> do
+        i <- indexToIntE subty subix
+        iadd offset i
+    emit $ Case idx alts IntTy
   _ -> error $ "Unexpected type " ++ pprint ty
 
 intToIndexE :: MonadEmbed m => Type -> Atom -> m Atom
@@ -748,6 +781,37 @@ intToIndexE ty@(TC con) i = case con of
     return $ Con $ SumCon isLeft li ri
   _ -> error $ "Unexpected type " ++ pprint con
   where iAsIdx = return $ Con $ Coerce ty i
+intToIndexE (RecordTy types) i = do
+  sizes <- traverse indexSetSizeE types
+  (strides, _) <- traverseFunM
+    (\sz prev -> do {v <- imul sz prev; return ((prev, v), v)}) sizes (IntVal 1)
+  offsets <- flip mapM (zip (toList types) (toList strides)) $
+    \(ty, (s1, s2)) -> do
+      x <- irem i s2
+      y <- idiv x s1
+      intToIndexE ty y
+  return $ Record (restructure offsets types)
+intToIndexE rty@(VariantTy types) i = do
+  sizes <- traverse indexSetSizeE types
+  (offsets, _) <- traverseFunM
+    (\sz prev -> do {v <- iadd sz prev; return (prev, v)}) sizes (IntVal 0)
+  let
+    reflect = reflectLabels types
+    -- Find the right index by looping through the possible offsets
+    -- TODO: is there a better way than generating a case statement? Select
+    -- doesn't work because variants aren't primitive types
+    go prev ((label, repeatNum), ty, offset) = do
+      beforeThis <- unsafeCoerce PreludeBoolTy =<< ilt i offset
+      hereCase <- buildNAbs Empty $ \[] -> do
+        shifted <- isub i offset
+        index <- intToIndexE ty shifted
+        return $ Variant types label repeatNum index
+      beforeCase <- buildNAbs Empty $ \[] -> return $ prev
+      emit $ Case beforeThis [hereCase, beforeCase] rty
+    ((l0, 0), ty0, _):zs = zip3 (toList reflect) (toList types) (toList offsets)
+  start <- Variant types l0 0 <$> intToIndexE ty0 i
+  choice <- foldM go start zs
+  return choice
 intToIndexE ty _ = error $ "Unexpected type " ++ pprint ty
 
 anyValue :: Type -> Atom
--- a/src/lib/Imp.hs
+++ b/src/lib/Imp.hs
@@ -24,6 +24,7 @@ import Control.Monad.Writer hiding (Alt)
 import Data.Text.Prettyprint.Doc
 import Data.Foldable (toList)
 import Data.Coerce
+import qualified Data.Map.Strict as M
 
 import Embed
 import Array
@@ -33,6 +34,8 @@ import Type
 import PPrint
 import Cat
 import qualified Algebra as A
+import Util
+
 -- Note [Valid Imp atoms]
 --
 -- The Imp translation functions as an interpreter for the core IR, which has a side effect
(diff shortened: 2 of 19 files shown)


src/lib/Embed.hs
```diff
@@ -97,8 +98,15 @@ emitOp op = emit $ Op op
...
-  let (TypeCon def params) = getType expr
-  let [DataConDef _ bs] = applyDataDefParams def params
+  bs <- case getType expr of
+    TypeCon def params -> do
+      let [DataConDef _ bs] = applyDataDefParams def params
+      return bs
+    RecordTy types -> do
+      -- TODO: is using Ignore here appropriate? We don't have any existing
+      -- binders to bind, but we still plan to use the results.
+      let bs = toNest $ map Ignore $ toList types
```
apaszke, line 108: That should be fine I think. The names are not used below for anything else than a name hint. The types are what's important.

src/lib/Embed.hs
```diff
@@ -256,6 +264,10 @@ getSnd :: MonadEmbed m => Atom -> m Atom
 getSnd (PairVal _ y) = return y
 getSnd p = emitOp $ Snd p
 
+getUnpacked :: MonadEmbed m => Atom -> m [Atom]
+getUnpacked (Record items) = return $ toList items
+getUnpacked a = emitUnpack (Atom a)
```
apaszke, line 269: Well single-case ADTs can also be unpacked, so this might end up being a little confusing.

apaszke: And it seems like there's only a single use-site, so it's better to inline this.

src/lib/Embed.hs
```diff
@@ -715,6 +741,25 @@ indexToIntE ty idx = case ty of
     imul rSize lIdx >>= iadd rIdx
   TC (IntRange _ _)     -> indexAsInt idx
   TC (IndexRange _ _ _) -> indexAsInt idx
+  RecordTy types -> do
```
apaszke, line 744: BTW is there any reason to keep the record type as an n-ary constructor instead of a binary one, like we do for tuples?

src/lib/Embed.hs
```diff
@@ -715,6 +741,25 @@ indexToIntE ty idx = case ty of
...
+    sizes <- traverse indexSetSizeE types
+    (strides, _) <- traverseFunM
+      (\sz prev -> do {v <- imul sz prev; return (prev, v)}) sizes (IntVal 1)
```
apaszke, line 747: Hmm one interesting question to ask would be whether we want to support dependent record types, and if the answer is yes, then this code will need to be re-worked a little.

src/lib/Util.hs
```diff
@@ -174,13 +175,20 @@ findReplace old new s@(x:xs) =
 traverseFun :: Traversable t => (a -> s -> (b, s)) -> t a -> s -> (t b, s)
 traverseFun f xs s = runState (traverse (asState . f) xs) s
 
-asState :: (s -> (a, s)) -> State s a
-asState f = do
+traverseFunM :: (Monad m, Traversable t)
+  => (a -> s -> m (b, s)) -> t a -> s -> m (t b, s)
+traverseFunM f xs s = runStateT (traverse (asStateT . f) xs) s
```
apaszke, line 180: Wouldn't `scanM` be a better name?

src/lib/Syntax.hs
```diff
@@ -766,6 +810,10 @@ instance Eq Atom where
   DataCon def params con args == DataCon def' params' con' args' =
     def == def' && params == params' && con == con' && args == args'
   TypeCon def params == TypeCon def' params' = def == def' && params == params'
+  Record lr == Record lr' = lr == lr'
+  Variant lr l i v == Variant lr' l' i' v' = (lr, l, i, v) == (lr', l', i', v')
+  RecordTy lr == RecordTy lr' = lr == lr'
+  VariantTy lr == VariantTy lr' = lr == lr'
```
apaszke, line 816: nit: can you please ad some alignment, because right not this is very hard to read

src/lib/Syntax.hs
```diff
@@ -71,6 +72,10 @@ data Atom = Var Var
           | Pi  PiType
           | DataCon DataDef [Atom] Int [Atom]
           | TypeCon DataDef [Atom]
+          | Record (LabeledItems Atom)
+          | RecordTy (LabeledItems Type)
+          | Variant (LabeledItems Type) Label Int Atom
```
apaszke, line 77: I wonder how is it possible that you were able to get away without something a'la `SumAsProd`? Are there any tests where the chosen variant depends on some run-time information?

src/lib/Simplify.hs
```diff
@@ -168,6 +176,11 @@ simplifyExpr expr = case expr of
       DataCon _ _ con args -> do
         let Abs bs body = alts !! con
         extendR (newEnv bs args) $ simplifyBlock body
+      Variant types label i value -> do
+        let LabeledItems ixtypes = enumerate types
+        let index = fst $ ixtypes M.! label !! i
```
apaszke, line 181: nit: can you please parenthesize the rhs of the `$`? It is unclear whether it's `M.!` or `!!` that takes precedence

src/lib/Parser.hs
```diff
@@ -816,3 +872,6 @@ eolf = void eol <|> eof
 failIf :: Bool -> String -> Parser ()
 failIf True s = fail s
 failIf False _ = return ()
+
+_debug :: Show a => String -> Parser a -> Parser a
+_debug s m = mapReaderT (Text.Megaparsec.Debug.dbg s) m
```
apaszke, line 877: what is this for?

src/lib/Imp.hs
```diff
@@ -588,12 +608,20 @@ zipWithDest dest@(Dest destAtom) atom f = case (destAtom, atom) of
...
+        zipWithM_ rec (map Dest (toList items)) (toList items')
   -- TODO: Check array type?
   (Var _, Var _)       -> f (fromArrayAtom destAtom) (fromScalarAtom atom)
   (Var _, Con (Lit _)) -> f (fromArrayAtom destAtom) (fromScalarAtom atom)
   (Con (SumAsProd _ tag payload), DataCon _ _ con x) -> do
     recDest tag (IntVal con)
     zipWithM_ recDest (payload !! con) x
+  (Con (SumAsProd _ tag payload), Variant types label i x) -> do
+    let LabeledItems ixtypes = enumerate types
+    let index = fst $ ixtypes M.! label !! i
```
apaszke, line 622: nit: Please add some parenthesis (as noted in the other place)

src/lib/Embed.hs
```diff
@@ -736,6 +781,37 @@ intToIndexE ty@(TC con) i = case con of
...
+    -- TODO: is there a better way than generating a case statement? Select
+    -- doesn't work because variants aren't primitive types
+    go prev ((label, repeatNum), ty, offset) = do
+      beforeThis <- unsafeCoerce PreludeBoolTy =<< ilt i offset
+      hereCase <- buildNAbs Empty $ \[] -> do
+        shifted <- isub i offset
+        index <- intToIndexE ty shifted
+        return $ Variant types label repeatNum index
+      beforeCase <- buildNAbs Empty $ \[] -> return $ prev
+      emit $ Case beforeThis [hereCase, beforeCase] rty
```
apaszke, line 810: This looks like a good candidate for a standalone function like

```hs
boolCase :: MonadEmbed m => Atom -> m Atom -> m Atom -> m Atom
boolCase intCond trueBlock falseBlock = ...
```

src/lib/Embed.hs
```diff
@@ -736,6 +781,37 @@ intToIndexE ty@(TC con) i = case con of
...
+    ((l0, 0), ty0, _):zs = zip3 (toList reflect) (toList types) (toList offsets)
```
apaszke, line 811: You ignore the first offset anyway, so why is it zipped?

src/lib/Embed.hs
```diff
@@ -736,6 +781,37 @@ intToIndexE ty@(TC con) i = case con of
...
+  start <- Variant types l0 0 <$> intToIndexE ty0 i
```
apaszke, line 812: Isn't this just `intToIndexE rty 0`? I understand if you want to make it more efficient, but it might deserve a comment

src/lib/Embed.hs
```diff
@@ -736,6 +781,37 @@ intToIndexE ty@(TC con) i = case con of
...
+  choice <- foldM go start zs
+  return choice
```
apaszke, line 814: nit: no need to have the trailing return

src/lib/Imp.hs
```diff
@@ -480,6 +495,11 @@ makeDest nameHint destType = do
               let dcs' = applyDataDefParams def params
               contents <- forM dcs' $ \(DataConDef _ bs) -> forM (toList bs) (rec . binderType)
               return $ Con $ SumAsProd ty tag contents
+        RecordTy types -> Record <$> forM types rec
+        VariantTy types -> do
+          tag <- rec IntTy
+          contents <- forM (toList types) rec
+          return $ Con $ SumAsProd ty tag $ map (\x->[x]) contents
```
apaszke, line 502: This looks incorrect to me, because it doesn't preserve the type!

src/lib/Embed.hs
apaszke: Also, IMO the case to `PreludeBoolTy` is ok!

src/lib/Embed.hs
hexahedria: I think the fallback case makes it so that this will still work for ADTs, but it just doesn't have the opportunistic optimization (which would be easy to add). Although I'm happy to just inline this. Wasn't sure what the bar was for adding helper functions to Embed (maybe it's two or three use-sites?).

src/lib/Embed.hs
hexahedria: By binary constructor do you mean something like `RecordTy Label Type Type`, and expressing records as something like `RecordTy "foo" x (RecordTy "bar" y (Unit))`? The main reasons I went with this over that are:
- I intended the types `{a:Int & b:Real}` and `{b:Real & a:Int}` to be the same. If we used a binary constructor, we would have to worry about the labels not being in the right order, and possibly need to do list sorting to check that things match. Using a single n-ary constructor lets you easily enforce a single order internally while allowing arbitrary orders in the parsing and typechecking logic.
- The "rest" argument in a binary constructor needs to be another `RecordTy` instance, but I don't think the type system can support that. For polymorphic extensible records we probably need to introduce a new kind for a row of types, but that didn't seem worth the effort in the short term. (Tuples don't have this problem because `(a, b)` is a valid type no matter what `b` is, but `{foo:a, ...b}` only makes sense if `b` is a record type.)

src/lib/Embed.hs
hexahedria: Dependent in the sense that one record field's type depends on the value from another field? I wonder if that would be useful. I'm not sure how you'd even write the type for that, since at least right now the labels of the record don't actually bind the label names to anything (although they kind of look like they do, I guess, which maybe is confusing), and there's no syntactical ordering imposed on the labels (so we can't say that earlier things are "in scope" for later things, because that depends on how the user wrote the type down).

My gut feeling is that if you want dependent things you should define an ADT, at least for now.

src/lib/Util.hs
hexahedria: Sounds reasonable to me. Should I change `traverseFun` to `scan` too?

src/lib/Syntax.hs
hexahedria: Maybe you noticed this later, but I use the existing `SumAsProd` as-is. I think some of my tests exercise that functionality when they loop over a table containing variants.

src/lib/Parser.hs
hexahedria: I kept finding myself debugging weird megaparsec error messages, so I lifted out the debug utility from megaparsec to help with them. You can wrap a parser with this (i.e. `_debug "foo" foo` instead of just `foo`) and it prints out a bunch of diagnostic info with `Debug.Trace` whenever that wrapped parser succeeds or fails. I found it helpful, so I thought maybe it was worth merging in in case it's needed later.

src/lib/Embed.hs
hexahedria: I ignore the first offset because I know it's exactly 0. I don't ignore the rest of the offsets (see `go`), and I still wanted them to be aligned correctly in `zs`.

src/lib/Embed.hs
hexahedria: Sorry, I don't follow? This is in the definition of `intToIndexE rty i` so I don't think I want to recursively call `intToIndexE rty` here. Also the offset isn't necessarily 0. This line is essentially saying "if the first variant is the correct variant to use, what would the index atom be", and then the `foldM` overrides this if it wasn't actually the first variant.

Maybe I'm not understanding the suggestion?

src/lib/Imp.hs
hexahedria: It saves the type into the `SumAsProd` constructor, which I think preserves the type for any downstream usage? At least the typechecker doesn't complain about it.

hexahedria:
I think it would definitely be possible to represent tuples as records internally (and @dougalm and I actually discussed this briefly at one point). Although that would probably change the semantics of the user-level language too. Right now `((a, b), c)` and `(a, b, c)` are the exact same thing, but if they were records internally then those would be different types `{v: {v:a, v:b}, v:c}` and `{v:a, v:b, v:c}`. I guess this would make it closer to Haskell where `(,,)` is different than `(,)`.

Similarly, if we represented anonymous sum types with variants, this would change how you match on them, since there's no such thing as `Left` or `Right` for a three-element variant. So the user-level pattern-matching syntax might have to change. (Although I guess we already sort of have two notions of sum types, the one defined as an ADT and the internal one.)

I guess a halfway thing that would be pretty easy would be to still require the tuple and sum constructors to be binary, so that `a & b` becomes `{left:a & right:b}` and `a | b` becomes `{left:a | right:b}`? But then I'm not sure we gain a lot.

src/lib/Embed.hs
hexahedria: I guess you still need to pass in the return type, right? Since that gets passed to `Case`. (Or maybe there's a way to easily infer that from the `trueBlock` and `falseBlock`? I guess we could wrap the case in a `Lam` and then check its type, but that seems a bit annoying...)

hexahedria: hm, actually, maybe I can just `getType body` for the body of the `Abs`?

src/lib/Embed.hs
dougalm: I agree that we should only do dependent constructors via ADTs for now. For a general dependent pair, you need some sort of annotation to indicate what the type of the second element is (e.g. `(2, [1,2])` could have type `(Int & (Fin 2 => Int))` or `(n:Int & (Fin n => Int))`. Extensible records are tricky enough as they are!

src/lib/Embed.hs
dougalm: Yep, you can just query the type. (`getType` doesn't need to know about the variables in scope since their types are cached at the occurrences.)

src/lib/Imp.hs
dougalm: `SumAsProd` was intended for ADTs (and if it passes the type checker then that was probably an oversight on my part). But this seems like a fine use of it too. Btw, we're going to remove it soon in favor of a `Minicase Var [AtomAlt]` atom, as mentioned in #179.

dougalm:
> Although I guess we already sort of have two notions of sum types, the one defined as an ADT and the internal one.
Yes, the plan is to remove the built-in `Either`. What's blocking it is that we don't have index set instances for ADTs (either built-in or user-definable).

dougalm:
Look great!

As we discussed, the variants are very neat, but they *might* be too much complexity, since we can always define sums with ADTs. We can make that call once we have more experience using them.

I'm looking forward to the extensible version. That will let us pull out a single field by name rather than pattern-matching on all the fields, right? Unfortunately, we already use the conventional `.` for table indexing. Maybe `myRecord#myField`? More radically, we could just use whitespace for table indexing (since it's actually just function application underneath) and then `.` would be free for record indexing.

dougalm:
It just occurred to me that an immediate application of these records would be to make the typeclass dictionaries a bit more readable:

```
data Add a:Type =
  MkAdd { add : (a->a->a) 
        & sub : (a->a->a) 
        & zero : a }
```

We could also add conventional `class ... where ...` and `instance ... where ...` syntax rather than this awkward explicit dictionary-construction business.

src/lib/Parser.hs
apaszke: Nice, thanks for explaining that

PR #184, opened 2020-07-31
Records and variants: minor follow up changes

hexahedria:
Addresses (most of) @apaszke's comments from #180.

### DIFF
--- a/src/lib/Embed.hs
+++ b/src/lib/Embed.hs
@@ -35,7 +35,6 @@ import Control.Monad.Reader
 import Control.Monad.Writer hiding (Alt)
 import Control.Monad.Identity
 import Data.Foldable (toList)
-import qualified Data.Map.Strict as M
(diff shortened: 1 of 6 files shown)


src/lib/Embed.hs
```diff
@@ -752,7 +760,7 @@ indexToIntE ty idx = case ty of
     foldM iadd (IntVal 0) scaled
   VariantTy types -> do
     sizes <- traverse indexSetSizeE types
-    (offsets, _) <- traverseFunM
+    (offsets, _) <- scanM
       (\sz prev -> do {v <- iadd sz prev; return (prev, v)}) sizes (IntVal 0)
```
apaszke, line 764: Oh a nice trick that might simplify the body of the lambda and let you avoid folding the line would be this:
```hs
\sz prev -> (prev,) <$> iadd sz prev
```

PR #196, opened 2020-08-11
Expose fixed-width types and remove the built-in Bool type

apaszke:
The built-in Bool type is replaced with Int8, which removes the awkward
compiler-builtin cast from `Bool` to `PreludeBool`. I also generalized
casts to allow conversion between all integer types and all
floating-point types (including int-to-float and float-to-int).


src/lib/Embed.hs
```diff
@@ -783,21 +756,19 @@ intToIndexE (VariantTy types) i = do
...
+      -- TODO: This might run intToIndex on negative indices. Fix this!
+      index   <- intToIndexE ty shifted
       beforeThis <- ilt i offset
-      boolCase beforeThis (return prev) $ do
-        shifted <- isub i offset
-        index <- intToIndexE ty shifted
-        return $ Variant types label repeatNum index
-    ((l0, 0), ty0, _):zs =
-      zip3 (toList reflect) (toList types) (toList offsets)
+      select beforeThis prev $ Variant types label repeatNum index
```
hexahedria, line 763: Does this work? I thought I tried something like this and ran into problems because the `Select` op implementation assumed its inputs were basic types, and Variants are not basic types. Maybe something else in this PR fixes that too? (edit: I think I understand now, see below)

src/lib/Imp.hs
```diff
@@ -216,23 +216,19 @@ toImpOp (maybeDest, op) = case op of
...
+    (_, BaseTy bt) -> castTo bt
+    (_, IntTy    ) -> castTo $ Scalar Int64Type
+    (_, FloatTy  ) -> castTo $ Scalar Float64Type
     _ -> error $ "Invalid cast: " ++ pprint (getType x) ++ " -> " ++ pprint destTy
     where
       castTo bt = do
-        result <- emitInstr (ICastOp (IValType bt) $ fromScalarAtom x)
+        result <- cast (fromScalarAtom x) bt
         returnVal $ toScalarAtom destTy result
+  Select p x y -> do
```
hexahedria, line 227: Ah, answering my own question, I think. Do I understand right that this makes it possible to `Select` on arbitrary types?

apaszke: Yes, exactly! It's a little sub-optimal for those simple types right now, but as far as I can see we really only use `Select` to pick between the prelude bool values.

PR #204, opened 2020-08-19
Add an initial version of the multicore backend

apaszke:
Turns out, that having MDImp in place, we can reuse a lot of the
CUDA-related infrastructure for generating multicore CPU code. With this
patch (and some hacks to make MDImp less restrictive) the ray tracer
example takes 98s to generate a 250x250 image using 16 threads, as opposed
to 584s when running with the single-threaded backend.

### DIFF
--- a/dex.cabal
+++ b/dex.cabal
@@ -1,10 +1,11 @@
+cabal-version:       2.2
+
 -- Copyright 2019 Google LLC
 --
 -- Use of this source code is governed by a BSD-style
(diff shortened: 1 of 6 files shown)


dougalm:
@jekbradbury convinced me to get myself a 16-core Ryzen 3950X for my birthday. With this patch, I'm seeing a 15.8X speedup on the ray tracing example :smile: :smile:

PR #205, opened 2020-08-20
Exploratory: isomorphism optics

danieldjohnson:
This is a first pass at applying functional optics (e.g. lenses and prisms) to extensible records and variants. I think something like this would be a cool way to support both record field getter functions (for concrete record values) and named axis reductions/slicing (for tables that are indexed by records and variants).

Instead of using a profunctor representation (which Dex can't do right now because you can't write types like "for any functor `f` this is a `f Int`), this uses an isomorphism representation: a lens is seen as a transformation to a tuple of what you want to look at and what you ignore, and a prism is a transformation to a sum type of what you want and what else you might have. The representation I'm actually using supports arbitrary compositions of lenses and prisms by saying that an optic is a transformation from a type `full` to a type `(focus & ignore) | alt`, i.e. you either get what you want and something you ignore, or you get some alternative thing. (Technically, there are types of optics that don't fit in this representation, but I'm ignoring those for now.)

Another difference is that I expose the types `ignore` and `alt` in the type of the optic. Usually, these are existentially quantified over and thus hidden from the top-level type, but exposing them has some neat benefits in that it allows you to combine optics in different ways (for instance, you can look at two things using `and`, or invert a lens using `except` to ignore a specific thing instead of focusing on it).

I was able to write this in pure Dex, taking advantage of the extensible records and variants from #201. If this seems like a good set of abstractions, we could add syntactic sugar in the compiler for the primitive record and variant accessor functions, which would remove some of the boilerplate (i.e. for `fieldFoo`, `fieldBar`, `altFoo`, `altBar`).

### DIFF
--- a/examples/optics-helper.dx
+++ b/examples/optics-helper.dx
@@ -1,0 +1,20 @@
+
+Void = {|}
+
+data Uninhabited a:Type = MkUninhabited (a -> b:Type -> b)
+
+def absurd (res : Type)?-> (a : Type)?-> (proof: Uninhabited a)?=> (contradiction: a) : res =
+  (MkUninhabited doit) = proof
+  doit contradiction res
+
+def promiseUninhabited (a:Type)?-> : Uninhabited a = MkUninhabited (\v. todo)
+
+@instance
+voidUninhabited : Uninhabited Void = promiseUninhabited
+
+@instance
+def sumUninhabited (_:Uninhabited a)?=> (_:Uninhabited b)?=> : Uninhabited (a|b) = promiseUninhabited
+
+-- note: can't define instance for both left and right!
+@instance
+def pairLeftUninhabited (_:Uninhabited a)?=> : Uninhabited (a&b) = promiseUninhabited
--- a/examples/optics.dx
+++ b/examples/optics.dx
@@ -1,0 +1,216 @@
+
+include "examples/optics-helper.dx"
+
+'## Basic definitions
+
+data DataOptic focus:Type full:Type ignore:Type alt:Type =
+  MkDataOptic { split : full -> ((focus & ignore) | alt)
+              & build : ((focus & ignore) | alt) -> full
+              }
+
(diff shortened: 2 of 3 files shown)


hexahedria:
Yeah, I agree that it's a bit confusing. I've been thinking maybe it would make sense to introduce special syntax for record/variant lenses/prisms that would make it easier to express common slicing and indexing patterns, maybe something like
```
sum $ slice ?{firstpart | secondpart} whole
```
in this case (where `?{firstpart | secondpart}` is a prism from `{firstpart:a | secondpart:b | ...}` to `{firstpart:a | secondpart:b}`).

(Separately, I'm also not sure there are really any compelling use cases for compositions of lenses and prisms in array-based code, so maybe having separate types for lenses and prisms would make sense.)

Agreed that there's some duplication between records, tuples, and ADTs. There's been some discussion that we could unify records and tuples internally, since indeed a record with a single repeated field should be equivalent to a tuple. At least right now there are things ADTs can do that records/variants can't, such as dependent pairs, but maybe it would make sense to unify those too at some point? I'm not sure.

apaszke (approved):
Wow, this is amazing! I've never used lenses, so I'll need to take another pass before I can fully grasp what's going on, but this is very cool. I wonder why do we need the `optics-helper` file with the `Uninhabited` type? Isn't the empty alternative uninhabited?

hexahedria:
Yes, the empty alternative `{|}` is uninhabited, but so is `({|}, a)` for any `a`, or `{|} | {|}`, for instance. The `Uninhabited` typeclass was an attempt to make any `DataOptic` with an uninhabited `alt` type act like a lens, even if it was one of these other forms. But strictly speaking, that's not really necessary for the idea. I'm going to try separating lenses and prisms as distinct types, which I think would mean that we don't actually have to talk about uninhabited types at all.

I'll leave this PR open for now but I might end up starting over with a new one depending on how much is shared with the new approach.

danieldjohnson:
Closing in favor of #227.

PR #209, opened 2020-08-23
JAX version of raytracer

dougalm:
Here are some timing numbers on my machine:

JAX CPU (32x32 pixels, 8 samples per pixel):  42s compile, 9s run
JAX GPU (32x32 pixels, 8 samples per pixel): 38s compile, 28s run
Dex Cpu (256x256 pixels, 64 samples per pixel): 3s compile, 6s run

Note that the timings are similar but the Dex problem size is 512X bigger.

Loops and control flow aren't JAX's strengths, but I've tried to use pmap and vmap in the JAX version as much as possible. The difference is still much larger than expected, so I suspect there's an error somewhere.

### DIFF
--- a/benchmarks/python/raytrace.py
+++ b/benchmarks/python/raytrace.py
@@ -1,0 +1,472 @@
+# Copyright 2020 Google LLC
+#
+# Use of this source code is governed by a BSD-style
+# license that can be found in the LICENSE file or at
+# https://developers.google.com/open-source/licenses/bsd
+
+from collections import namedtuple
+from functools import partial
+from os import environ
+import time
+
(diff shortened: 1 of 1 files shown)


duvenaud:
Wow, that's satisfying!  I winced every time I introduced a difference between the original JAX version and the initial Dex version, knowing it would make comparisons less meaningful.  But this resolves that problem.

Looking up the [timings of the original version](https://blog.evjang.com/2019/11/jaxpt.html) by Eric Jang, it took about 3 seconds on a GPU for a 150x150 pixel, 100 sample image.  So I agree this version looks like it's extra slow for some reason.

Issue #208, opened 2020-08-22
More build errors on Mac


duvenaud:
After commit 9255389b19cec10ba9d89072f3d3e5f6e8fcad53, building on my Mac fails.

The first error was 
```
clang++ -fPIC -std=c++11 -fno-exceptions -fno-rtti -c -emit-llvm src/lib/dexrt.cpp -o src/lib/dexrt.bc
src/lib/dexrt.cpp:25:42: error: use of undeclared identifier 'aligned_alloc'
  char* result = reinterpret_cast<char*>(aligned_alloc(alignment, nbytes));
```
which I found a fix for.  Apparently ```aligned_alloc``` was introduced in C++17, so I just needed to change line 35 in ```makefile``` to
```
CXXFLAGS := $(CFLAGS) -std=c++17 -fno-exceptions -fno-rtti
```

However, after that when I run ```make```, the build fails when compiling ```LLVMExec```:

```
clang++ -fPIC -std=c++17 -fno-exceptions -fno-rtti -c -emit-llvm src/lib/dexrt.cpp -o src/lib/dexrt.bc
stack build 
Building all executables for `dex' once. After a successful build of all of them, only specified executables will be rebuilt.
dex> configure (lib + exe)
Configuring dex-0.1.0.0...
dex> build (lib + exe)
Preprocessing library for dex-0.1.0.0..
Building library for dex-0.1.0.0..
[ 1 of 28] Compiling Array
[ 2 of 28] Compiling CUDA
[ 3 of 28] Compiling Cat
[ 4 of 28] Compiling Env
[ 5 of 28] Compiling PipeRPC
[ 6 of 28] Compiling Plot
[ 7 of 28] Compiling Util
[ 8 of 28] Compiling Syntax
[ 9 of 28] Compiling PPrint
[10 of 28] Compiling Type
[11 of 28] Compiling Parser
[12 of 28] Compiling Logging
[13 of 28] Compiling MDImp
[14 of 28] Compiling Flops
[15 of 28] Compiling JIT
[16 of 28] Compiling LLVMExec

--  While building package dex-0.1.0.0 using:
      /Users/bmo/.stack/setup-exe-cache/x86_64-osx/Cabal-simple_mPHDZzAJ_2.4.0.1_ghc-8.6.5 --builddir=.stack-work/dist/x86_64-osx/Cabal-2.4.0.1 build lib:dex exe:dex --ghc-options " -fdiagnostics-color=always"
[...]

PR #211, opened 2020-08-25
Use C++17 flag

MikeInnes:
I was getting the following error on OS X with Clang 11.0.3.

```bash
$ make
clang++ -fPIC -std=c++11 -fno-exceptions -fno-rtti -c -emit-llvm src/lib/dexrt.cpp -o src/lib/dexrt.bc
src/lib/dexrt.cpp:25:42: error: use of undeclared identifier 'aligned_alloc'
  char* result = reinterpret_cast<char*>(aligned_alloc(alignment, nbytes));
                                         ^
```

Looks like `aligned_alloc` is defined in C++17, so the fix is to change `-std=c++11` to `-std=c++17`. I think only the `makefile` change was strictly needed to avoid the error, but I assumed it's sensible for `dex.cabal` to be consistent and changed that too.

### DIFF
--- a/dex.cabal
+++ b/dex.cabal
@@ -34,7 +34,7 @@ library
   hs-source-dirs:      src/lib
   ghc-options:         -Wall -O0
   cxx-sources:         src/lib/dexrt.cpp
-  cxx-options:         -std=c++11
+  cxx-options:         -std=c++17
   default-extensions:  CPP, DeriveTraversable, TypeApplications, OverloadedStrings,
                        TupleSections, ScopedTypeVariables, LambdaCase, PatternSynonyms
   if flag(cuda)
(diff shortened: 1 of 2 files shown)


apaszke:
Hey, thanks for the PR! I'm actually preparing a patch that's meant to unbreak macOS, but I really don't want to switch to C++17, because requiring that from all Linux systems we want to build on might be a bit of a stretch. I should have it out by today evening, sorry for the trouble.

apaszke:
See #208 for reference

apaszke:
Ok, I've merged my temporary patch that includes this change into `main`. I'd like to replace `aligned_alloc` with `posix_memalign` since that would allow us to stay at C++11, but for some reason it causes some trouble in the codegen. 

Thanks once again!

MikeInnes:
Sounds good to me, thanks.

PR #213, opened 2020-08-26
Add macOS CI

apaszke:


duvenaud:
Hooray, my job is being replaced by a machine.  Fixing the ```stack.yaml``` thing will make it way easier for me to change git branches.

oxinabox:
does this wean also we can just close #2 ? and add building isntructions for Mac to the readme>

apaszke:
Yeah I would really like to make it easier to run on mac, but I'm not sure if we're there. There's still the caveat that you need to keep specifying `stack --stack-yaml stack-macos.yaml exec dex` instead of just `stack exec dex` whenever you run dex which is a really bad UX. I figured it might be better to get `stack install` to work, and then have people just used the installed version.

But I do think that we should close `#2` and at least have a proper writeup in the README, even if it ends up forcing you to use somewhat awkward commands (we can suggest aliases).

duvenaud:
FYI I ran make after this PR was merged and it worked, once I updated my dex alias as suggested.

apaszke:
Awesome, thanks for the update! Hopefully that will help us avoid breaking your work in the future :)

PR #241, opened 2020-09-24
Make it possible to `make install` Dex.

apaszke:
We no longer depend on any files present in-tree as everything is
bundled using Cabal or Template Haskell. `stack install` seems to just
work!


apaszke:
cc @duvenaud. You should be able to periodically `make install` when you want to update, but `dex` should now work out of tree without any aliases.

PR #254, opened 2020-10-02
Add an explicit syntax for type annotation expressions.

danieldjohnson:
Before, the only way to annotate the type of an expression was to
bind it to a temporary variable; we briefly had explicit annotations
for variants and table literals but those were special cases that
have since been removed.

This change introduces a new syntactic construct `val : type` that is
valid anywhere an expression is allowed, which makes it easier to
annotate subexpressions. It also lets us get rid of the `astype`
prelude helper function, which is a less-readable way of doing the
same thing.

I chose to make `:` left-associative, but that was a pretty arbitrary
choice; I don't imagine people will want to chain this very often.


src/lib/Parser.hs
```diff
@@ -750,6 +750,7 @@ ops =
   [ [InfixL $ sym "." $> mkGenApp TabArrow, symOp "!"]
   , [InfixL $ sc $> mkApp]
   , [prefixNegOp]
+  , [annotatedExpr]
```
apaszke, line 753: I always find myself wondering how strongly does the type annotation operator bind. Do you think it would maybe make sense to enforce people to always wrap those annotations in parens? Otherwise it's super unclear whether the type annotation in `arr.i:type` applies to the index or the array element.

danieldjohnson: I guess to avoid all ambiguity you'd need two sets of parens, right? As in `f ((arr.i) : type)`? Otherwise you still need to define its precedence with respect to other operators.

danieldjohnson: I suppose you could force the user to be explicit by making `:` have the lowest possible precedence but then throwing an error afterward if it looks ambiguous (i.e. try to take as much as possible, but then error if you were able to take more than a single leaf expression)?

src/lib/Inference.hs
```diff
@@ -241,6 +241,10 @@ checkOrInferRho (WithSrc pos expr) reqTy =
   UHole -> case reqTy of
     Infer -> throw MiscErr "Can't infer type of hole"
     Check ty -> freshType ty
+  UAnnotatedExpr val ty -> do
+    ty' <- checkUType ty
+    val' <- checkRho val ty'
```
apaszke, line 246: I think that this could actually be `checkOrInferRho val Concrete ty'`, which would let us e.g. do better inference for table literals.

src/lib/Parser.hs
apaszke: Yeah I agree it might be tricky. I'm not sure if the heuristic you proposed would work, since sometimes you do want to have an application on the right. That's the basic case of `Fin n`.

src/lib/Syntax.hs
```diff
@@ -215,6 +215,7 @@ data UExpr' = UVar UVar
             | UFor Direction UPatAnn UExpr
             | UCase UExpr [UAlt]
             | UHole
+            | UAnnotatedExpr UExpr UExpr
```
apaszke, line 218: nit: `UTypeAnn` seems to be a nice short name, but I don't mind keeping this one.

apaszke (approved):

apaszke:
This looks ready to merge, but can you please update inference so that we can use the annotation for table literal inference?

src/lib/Inference.hs
danieldjohnson: That's something I don't quite understand. What is the difference between`Concrete ty` and `Suggest ty`?

danieldjohnson: In particular I notice that regular binders sometimes use `Concrete` and sometimes use `Suggest`? https://github.com/google-research/dex-lang/blob/208584d880d6406f35ad1da76edd8e7d9d8003cc/src/lib/Inference.hs#L359

Also, I wonder if this should call `checkSigma` instead of `checkOrInferRho`. Maybe an explicit annotation should act the same way as a `UDecl`.

danieldjohnson: I've made it match `UDecl`; let me know if that makes sense.

src/lib/Parser.hs
danieldjohnson: I think I'll leave it as-is for now, then.

src/lib/Inference.hs
apaszke: Yes, that's perfect. The difference between `Check` and `Concrete` is that in `Concrete` checking mode you have a guarantee that the type we're pushing down has no inference variables inside it, while `Check` is a best effort thing. For example, in `Check`ing mode you can find only a partially unified type (e.g. `(n:Type) ?-> (n, ?1)`) or even a completely non-informative inference variable you know nothing about.

I've been recently working on type inference for dependent table literals and I realized that to do it correctly, we'd either need to implement some form of higher order unification or the `Concrete` workaround. In the first case, we'd unify the table type with `i:?1 -> ?2 i` (note that `?2` is a function!) and proceed. Otherwise, we have no way of actually allowing dependency on `i`, unless it is clearly visible. Anyway, that is complicated, so it is just a lot simpler to allow typing dependent tables only when they're explicitly annotated with a type, and in that case we know all the details about dependency.

apaszke:
So I just tried using this, and the precendence issues bit me pretty much immediately 😕 Right now `:` binds stronger than `=>`, so `((for i. ...) : (Fin 5)=>Float)` fails, because it's parsed as `((for i. ...):(Fin 5))=>Float

apaszke:
Yeah, lowest precedence sounds like a reasonable choice.

apaszke:
(Although, should it be lower than `$`? 🤔)

PR #275, opened 2020-10-26
Get rid of AnyValue

apaszke:
We didn't really need it anymore, because in case we know the exact
contructor of an ADT statically, we no longer use the sum-as-product
representation. The only remaining place was in Imp error handling, but
this should be handled differently. We don't need the operation to
produce any result, but instead we abort the code generation for the
current block, and replace its body with just the throwing instruction.

Note that we might want to revisit this strategy if we add debug
instructions. A block that is guaranteed to fail currently
doesn't evaluate anything, but the failure might be only because the
code is incomplete and the user wants to debug something else.


apaszke:
cc @duvenaud

PR #288, opened 2020-11-23
Add more detail about developing on macOS

oxinabox:
I couldn't get it running at first because i didn't do these two things

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ development. Contributions welcome!
   * Install [stack](https://www.haskellstack.org)
   * Install LLVM 9
     * `apt-get install llvm-9-dev` on Ubuntu/Debian,
-    * `brew install llvm@9` on macOS.
+    * `brew install llvm@9` on macOS, and ensure it is on your `PATH` e.g. via `export PATH="/usr/local/opt/llvm@9/bin:$PATH"` before building.
 
 ## Building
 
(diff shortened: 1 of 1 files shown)


apaszke (commented):
This is great, thanks for adding that! Can you only modify the `PATH` suggestion to say `$(brew --prefix llvm@9)/bin`? The path that you've added does not work for the latest LLVM 9.0 tap I think.

apaszke:
Also, can you please switch the destination to the `dev` branch instead of `master`? That's where all the development happens now and we don't want them to get out of sync.

README.md
```diff
@@ -29,7 +29,7 @@ development. Contributions welcome!
   * Install [stack](https://www.haskellstack.org)
   * Install LLVM 9
     * `apt-get install llvm-9-dev` on Ubuntu/Debian,
-    * `brew install llvm@9` on macOS.
+    * `brew install llvm@9` on macOS, and ensure it is on your `PATH` e.g. via `export PATH="/usr/local/opt/llvm@9/bin:$PATH"` before building.
```
oxinabox, line 32:
```suggestion
    * `brew install llvm@9` on macOS, and ensure it is on your `PATH` e.g. via `export PATH="$(brew --prefix llvm@9)/bin:$PATH"` before building.
```

PR #297, opened 2020-11-27
Added Quicksort example

### LINKED ISSUE #292: a sorting function
It would be nice to include some form of sort in the prelude.
Probably quick sort would be suitable,
but even bubble sort would be nice to have.

This could then be used to implement median and quantiles

duvenaud:
Due to missing features there are still a few  opportunities for taking advantage of Dex's type system that, when fixed, should make it faster, shorter, and safer:
 1. Writing a stack that allows in-place replacement of its underlying buffer with a larger one seems to be beyond the capabilities of Dex right now.  After conversations with @dougalm, it seems that adding dependent deferencing is all that's needed.   After conversations with @apaszke, I think supporting nested references would also be enough.  For now I just used a fixed-sized buffer.
 2. The `partition` function I wrote looks similar to the existing implementations I found online.  There is a shorter and less error-prone way to write it, which operates only on entire tables (having its caller implicitly specify `high` and `low` in the index set).  I wrote it that way first, but couldn't figure out how to create a reference to an in-place slice of an existing table.  I think I want something like this signature:
```
def sliceRef (xsRef: Ref h (n=>t)) (low:Int) (high:Int) :
  {State h} Ref h ((Range low high)=>t) =
    todo
```
this typechecks, but I don't know how to write the body.

Writing this also raised some style questions for Dex code:
3. For in-place updates, should we prefer relying on linearity, or the `State` effect?  My current feeling is that linearity would be preferable if we can count on it, since it's much less obtrusive.  It also keeps function signatures simpler.  However it's also much less flexible.
4. I understand that partial functions are declasse, so I wrote e.g. `pop` to return a `Maybe`.  But I ended up needed to call unsafe functions on the output of `pop` when I knew the stack wasn't empty.  @apaszke suggested we might want to make the condition function of `while` return a `Maybe`, where the value is passed into the body, which would remove one unnecessary unsafe check.
5. Should `sort` go in the prelude?

Fixes #292.

### DIFF
--- a/examples/sort.dx
+++ b/examples/sort.dx
@@ -1,15 +1,40 @@
 '## Stack operations.
+This is a fake stack for now until Dex gets dependent dereferencing,
+or can handle nested references.
 
-def push (x:a) (xs:List a) --o : List a =
-  xs <> (AsList 1 [x])
+bufsize = 1000
 
-def pop (list: List a) --o : (Maybe a & List a) =
-  (AsList nx xs) = list
-  case nx == 0 of
-    True -> (Nothing, list)
+data Stack a:Type h:Type =
+  AsStack size: (Ref h Int) buf: (Ref h ((Fin bufsize)=>a))
+
+def pop (stack:Stack a h) : {State h} Maybe a =
+  (AsStack size buf) = stack
+  case get size == 0 of
+    True -> Nothing
     False ->
-      newLength = nx - 1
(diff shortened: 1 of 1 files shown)


dan-zheng:
`dev` branch has been merged into `main`, now that plotting has been revamped. `dev` branch will be deleted now - we can continue development on `main`.

I'll change the base branch of this PR to `main` now.

duvenaud:
Since merge sort [#571] got merged, I think this one is mostly obsolete.

PR #322, opened 2020-12-08
Remove built in fst and snd

oxinabox:
AFAICT `fst` and `snd` can be implemented in pure dex, using pattern matching.
So it should be possible to delete all the code for handling them as a built-in operator.
Code that doesn't exist has no bugs.

Though, I think it's been possible for quite a while to do this, because tuple destructuring was implemented long ago.
So perhaps they are still here for a reason?

### DIFF
--- a/src/lib/Syntax.hs
+++ b/src/lib/Syntax.hs
@@ -307,7 +307,7 @@ data PrimCon e =
         deriving (Show, Eq, Generic, Functor, Foldable, Traversable)
 
 data PrimOp e =
-      | TabCon e [e]                 -- table type elements
+        TabCon e [e]                 -- table type elements
       | ScalarBinOp BinOp e e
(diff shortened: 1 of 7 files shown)


src/lib/Syntax.hs
```diff
@@ -307,8 +307,6 @@ data PrimCon e =
         deriving (Show, Eq, Generic, Functor, Foldable, Traversable)
 
 data PrimOp e =
-        Fst e
-      | Snd e
       | TabCon e [e]                 -- table type elements
```
oxinabox, line 310:
```suggestion
        TabCon e [e]                 -- table type elements
```

apaszke:
I think that this is going to conflict with #290 which contains this and other changes, so I'll close it if you don't mind. But thanks for the PR!

Issue #321, opened 2020-12-08
Pretty-print `n=>Char` tables as string literals


dan-zheng:
It would be better to print `n=>Char` values as `"xyz"` string literals rather than the standard `['x', 'y', 'z']` format.

```console
$ dex repl
>=> :t "Hello world"
((Fin 11) => Char)
">=> "Hello world"
['H', 'e', 'l', 'l', 'o', ' ', 'w', 'o', 'r', 'l', 'd']
```

dan-zheng:
I'll give this a shot tonight.

apaszke:
FYI the printing you see there is not that of `UTabCon`, but `pprintVal` from `Serialize.hs`

Issue #302, opened 2020-12-05
Multibyte unicode characters become different single byte characters when used in strings


oxinabox:
This seems incorrect

```
>=> "a🍕b"
['a', 'U', 'b']
>=> "a平b"
['a', 's', 'b']
```

oxinabox:
Oh this happess for simple character literals also
```
>=> :p '🍕'
'U'
>=> :p '平'
's'
```

dougalm:
Right, our `String` type is currently just a list of bytes, which only works for ASCII. We should figure out how to support Unicode.

We want to avoid making the same mistake as Haskell, which is [notorious for having too many string types](http://dev.stephendiehl.com/hask/#strings): `String`, `Text`, `Text.Lazy`, `ByteString`, `ByteString.Lazy`, `ShortByteString`, ...

The Text/ByteString distinction is fundamental but the rest of Haskell's zoo of string types is just about navigating a performance-expressiveness tradeoff. The default `String` is a native list of characters, which is a very convenient object to work with. But it's physically represented as a linked list of pointers to characters, which can be slow.

Dex has it easier here. Our native list type is just a length-hidden table, whose physical representation is a contiguous array of unboxed data. So a Dex byte string represented as a `List Byte` will automatically be stored as a contiguous array of bytes. Great!

But Unicode is harder. The simplest approach would be to treat strings as lists of UTF-32-encoded characters like this:

    data Char = MkChar Word32  -- opaque type to represent a single Unicode character
    String : Type = List Char  -- ordinary list

But that uses 4X more space than necessary for mostly-ASCII strings. Alternatively, we could use a wrapper around a UTF-8 encoded byte string:

    data String = MkString (List Byte)   -- Opaque type to represent a Unicode string

But then we lose the list-of-characters structure. Specifically, we no longer have O(1) random access to characters, which all of Dex's native array-processing idioms are all built on.

Or we could have both. Maybe a three-member zoo isn't so bad?

apaszke:
To be honest even with UTF-32 encoding you don't get O(1) random access to characters, because the encoding of a single "charater" in the sense that we usually mean it might actually span multiple Unicode codes (they all form a single _grapheme_).

Citing [this SO answer](https://stackoverflow.com/a/27331885):
> For example, both a and ä are graphemes, but they may consist of multiple code points (e.g. ä may be two code points, one for the base character a followed by one for the diaeresis ...)

[...]

PR #329, opened 2020-12-10
Pretty-print `n => Char` tables as string literals.

dan-zheng:
Make `prettyVal` print `n => Char` tables as string literals.
Add `make update-all` command for updating all test files. Update tests.

Resolves https://github.com/google-research/dex-lang/issues/321.
Depends on https://github.com/google-research/dex-lang/pull/327, which should be reviewed and merged first.

---

Examples:

```
>=> "hello world"
"hello world"

>=> AsList _ "hello " <> AsList _ "world"
(AsList 11 "hello world")

>=> {a="1234", b=[1, 2, 3]}
{a = "1234", b = [1, 2, 3]}

>=> x = "ab"
>=>  for i,j. [x.i, x.j]
["aa", "ab", "ba", "bb"]@(Fin 2 & Fin 2)
```


src/lib/Serialize.hs
```diff
@@ -79,6 +91,18 @@ prettyVal val = case val of
             variant = Variant (NoExt types) theLabel repeatNum value
         _ -> error "SumAsProd with an unsupported type"
     _ -> return $ pretty con
+  Record (LabeledItems row) -> do
```
apaszke, line 94: Is this to fix the record-of-tables issue? Or did you have some other reason to support this?

src/lib/Serialize.hs
```diff
@@ -79,6 +91,18 @@ prettyVal val = case val of
...
+    let separator = line' <> ","
+    let bindwith = " ="
+    let elems = concatMap (\(k, vs) -> map (k,) (toList vs)) (M.toAscList row)
+    let fmElem = \(label :: Label, v) -> do
+          vStr <- prettyVal v
+          return $ pretty label <> bindwith <+> vStr
```
apaszke, line 100: If you want to be super fancy than this can be a one liner I think: `fmap (pretty label <> bindwidth <+>) $ prettyVal v` 😛

apaszke (approved):

apaszke:
This needs a rebase now. @dougalm this is going to conflict with your PR. What order should we merge those in?

src/lib/Serialize.hs
dan-zheng: Yes, this is to fix the record-of-tables issue, which was merged in https://github.com/google-research/dex-lang/pull/327. I've rebased.

src/lib/Serialize.hs
dan-zheng: Thanks for the tip! I applied a similar one-liner.

`hlint` suggested `((pretty label <> bindwith) <+>) <$> prettyVal v` instead of using `fmap`.

apaszke: `<$> = fmap` 😄

src/lib/Serialize.hs
```diff
@@ -37,19 +37,31 @@ getDexString (DataCon _ _ 0 [_, xs]) = do
...
     elems <- forM idxs $ \idx -> do
-      ans <- evalBlock mempty $ snd $ applyAbs abs idx
-      asStr <$> prettyVal ans
-    let idxSetStr = case idxSet of
-                      FixedIntRange l _ | l == 0 -> mempty
-                      _                          -> "@" <> pretty idxSet
-    return $ pretty elems <> idxSetStr
+      atom <- evalBlock mempty $ snd $ applyAbs abs idx
+      case atom of
+        Con (Lit (Word8Lit c)) ->
```
dan-zheng, line 56: Note: https://github.com/google-research/dex-lang/pull/315 removed the pattern `CharLit c`, so the replacement is now `Con (Lit (Word8Lit c))`.

Could we add a new abbreviated pattern for this, while being future-facing for [potential Unicode support](https://github.com/google-research/dex-lang/issues/302)? I wonder what would be a good name.

src/lib/Serialize.hs
```diff
@@ -37,19 +37,31 @@ getDexString (DataCon _ _ 0 [_, xs]) = do
...
+          return $ showChar (toEnum @Char $ fromIntegral c) ""
+        _ -> pprintVal atom
+    let bodyType = getType body
+    let elemsDoc = case bodyType of
+          -- Print table of characters as a string literal.
+          TC (BaseType (Scalar Word8Type)) -> pretty ('"': concat elems ++ "\"")
```
dan-zheng, line 62: Note: https://github.com/google-research/dex-lang/pull/315 removed the pattern `CharType`, so the replacement is now `TC (BaseType (Scalar Word8Type))`.

Could we add a new abbreviated pattern for this, while being future-facing for [potential Unicode support](https://github.com/google-research/dex-lang/issues/302)? I wonder what would be a good name.

dan-zheng (commented):
Rebased on `dev` after https://github.com/google-research/dex-lang/pull/315 was merged. Tests pass again.

apaszke (approved):

apaszke:
Thanks, Dan!

PR #345, opened 2020-12-14
Misc syntax fixes

dougalm:


apaszke (approved):
LGTM. Though to be fair for some reason I think I now prefer the `case ... of True -> ... False -> ...` form in Haskell anyway, because it makes it easier to rearrange the cases (to make the code read more naturally) without having to modify the condition.

PR #360, opened 2020-12-19
[README] Edit build dependency instructions.

dan-zheng:
Use consistent formatting and punctuation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,9 +28,12 @@ development. Expect monstrous bugs and razor-sharp edges. Contributions welcome!
 
   * Install [stack](https://www.haskellstack.org)
   * Install LLVM 9
-    * `apt-get install llvm-9-dev` on Ubuntu/Debian,
-    * `brew install llvm@9` on macOS, and ensure it is on your `PATH` e.g. via `export PATH="$(brew --prefix llvm@9)/bin:$PATH"` before building.
-  * Install libpng (often included by default in *nix)
+    * Ubuntu/Debian: `apt-get install llvm-9-dev`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -28,9 +28,12 @@ development. Expect monstrous bugs and razor-sharp edges. Contributions welcome!
 
   * Install [stack](https://www.haskellstack.org)
   * Install LLVM 9
-    * `apt-get install llvm-9-dev` on Ubuntu/Debian,
-    * `brew install llvm@9` on macOS, and ensure it is on your `PATH` e.g. via `export PATH="$(brew --prefix llvm@9)/bin:$PATH"` before building.
-  * Install libpng (often included by default in *nix)
+    * Ubuntu/Debian: `apt-get install llvm-9-dev`
+    * macOS: `brew install llvm@9`
+      * Make sure `llvm@9` is on your `PATH` before building: `export PATH="$(brew --prefix llvm@9)/bin:$PATH"`
```
oxinabox, line 33: I don't mind if you remove it, but when I wrote it the `e.g.` was intentional.
That's only an example of how one might go about putting llvm on your path. It doesn't work on all systems.
In particular, it doesn't work on mine because I use fish-shell.
It also doesn't work on old-school borne shell, not that that matters.

dan-zheng: Thanks for the clarification! I added `Example:` to that line to emphasize it's not the only/correct way.

dan-zheng:
I think this should be non-controversial (and can easily be edited again), so I'll merge.

PR #363, opened 2020-12-21
Fix bug in Atan2

### LINKED ISSUE #362: atan2 implementation is incorrect
```
>=> atan2 (-1.0) (-1.0)
-3.1415927
```
which equals `pi`, but really should be `-3/4*pi`.

cc @duvenaud

duvenaud:
Fixes #362

I originally tested 14 different inputs to `atan2`, but somehow managed to miss two quadrants.  I also added a test that goes systematically around the circle.


apaszke (approved):
This is great, thanks!

PR #366, opened 2020-12-22
Automate wrapping of JITed function pointers in Python callables

apaszke:
This patch extends the export functionality with the ability to generate
a description of how the user-facing arguments and results map to the
exported native function. Then, I created a mini-language that can
faithfully encode this description, and is exposed in the foreign API
via the `dexGetFunctionSignature` function.

On the Python side, I have implemented a parser for that language, which
turns it into a few `NativeType` objects that describe how the arguments
are to be (de)serialized when interfacing with the native function. This
completely automates the manual labor of creating ctype wrappers and
allows used to call the compiled function as if it was any other Python
callable.

Note that this interface also supports zero-copy bidirectional
conversion of NumPy arrays and Dex tables with `Fin` index sets.

For example:
```py
import dex
import numpy as np
from textwrap import dedent

sigmoid = dex.eval(r"\x:Float. 1.0 / (1.0 + exp(-x))").compile()
print(sigmoid(-1.0), sigmoid(0.0), sigmoid(1.0))

transpose = dex.Module(dedent("""
def myTranspose (n: Int) ?-> (m: Int) ?->
                (x : (Fin n)=>(Fin m)=>Float) : (Fin m)=>(Fin n)=>Float =
  for i j. x.j.i
""")).myTranspose.compile()
example = np.arange(25, dtype=np.float32).reshape((5, 5))
print(transpose(example))  # NB: Implicit arguments get inferred from shapes
print(example.T)
```


python/tests/jit_test.py
```diff
@@ -0,0 +1,79 @@
...
+def test_polymorphic_array_2d():
+  m = dex.Module(dedent("""
+  def myTranspose (n: Int) ?-> (m: Int) ?->
+                  (x : (Fin n)=>(Fin m)=>Float) : (Fin m)=>(Fin n)=>Float =
+    for i j. x.j.i
+  """))
+  check_atom(m.myTranspose, lambda x: x.T,
+             [(np.arange(a*b, dtype=np.float32).reshape((a, b)),)
+              for a, b in it.product((2, 5, 10), repeat=2)])
+
```
dan-zheng, line 79: I wonder what's the status/plan for supporting features like:
* Calling Autodiff APIs (e.g. `grad`, `linearize`, `transpose`).
* Defining Dex ADTs within Python, then using them.

Maybe these work already? If so, how about adding some tests?

dan-zheng (commented):
Very neat!

python/tests/jit_test.py
apaszke: None of those work at the moment. AD and vmap are the next things to do here. I would imagine that one could also add support for record and tuple return/arguments. ADTs might be a little more difficult, but doable at some point too. Definitely not a priority though.

PR #369, opened 2020-12-23
Clean up the Cabal file

apaszke:
Remove unused dependencies, group the ones we actually need in
categories.

Also, remove some files that have bitrotted.


dougalm:
Dead code elimination always feels good!

dan-zheng:
Nice reorganization!

I'm curious about some of the deleted files: why they were outdated/deleted, and what's the future plan in that area?
- JAX bridging: `JAX.hs`, JAX IR, Python stubs for bridging with JAX.
  - Did Dex previously support generating "JAX IR" usable from Python?
- `PipeRPC.hs`: what was this used for?

apaszke:
It was mostly an experiement, to learn more about the expressive differences between Dex and JAX. Yes, Dex did support emitting a jaxpr-like IR, which later would be evaluated from Python. PipeRPC was the RPC protocol for doing that evaluation.

dan-zheng:
Got it, thank for the context!

Issue #348, opened 2020-12-16
Segfaults with sum types containing tables


dougalm:
The following causes segfaults about 50% of the time for me:

```
data MySum =
  Foo Float
  Bar String

:p
  xs = for i:(Fin 3).
    if ordinal i < 2
      then Foo 2.0
      else Foo 1.0
  (xs, xs)
```

The segfaults don't happen if we use the other constructor (`Bar`). The problem is that we store sum types as products with a tag, so `MySum` would be stored in something like a `(Ref Tag & Ref Float & Ref String)` . When we create a `Foo 1.0`, we allocate the memory for the whole sum type but we only initialize the tag and the memory associated with the `Foo` constructor, the `Ref Float`. The `Ref String` goes uninitialized. This causes problems when we try to load the value (`destToAtom` in `Imp.hs`). This just loads from all of the elements, including the `Ref String`, which is a pointer to a pointer. Dereferencing the pointer gives a garbage pointer, and dereferencing that gives a segfault.

Interestingly, we can get the same bug (I think?) with a fixed-sized array instead of a list, but it's a bit harder:

```
data MySum2 =
  Foo2
  Bar2 (Fin 3 => Int)

:p concat for i:(Fin 4). AsList _ [(Foo2, Foo2)]
```

PR #376, opened 2020-12-23
Add Text support to Diagram

oxinabox:
The following code:
```
mydiagram : Diagram =
   (  (circle 7.0 |> moveXY (20.0, 20.0) |> setFillColor blue |> setStrokeColor red)
   <> (circle 5.0 |> moveXY (40.0, 40.0))
   <> (rect  10.0 20.0 |> moveXY (5.0, 10.0) |> setStrokeColor red)
   <> (text (str "DexLang")  |> moveXY (30.0, 10.0) |> setStrokeColor green)
   <> (pointDiagram |> moveXY (15.0, 5.0) |> setStrokeColor red)
   )

:html renderSVG mydiagram ((0.0, 0.0), (100.0, 50.0))
```
produces the following  output:
Screenshot 2020-12-23 at 22 56 28

I also did a minor refactor of `renderGeom`, though I tried to avoid changing too much.
(I may make another PR to try and refactor this whole file somewhat, see if i can find something cleaner for some bit.)

### DIFF
--- a/lib/diagram.dx
+++ b/lib/diagram.dx
@@ -174,7 +174,10 @@ def renderGeom (attr:GeomStyle) ((x,y):Point) (geom:Geom) : String =
     Text content ->
       textEle = tagBracketsAttr "text" $
         ("x" <=> x <.>
-         "y" <=> y)
+         "y" <=> y <.>
+         "text-anchor" <=> "middle" <.>    -- horizontal center
+         "dominant-baseline" <=> "middle"  -- vertical center
+        )
       groupEle solidAttr $ textEle content
 
 BoundingBox : Type = (Point & Point)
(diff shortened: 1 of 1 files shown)


oxinabox:
The "exciting" follow up that this will allow us to put labelled ticks on plot axes

oxinabox:
Should prob set `text-anchor="middle"`
to agree with rectangle and circle being anchored in their center
.https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/text-anchor

oxinabox:
It did need centering

### before
Screenshot 2020-12-29 at 18 55 34

## after
Screenshot 2020-12-29 at 18 54 36

oxinabox:
should be good to merge now.
Doesn't have any tests, but AFAICT nothing else in diagram.dx does?
Not sure though, maybe i missed them

dougalm:
Lovely!

>Doesn't have any tests, but AFAICT nothing else in diagram.dx does?
Only indirectly, since we test the example programs that use diagram.dx

I'm actually surprised this isn't segfaulting because of #348 but I won't complain!

oxinabox:
Need to remember to resubmit this PR after #348 is closed

PR #380, opened 2020-12-26
Improve HTML doc generation.

dan-zheng:
## Genotypical changes

Improve makefile:
- Add support for building docs for files in lib/.
- Add `mkdir -p doc` to commands so they do not fail if doc/ does not exist.
- Add standalone `doc-prelude` target so that `make docs` is fully incremental.

Improve prelude formatting:
- Use consistent formatting (h2, ##) for section headers.

## Phenotypical consequences

`make docs` always works now.

Prelude headings look a bit more uniform.

Screen Shot 2020-12-26 at 9 37 18 AM

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,14 +12,14 @@ To learn more, check out our
 or these example programs:
 
   * [Dex prelude](https://google-research.github.io/dex-lang/prelude.html)
-  * [Mandelbrot set](https://google-research.github.io/dex-lang/mandelbrot.html)
-  * [Ray tracer](https://google-research.github.io/dex-lang/raytrace.html)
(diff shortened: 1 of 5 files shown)


dan-zheng:
The prelude looks fine ([preview from my fork](http://danzheng.me/dex-lang/prelude.html)) because it's top-level in the `gh-pages` branch, but nested HTML files in `lib/` and `examples/` don't quite work yet due to incorrect relative path usage:

<details>

Screen Shot 2020-12-26 at 11 26 52 AM

</details>

Will look into this, wonder how to fix it in the least invasive way.

PR #383, opened 2020-12-26
Update MCMC example

devmotion:
The comments of @oxinabox in the Julia Slack channel got me interested in dex-lang and therefore I had a look at the MCMC example. This PR contains some minor changes that I assumed could be useful for this example:
- The computations in the acceptance/rejection step of the MH algorithm are all performed in log space
- An explanation is added to the example
- The number of samples is increased to show more clearly that the mean and covariance of the samples converge to the expected values for both MH and HMC

### DIFF
--- a/examples/mcmc.dx
+++ b/examples/mcmc.dx
@@ -1,5 +1,6 @@
+'# Markov Chain Monte Carlo
 
--- === General MCMC utilities ===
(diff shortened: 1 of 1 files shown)


dougalm:
Looks great. Thanks for the PR!

PR #384, opened 2020-12-27
Extend rejection sampling example

devmotion:
This PR extends the rejection sampling example.
- It adds some additional explanations
- It replaces one occurrence of `log(1 + ...)` in the log density of the Binomial distribution with `log1p(...)`
- It performs all calculations in log space and therefore removes `binomialProb`
- It adds evaluations of the mean and the variance of a larger number of samples
- It adds implementations of inversion sampling as a comparison


dougalm:
Thanks!

PR #385, opened 2020-12-27
Fix tests of MCMC example

devmotion:
It seems that tests are broken due to my first PR yesterday (https://github.com/google-research/dex-lang/pull/383). I noticed that the resulting mean and covariances that I get locally are different from the [logs on Github](https://pipelines.actions.githubusercontent.com/0ViJRnwNK463d0Kvz6WthzZIlC8yPZaOpHuC7LD3yQl3tCHPsB/_apis/pipelines/1/runs/331/signedlogcontent/3?urlExpires=2020-12-27T10%3A34%3A30.7028112Z&urlSigningMethod=HMACV1&urlSignature=tCiSUr4jMWQx7RqeRmHsjW2MU0uDo6csQx9JweLBI0k%3D), so I assume that this test is just not stable enough and different architectures and systems yield slightly different results. Therefore I replaced the test with a check that the empirical mean and variances are close to the true values (within a relative tolerance of 0.1).

### DIFF
--- a/examples/mcmc.dx
+++ b/examples/mcmc.dx
@@ -88,15 +88,20 @@ def myLogProb (x:(Fin 2)=>Float) : LogProb =
   x' = x - [1.5, 2.5]
   neg $ 0.5 * inner x' [[1.,0.],[0.,20.]] x'
 
+-- check that mean and variance of samples is close to true values with relative tolerance 0.1.
+def testMeanAndCovariance (samples:n=>(Fin 2)=>Float) : Bool =
+  (m, c) = meanAndCovariance samples
+  isClose = \x y. all $ for i. allclose atol 0.1 x.i y.i
+  isClose m [1.5, 2.5] && isClose (diagonal c) [1., 0.05]
+
 numSamples = 10000
 k0 = newKey 1
 
 mhParams = 0.1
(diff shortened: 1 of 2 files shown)


oxinabox:
>  I noticed that the resulting mean and covariances that I get locally are different from the logs on Github, so I assume that this test is just not stable enough and different architectures and systems yield slightly different results.

Could be the fact that Dex currently uses the LLVM intrinsics for basic floating point functions like log and the like.
Which hit the system libm.
And the system libm is not going to be the same across systems.
(It's also not certain to be particularly good)
Bundling e.g.openlibm would fix that, but it's a lot of effort.

dougalm:
Thanks for fix. I merged the original PR too hastily without realizing that the CI was failing. In #389 I made the float pretty-printer print fewer decimal places, which should make these quine tests less sensitive to small numerical differences between platforms. (I also prefer fewer decimal places anyway. It's almost always the right thing to do for real data.)

We could also use the approximate equality operator, `~~`, that @duvenaud  added to the prelude. But since this MCMC example is primarily an example, and only incidentally a test, it makes more sense to me to show to the actual results rather than just `True`.

I'll close this PR because of #389 but thanks for the quick follow-up!

PR #386, opened 2020-12-28
Add shell.nix.

athas:
This makes it much more convenient to compile Dex for the lucky few who use Nix or NixOS.


athas:
@googlebot I signed it!

apaszke:
We should also set up CI for this build method, but merging for now. Thanks Troels!

PR #387, opened 2020-12-28
Add cabal.project.

athas:
This is necessary for cabal to find dependencies stored in Git repositories.

This is not by itself sufficient for `cabal` to work, since the `install` rule in the `makefile` contains stuff that only works for `stack`, but it does make `make` and `cabal run dex` work.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -1,0 +1,13 @@
+packages:        dex.cabal
+
+source-repository-package
+    type:       git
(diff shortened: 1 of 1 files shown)


athas:
@googlebot I signed it!

apaszke (approved):

apaszke:
Yeah, we haven't been using the `cabal` build path for quite a while apparently. Another thing to add to CI... Thanks!

PR #389, opened 2020-12-29
Print floats with fewer decimal places to make quine tests less brittle.

dougalm:


apaszke (commented):
Printing with fewer digits is not the problem here... The quine tests should use approximate equality instead (`~~`). This also has the bad side effect of printing integral values as `9.` which is not a valid float literal.

dougalm:
I give a bit more context in #385. I agree that tests should use `(~~)`. But we also use our *examples* as tests, and for those I think it's better to show actual values rather than merely assert closeness to serve their incidental role as tests. I also prefer printing fewer decimal places anyway, so I'd be happy with the change even if it didn't fix the tests.

Nice catch with the `9.`. But I think `9.` *is* a valid dex float literal right? I'm sure there are other edge cases I missed though.

PR #397, opened 2020-12-29
RFC: make print polymorphic and fluent

oxinabox:
Putting this out there for comments as to if it is a good idea.
If it is then this PR still needs tests added.

 - Making it call `show` on its input just seems more convient
 - Making it return its input means you can (for debugging purposes) slip it inside some compouind expression
     -  e.g. `x |> f |>g` can become `x |> f |> print |> g`

idk how useful this really is.


dougalm:
An alternative would be to have three different functions, like Haskell:

```
def putString (s:String) : {State World} Unit = ...

def print (_:Show a) ?=> (x:a) : {State World} Unit =
  putString $ show x

def trace (s:String) : String = unsafeIO do
  putString s -- has unsafe side effects
  s
```

Notice that `trace` doesn't need the IO effect exposed, because it won't get DCEd as long as its result it used.

PR #398, opened 2020-12-29
add renderScaledSVG

oxinabox:
Basically I would like to just workout the bounding box that is big enough to show everything contained in the diagram,
and then leave it to the SVG view-box code to workout how to make it fit.

Possibly this could replace some of the scaling code in `plot.dx`,
though that would still at very least be needed for colours.
And it is the natural point where logScales and the like could be added.
So probably not.
Though maybe the position scales could default to identity, which would make for simplier manipulation of the Plot after converting to a`Diagram` form for adding in text anotation etc.
Though one could just use the plot.dx scaling stuff for that anyway, and then it would be general to any scale.

So really this is mostly useful for if someone is working with Diagramss directly, not with plots which are already scaled.

WIP because it currently `SIGSEGV (Address boundary error)`'s for me for the first diagram.
I will have to do some poking to workout why.

### DIFF
--- a/lib/diagram.dx
+++ b/lib/diagram.dx
@@ -181,6 +181,27 @@ def renderGeom (attr:GeomStyle) ((x,y):Point) (geom:Geom) : String =
 
 BoundingBox : Type = (Point & Point)
 
+def computeBounds (d:Diagram) : BoundingBox =
+  computeSubBound = \sel op (_, p, geom).
+    sel p + case geom of
+      PointGeom     -> 0.0
+      Circle r      -> op r
+      Rectangle w h -> op $ (sel (w,h))/2.0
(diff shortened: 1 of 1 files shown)


dougalm:
Looks  good!

> WIP because it currently SIGSEGV (Address boundary error)'s for me for the first diagram.

I'm actually seeing segfaults on main since merging #376  so I don't think it's due to this change. I'm pretty sure it's #348 . I'll see if I can fix it today.

> Possibly this could replace some of the scaling code in plot.dx,
> though that would still at very least be needed for colours.

Also for axis ticks. The idea with the scaling information in plot.dx is that it contains the mapping between data space and plot space, so it lets us place fiducial markers like axis ticks to help humans mentally invert that mapping.

oxinabox:
seen  #348 is fixed this should be good now.

apaszke (approved):

PR #399, opened 2020-12-30
Expose `view`, a "lazy" variant of `for`

dougalm:
Dex tables are already represented internally as lambda expression that describe how to obtain each index. This is the bridge between the physical representation (buffers of flat arrays) and the logical representation (index-to-value mappings). Different lambdas can represent different views of the same underlying data. It's similar to strided arrays but more general.

This PR exposes that representation to the surface language. This lets us write reindexing functions like `transpose` and ensure they don't materialize new data, without relying on a compiler optimization. For example, we can use `view` to define `transpose`, and guarantee that it will have `O(1)` cost instead of potentially `O(N^2)` cost.

      def transpose (x:n=>m=>a) : m=>n=>a = view i j. x.j.i

One objection to exposing `view` is that it should be the compiler's job to figure out when to materialize tables and when to keep them as views. That might still be the right way to do things eventually, but in the meantime I want to see what happens if we give some of this power to users.  One clear benefit is that it makes it possible to reason about asymptotic performance based on a simple baseline operational semantics without having to reason about compiler optimizations. (Of course, the compiler is always free to *improve* performance over this baseline.)

For example, with the old `for` definition of `transpose`, this function may run in `O(N^3)` time if the inlining optimization doesn't fire. With the new `view` version, it's guaranteed to be `O(N^2)`.

      def diagOfXTimesXT (mat:n=>n=>Float) : n=>Float =
        for i. vdot mat.i (transpose mat).i

Similarly, we can write matmul using `view`, and be sure that the argument to `sum` won't be fully materialized.

     matmul : (l=>m=>Float) -> (m=>n=>Float) -> (l=>n=>Float) = \x y.
       for i k. sum view j. x.i.j * y.j.k

Note that these are not *mutable* views. However, we can achieve something very similar to mutable views using views of tables of references.

### DIFF
--- a/lib/prelude.dx
+++ b/lib/prelude.dx
@@ -397,7 +397,7 @@ def Fin (n:Int) : Type = Range 0 n
 def ordinal (i:a) : Int = %toOrdinal i
 def size (n:Type) : Int = %idxSetSize n
 def unsafeFromOrdinal (n : Type) (i : Int) : n = %unsafeFromOrdinal n i
-def iota (n:Type) : n=>Int = for i. ordinal i
+def iota (n:Type) : n=>Int = view i. ordinal i
 
 -- TODO: we want Eq and Ord for all index sets, not just `Fin n`
 @instance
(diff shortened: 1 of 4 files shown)


oxinabox:
> One objection to exposing view is that it should be the compiler's job to figure out when to materialize tables and when to keep them as views. That might still be the right way to do things eventually, but in the meantime I want to see what happens if we give some of this power to users.

I do like the notion that it is the compilers problem to work this out.
I think it is reasonable to expose a option that forces it to be lazy, just line `noinline` forces something not to be inlined (even though deciding about inlining is the compilers job.).

On the basis that `for` is to let the compiler work it out,
maybe we should call this `lazy`, and have its complement `eager`, as ways to force the compiler to do a particular behavour.
or: `materialize for.` and `nomaterialize for`, but that is long.

PR #405, opened 2020-12-31
clang install instructions

joaogui1:
Clang is [not necessarily](https://unix.stackexchange.com/a/483516) installed with llvm, so it would be better to add it to the installation instructions


apaszke:
Great point, thanks!

PR #407, opened 2021-01-03
Add completions to REPL

oxinabox:
This adds completions to the REPL, both of language keywords and of currently defined top level names;
as well as restricting filename completions to only happen within a quoted string.

Shown is me pressing tab at the end of each line.

image

<s>
A follow up PR would add globally inscope variables to this.
I think it would not be too hard, though I don't think I will work out out anytime soon.
There is discussion in [this Reddit thread](https://www.reddit.com/r/haskell/comments/1os0yq/haskeline_woes/) about how to dynamically update the list of completions for Haskline

It also would be good to add other things like `:p` and `:html` filtered to only be allowed occur at line start
</s>

### DIFF
--- a/src/dex.hs
+++ b/src/dex.hs
@@ -17,7 +17,6 @@ import System.Exit
 import System.Directory
 import Data.List
 import Data.Text.Prettyprint.Doc
-import Debug.Trace
 
 import Syntax
 import PPrint
--- a/src/dex.hs
+++ b/src/dex.hs
@@ -16,7 +16,6 @@ import System.Posix.IO (stdOutput)
 import System.Exit
 import System.Directory
 import Data.List
-import Data.Text.Prettyprint.Doc
 
 import Syntax
 import PPrint
@@ -27,8 +26,8 @@ import Resources
 import TopLevel
 import Parser  hiding (Parser)
 import LiveOutput
+import Env (envNames)
 import Export
-import Env
 
 data ErrorHandling = HaltOnErr | ContinueOnErr
 data DocFmt = ResultOnly | TextDoc | HTMLDoc | JSONDoc
@@ -40,29 +39,6 @@ data EvalMode = ReplMode String
 
 data CmdOpts = CmdOpts EvalMode (Maybe FilePath) EvalConfig
 
-
-
-filteredCompletions candidates str = map simpleCompletion $
-  filter (str `isPrefixOf`) candidates
-
-completions :: CompletionFunc (StateT TopEnv IO)
-completions (line, _) = do
-  env <- get
-  let varNames = map (show . pretty) $ envNames env
-  -- note: line and thus word and rest have character order reversed
-  let (word, rest) = break (== ' ') line
-  let anywhereKeywords = ["def", "for", "rof", "case", "data", "where", "of", "if", "then", "else", "interface", "instance", "do", "view"]
-  let startoflineKeywords = ["%bench \"", ":p", ":t", ":html", ":export"]
-  let candidates = varNames ++ anywhereKeywords ++ if null rest
-                                                    then startoflineKeywords
-                                                    else []
-  return (rest, filteredCompletions candidates $ reverse word)
-
-dexCompletions = completeQuotedWord (Just '\\') "\"'" listFiles completions
-
-hasklineSettings :: Settings (StateT TopEnv IO)
-hasklineSettings = setComplete dexCompletions defaultSettings
-
 runMode :: EvalMode -> Maybe FilePath -> EvalConfig -> IO ()
 runMode evalMode preludeFile opts = do
   key <- case preludeFile of
@@ -71,7 +47,9 @@ runMode evalMode preludeFile opts = do
   env <- cached "prelude" key $ evalPrelude opts preludeFile
   let runEnv m = evalStateT m env
   case evalMode of
-    ReplMode prompt ->
+    ReplMode prompt -> do
+      let filenameAndDexCompletions = completeQuotedWord (Just '\\') "\"'" listFiles dexCompletions
+      let hasklineSettings = setComplete filenameAndDexCompletions defaultSettings
       runEnv $ runInputT hasklineSettings $ forever (replLoop prompt opts)
     ScriptMode fname fmt _ -> do
       results <- runEnv $ evalFile opts fname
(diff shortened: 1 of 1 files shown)


oxinabox:
I added `:p`, `:t` etc, though they think they can occur anywhere in the line right now,
and `%bench "` though that inserts an extra space afterwards that it shouldn't.
I think still better to have than not, though I am going to play around to see if i can do them right.

oxinabox:
Apparently it is impossible not to inser the ` ` after `%bench "`
https://github.com/judah/haskeline/issues/151

but I did sort out `:p` etc needing to be at start of line.

src/dex.hs
```diff
@@ -37,6 +38,26 @@ data EvalMode = ReplMode String
...
+  filter (str `isPrefixOf`) keywordList
+
+keywordCompletions :: Monad m => CompletionFunc m
+keywordCompletions (line, _) = do
+  -- note: line and thus word and rest have character order reversed
+  let (word, rest) = break (== ' ') line
+  let anywhereKeywords = ["def", "for", "rof", "case", "data", "where", "of", "if", "then", "else", "interface", "instance", "do", "view"]
+  let startoflineKeywords = ["%bench \"", ":p", ":t", ":html", ":export"]
+  let keywords = anywhereKeywords ++ if null rest
+                                        then startoflineKeywords
```
oxinabox, line 53: haskells indentation rules are really grosss.
idk if there is a nice way to write this without the huge hanging indent

src/dex.hs
```diff
@@ -46,7 +67,7 @@ runMode evalMode preludeFile opts = do
   let runEnv m = evalStateT m env
   case evalMode of
     ReplMode prompt ->
-      runEnv $ runInputT defaultSettings $ forever (replLoop prompt opts)
+      runEnv $ runInputT hasklineSettings $ forever (replLoop prompt opts)
```
oxinabox, line 70: AFAICT all the things we need to have completion of global function names etc is available to us in the `env` monad.
But I have no real idea how to get it out of there and into the monad that `hasklineSettings`/`keywordCompletions` gets.

oxinabox:
I worked out how to make global names work in general
I am so satistified.
I mean this is probably the worst haskell ever but still its by far the most complex i ever wrote.

Screenshot 2021-01-03 at 21 10 01

I guess there are some names we want to somehow filter out, like `showInt32`

src/dex.hs
```diff
@@ -37,6 +41,29 @@ data EvalMode = ReplMode String
...
+  -- note: line and thus word and rest have character order reversed
+  let (word, rest) = break (== ' ') line
+  let anywhereKeywords = ["def", "for", "rof", "case", "data", "where", "of", "if", "then", "else", "interface", "instance", "do", "view"]
+  let startoflineKeywords = ["%bench \"", ":p", ":t", ":html", ":export"]
+  let candidates = varNames ++ anywhereKeywords ++ if null rest
+                                                    then startoflineKeywords
+                                                    else []
+  return (rest, filteredCompletions candidates $ reverse word)
+
+dexCompletions = completeQuotedWord (Just '\\') "\"'" listFiles completions
```
oxinabox, line 62: Probably all of the above should be in another file. idk where though

src/dex.hs
```diff
@@ -37,6 +41,29 @@ data EvalMode = ReplMode String
...
+
+filteredCompletions candidates str = map simpleCompletion $
+  filter (str `isPrefixOf`) candidates
+
+completions :: CompletionFunc (StateT TopEnv IO)
+completions (line, _) = do
+  env <- get
+  let varNames = map (show . pretty) $ envNames env
+  -- note: line and thus word and rest have character order reversed
+  let (word, rest) = break (== ' ') line
```
oxinabox, line 54: Possibly other whitespace should be allowed

oxinabox, line 54: Possibly other whitespace should be allowed

src/dex.hs
```diff
@@ -37,6 +41,29 @@ data EvalMode = ReplMode String
...
 
+
+
+filteredCompletions candidates str = map simpleCompletion $
+  filter (str `isPrefixOf`) candidates
+
+completions :: CompletionFunc (StateT TopEnv IO)
+completions (line, _) = do
+  env <- get
+  let varNames = map (show . pretty) $ envNames env
```
oxinabox, line 52: `(show . pretty)` is almost certainly not the right way to get the names as strings.
idk what is though

src/dex.hs
```diff
@@ -15,6 +15,9 @@ import System.Posix.Terminal (queryTerminal)
 import System.Posix.IO (stdOutput)
 import System.Exit
 import System.Directory
+import Data.List
+import Data.Text.Prettyprint.Doc
+import Debug.Trace
```
oxinabox, line 20:
```suggestion
```

src/dex.hs
```diff
@@ -37,6 +40,29 @@ data EvalMode = ReplMode String
...
+completions (line, _) = do
+  env <- get
+  let varNames = map (show . pretty) $ envNames env
+  -- note: line and thus word and rest have character order reversed
+  let (word, rest) = break (== ' ') line
+  let anywhereKeywords = ["def", "for", "rof", "case", "data", "where", "of", "if", "then", "else", "interface", "instance", "do", "view"]
+  let startoflineKeywords = ["%bench \"", ":p", ":t", ":html", ":export"]
+  let candidates = varNames ++ anywhereKeywords ++ if null rest
+                                                    then startoflineKeywords
+                                                    else []
```
oxinabox, line 58: I have no idea what the preferred way to indent this is. 
I hope it isn't this.

src/dex.hs
```diff
@@ -37,6 +40,29 @@ data EvalMode = ReplMode String
...
+  return (rest, filteredCompletions candidates $ reverse word)
+
+dexCompletions = completeQuotedWord (Just '\\') "\"'" listFiles completions
+
+hasklineSettings :: Settings (StateT TopEnv IO)
+hasklineSettings = setComplete dexCompletions defaultSettings
```
apaszke, line 64: nit: small local definitions like this one (or `dexCompletions`) are usually better placed in the `where` clauses of the functions that use them (in this case`runMode`).

src/dex.hs
```diff
@@ -37,6 +40,29 @@ data EvalMode = ReplMode String
 
 data CmdOpts = CmdOpts EvalMode (Maybe FilePath) EvalConfig
 
+
+
+filteredCompletions candidates str = map simpleCompletion $
+  filter (str `isPrefixOf`) candidates
```
apaszke, line 46: This would also be better off in a `where` clause, or just inlined below.

src/dex.hs
apaszke: `pprint` would do the trick here.

src/dex.hs
apaszke: Well, you could put it on a separate line and make it a one-liner

apaszke (commented):
Nice, this looks sweet!

apaszke:
Let's just clean up the formatting a little and this is going to be good to merge!

src/dex.hs
oxinabox: It seems like if i do that it is too far away from where it is used.
So I have put it side that `ReplMode` case

src/dex.hs
```diff
@@ -81,6 +85,19 @@ replLoop prompt opts = do
...
+dexCompletions :: CompletionFunc (StateT TopEnv IO)
+dexCompletions (line, _) = do
+  env <- get
+  let varNames = map pprint $ envNames env
+  -- note: line and thus word and rest have character order reversed
+  let (word, rest) = break (== ' ') line
+  let anywhereKeywords = ["def", "for", "rof", "case", "data", "where", "of", "if", "then", "else", "interface", "instance", "do", "view"]
+  let startoflineKeywords = ["%bench \"", ":p", ":t", ":html", ":export"]
+  let candidates = varNames ++ anywhereKeywords ++
+                   if null rest then startoflineKeywords else []
```
apaszke, line 97: This is quite minor, but `++` has complexity proportional to the size of the lhs, so it is usually better to sort the concatenated lists according to increasing length. This concatenates them in the complete reverse order, which is the worst case scenario.

apaszke (approved):

src/dex.hs
oxinabox: TIL, it's an actual linked list i guess

apaszke: Yup, Haskell lists are linked lists. If you want a contiguous arrays you have to use `Data.Vector` (which is a packed array of _boxed values_; if you want to go unboxed there are `Data.Vector.Unboxed` types).

PR #436, opened 2021-01-07
[README] Fix typo.

dan-zheng:
`tematic` -> `thematic`


dan-zheng:
Online dictionaries don't seem to recognize "tematic" as an alternate spelling of "thematic", so I assume it is a typo. Merging.

PR #437, opened 2021-01-07
Gardening.

dan-zheng:
NFC changes:

- Address some hlint warnings.
- Add doc comments to `Dag` in `LiveOutput.hs` - added in passing while making sense of the source code.
- Improve `dex` help option printing with `helpOption` helper function.

Minor functional changes:

- Rename `dex --backend interp` flag to `--backend interpret`. `interpret` reads more naturally.

### DIFF
--- a/src/dex.hs
+++ b/src/dex.hs
@@ -11,9 +11,10 @@ import System.Exit
 import Control.Monad
 import Control.Monad.State.Strict
 import Options.Applicative
+import Text.PrettyPrint.ANSI.Leijen (text, hardline)
 import System.Posix.Terminal (queryTerminal)
 import System.Posix.IO (stdOutput)
-import System.Exit
+
 import System.Directory
 import Data.List
 
@@ -59,7 +60,7 @@ runMode evalMode preludeFile opts = do
     WebMode    fname -> runWeb      fname opts env
     WatchMode  fname -> runTerminal fname opts env
     ExportMode dexPath objPath -> do
-      results <- fmap snd <$> (runEnv $ evalFile opts dexPath)
+      results <- fmap snd <$> runEnv (evalFile opts dexPath)
       let outputs = foldMap (\(Result outs _) -> outs) results
       let errors = foldMap (\case (Result _ (Left err)) -> [err]; _ -> []) results
       putStr $ foldMap (nonEmptyNewline . pprint) errors
@@ -71,7 +72,7 @@ runMode evalMode preludeFile opts = do
 evalPrelude :: EvalConfig -> Maybe FilePath -> IO TopEnv
 evalPrelude opts fname = flip execStateT initTopEnv $ do
   source <- case fname of
-              Nothing   -> return $ preludeSource
+              Nothing   -> return preludeSource
               Just path -> liftIO $ readFile path
   result <- evalSource opts source
   void $ liftErrIO $ mapM (\(_, Result _ r) -> r) result
@@ -96,7 +97,7 @@ dexCompletions (line, _) = do
   let startoflineKeywords = ["%bench \"", ":p", ":t", ":html", ":export"]
   let candidates = (if null rest then startoflineKeywords else []) ++
                    anywhereKeywords ++ varNames
-  let completions = map simpleCompletion $ filter ((reverse word) `isPrefixOf`) candidates
+  let completions = map simpleCompletion $ filter (reverse word `isPrefixOf`) candidates
   return (rest, completions)
 
 liftErrIO :: MonadIO m => Except a -> m a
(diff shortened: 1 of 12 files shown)


src/lib/Parser.hs
```diff
@@ -116,10 +116,10 @@ sourceBlock' :: Parser SourceBlock'
...
       proseBlock
   <|> topLevelCommand
-  <|> liftM declToModule (topDecl <* eolf)
-  <|> liftM declToModule (instanceDef  <* eolf)
-  <|> liftM declToModule (interfaceDef <* eolf)
-  <|> liftM (Command (EvalExpr Printed) . exprAsModule) (expr <* eol)
+  <|> fmap declToModule (topDecl <* eolf)
+  <|> fmap declToModule (instanceDef  <* eolf)
+  <|> fmap declToModule (interfaceDef <* eolf)
+  <|> fmap (Command (EvalExpr Printed) . exprAsModule) (expr <* eol)
```
apaszke, line 122: I'm pretty sure @dougalm has just changed those to `liftM` a few commits ago 😛

apaszke (approved):
Thanks! While I will accept this, I am always a little wary of linter-based patches. For one, they do tend to overshadow the real history of changes to the code, and also they like to pick one normal form for many idioms (e.g. always `fmap`, never `liftM`), but I really don't mind using the latter. In fact, in some places it might even be clearer as it is more specific.

src/dex.hs
```diff
@@ -170,13 +176,14 @@ optionList opts = eitherReader \s -> case lookup s opts of
...
 parseEvalOpts = EvalConfig
-  <$> (option
+  <$> option
          (optionList [ ("LLVM", LLVM)
                      , ("LLVM-CUDA", LLVMCUDA)
                      , ("LLVM-MC", LLVMMC)
-                     , ("interp", Interp)])
-         (long "backend" <> value LLVM <> help "Backend (LLVM(default)|LLVM-CUDA|interp)"))
-  <*> (optional $ strOption $ long "logto"
+                     , ("interpret", Interpret)])
```
apaszke, line 183: How about "interpreted" or "interpreter"? "backend" sounds like a query for a noun, not a verb

src/dex.hs
```diff
@@ -11,9 +11,10 @@ import System.Exit
 import Control.Monad
 import Control.Monad.State.Strict
 import Options.Applicative
+import Text.PrettyPrint.ANSI.Leijen (text, hardline)
```
apaszke, line 14: Where is this used?

src/dex.hs
```diff
@@ -131,34 +132,39 @@ printLitProg JSONDoc prog =
...
 
 parseOpts :: ParserInfo CmdOpts
 parseOpts = simpleInfo $ CmdOpts
   <$> parseMode
-  <*> (optional $ strOption $ long "prelude" <> metavar "FILE" <> help "Prelude file")
+  <*> optional (strOption $ long "prelude" <> metavar "FILE" <> help "Prelude file")
   <*> parseEvalOpts
 
+helpOption :: String -> String -> Mod f a
+helpOption optionName options =
```
apaszke, line 146: How does this change the printing?

src/lib/Parser.hs
dan-zheng: Thanks for mentioning - I indeed found [@dougalm's changes to `liftM`](https://github.com/google-research/dex-lang/commit/599e2ee47c53b355954012c96e9515d3cbdc863c#diff-7352b0160734bf8ae9c7f0c87b1ff7641db108b16e448fa3393a1804ce6c1b3fR119-R122) in [`599e2e`](https://github.com/google-research/dex-lang/tree/599e2ee47c53b355954012c96e9515d3cbdc863c).

I reverted all changes renaming `liftM` to `fmap`. I also added a `.hlint.yaml` file with a `- ignore: {name: "Use fmap"}` directive to track this intentional coding style preference.

dougalm: Ha, yes, I probably did. I prefer `liftM` for cases like this because it's more specific than `fmap` and it signals "we're dealing with a monad here". To me `fmap` connotes "simple container" functors like lists. Even though monads *are* functors, we tend to use them in quite a different way from ordinary containers like lists, trees, etc.

src/dex.hs
dan-zheng: Renamed to `--backend interpreter`, thanks. Definitely agree that a noun is ideal, wasn't sure about the comfort level for using longer names.

"Swift API Design Guidelines"-level of descriptive naming [("clarity is more important than brevity")](https://swift.org/documentation/api-design-guidelines/#clarity-over-brevity) certainly seems too much, for Haskell and Dex.

dan-zheng: Another random renaming change that might help clarity is consistently renaming `Tab` → `Table` in types (e.g. `UTabCon` → `UTableCon`). Any thoughts here?

src/dex.hs
dan-zheng: Previously, help option strings (`(literate(default)|result-only|HTML|JSON`) where placed on their own line simply because the string was so long and didn't have spaces.

Now, help option strings have an explicitly added line break (`hardline`), allowing us to make the messages themselves nicer by adding spaces. Outer parentheses are also removed.

---

Before:
```console
$ dex --help
Usage: dex COMMAND [--prelude FILE] [--backend ARG] [--logto FILE]

Available options:
  --prelude FILE           Prelude file
  --backend ARG            Backend (LLVM(default)|LLVM-CUDA|interp)
  --logto FILE             File to log to
  -h,--help                Show this help text

$ dex script --help
Usage: dex script FILE [--outfmt ARG] [--allow-errors]

Available options:
  FILE                     Source program
  --outfmt ARG             Output format
                           (literate(default)|result-only|HTML|JSON
  --allow-errors           Evaluate programs containing non-fatal type errors
  -h,--help                Show this help text
```

After:

```console
$ dex --help
Usage: dex COMMAND [--prelude FILE] [--backend ARG] [--logto FILE]

Available options:
  --prelude FILE           Prelude file
  --backend ARG            Backend
                           LLVM (default) | LLVM-CUDA | LLVM-MC | interpreter
  --logto FILE             File to log to
  -h,--help                Show this help text

$ dex script --help
Usage: dex script FILE [--outfmt ARG] [--allow-errors]

Available options:
  FILE                     Source program
  --outfmt ARG             Output format
                           literate (default) | result-only | HTML | JSON
  --allow-errors           Evaluate programs containing non-fatal type errors
  -h,--help                Show this help text
```

dan-zheng: Quick bikeshed: any opinion on renaming `dex script --outfmt {HTML,JSON}` to `dex script --outfmt {html,json}` for consistent casing? @dougalm

src/dex.hs
apaszke: Those things are internal, so I don't think that really matters. I don't think that `Tab` allows any confusion, so let's avoid changing things too much unless there's a good reason for that.

src/dex.hs
dan-zheng: Most CLIs seem to use lowercase option values. @apaszke seems supportive of renaming option values to lowercase for consistency: done in https://github.com/google-research/dex-lang/pull/437/commits/37e4f50f555a1966a48c0ed09449639848bb9cf3.

Meta: in general, I don't think it is a good idea to spend too much time bikeshedding - wanted to quickly bounce this change sooner rather than later.

src/lib/Parser.hs
dan-zheng: That makes sense!

Haskell neophyte question: I wonder if there are any meaningful differences (e.g. performance) between using `fmap` and `liftM`?

### Self-answer

[`liftM`'s definition](https://hackage.haskell.org/package/base-4.14.1.0/docs/src/GHC.Base.html#liftM): top-level function defined for `Monad m`.
```hs
-- | Promote a function to a monad.
liftM   :: (Monad m) => (a1 -> r) -> m a1 -> m r
liftM f m1              = do { x1 <- m1; return (f x1) }
```

[`fmap`'s definition](https://hackage.haskell.org/package/base-4.14.1.0/docs/src/GHC.Base.html#fmap): typeclass method.
```hs
class  Functor f  where
    fmap        :: (a -> b) -> f a -> f b
```

Monadic types seem to implement `Functor` via `fmap = liftM`:

```hs
newtype Identity a = Identity { runIdentity :: a }

instance Functor Identity where
    fmap = liftM
```

So calling `liftM` is actually more direct than calling `fmap` (which involves one extra indirection).

dan-zheng:
Merged.

Note: https://github.com/google-research/dex-lang/pull/437/commits/37e4f50f555a1966a48c0ed09449639848bb9cf3 changed option names, so externally-saved `dex` commands may require updating. Hope that's okay and not too inconvenient. [Minor meta comment](https://github.com/google-research/dex-lang/pull/437#discussion_r553740420) about avoiding excessive bikeshedding.

Changed option names:

- `dex --backend llvm` (was `LLVM`)
- `dex --backend llvm-cuda` (was `LLVM-CUDA`)
- `dex --backend llvm-mc` (was `LLVM-MC`)

- `dex --outfmt html` (was `HTML`)
- `dex --outfmt json` (was `JSON`)

PR #438, opened 2021-01-07
Use llvm-hs/llvm-hs@llvm-9 in cabal.project.

dan-zheng:
Previously, a downstream commit at `apaszke/llvm-hs` was referenced.
That commit has now been merged into `llvm-hs/llvm-hs@llvm-9`.


apaszke:
Hah, I forgot to update it after it was merged. Thanks!

PR #445, opened 2021-01-09
Remove extra CSS bottom padding in HTML-rendered Dex.

dan-zheng:
The bottom padding adds unnecessary empty bottom scrolling, which slightly hurts UX.

---

I wonder if the bottom padding was added for some reason in https://github.com/google-research/dex-lang/commit/33afb4a6? I couldn't tell.


dan-zheng:
I'll go ahead and merge, since this is a small change. Happy to discuss afterwards.

Issue #426, opened 2021-01-05
Render LaTeX equations in web notebooks


apaszke:
Sounds like a pretty useful feature to have in a computing environment if you ask me.

dan-zheng:
I've started looking into this! Supporting LaTeX snippets seems straightforward using [Pandoc](https://hackage.haskell.org/package/pandoc):

```haskell
markdownToHtml :: String -> Either P.PandocError Html
markdownToHtml s = P.runPure $ do
  let latexExtensions = P.getDefaultExtensions "commonmark_x"
  let readerOptions = P.def {
                        P.readerExtensions = latexExtensions
                      }
  let writerOptions = P.def {
                        P.writerExtensions = latexExtensions,
                        P.writerHTMLMathMethod = P.KaTeX P.defaultKaTeXURL
                      }
  let markdownText = strip (pack s) -- stripping leading whitespace is important for correctness
  markdown <- P.readMarkdown readerOptions markdownText
  result <- P.writeHtml5 writerOptions markdown
  return result
```

Pandoc is unfortunately quite large as a dependency, but that seems acceptable.

Still need to figure out how to appropriately hook into KaTeX so that `renderMathInElement(document.body)` fires at the right time (after the DOM has loaded). I tried listening for the `DOMContentLoaded` event, but it doesn't work yet [(code from here)](https://katex.org/docs/autorender.html#usage):

```html
<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/katex@0.12.0/dist/katex.min.css" integrity="sha384-AfEj0r4/OFrOo5t7NnNe46zW/tFgW6x/bCJG8FqQCEo3+Aro6EYUG4+cU+KJWu/X" crossorigin="anonymous">
<script defer src="https://cdn.jsdelivr.net/npm/katex@0.12.0/dist/katex.min.js" integrity="sha384-g7c+Jr9ZivxKLnZTDUhnkOnsh30B4H0rpLUpJ4jAIKs4fnJI+sEnkvrMWph2EDg4" crossorigin="anonymous"></script>
<script defer src="https://cdn.jsdelivr.net/npm/katex@0.12.0/dist/contrib/auto-render.min.js" integrity="sha384-mll67QQFJfxn0IYznZYonOWZ644AWYC+Pt2cHqMaRhXVrursRwvLnLaebdGIlYNa" crossorigin="anonymous"></script>
<script>
    document.addEventListener("DOMContentLoaded", function() {
[...]

PR #447, opened 2021-01-09
Do not perform `dex web` LaTeX rendering for non-prose blocks.

dan-zheng:
Apply KaTeX `renderMathInElement` render function only to `.prose-block` HTML elements.
Skip code blocks (`.code-block`) and error-output blocks (`err-block`).

Add "test case" (example) to `examples/latex.dx`.

---

Fixes this bug: https://github.com/google-research/dex-lang/issues/426#issuecomment-757116749.

Before:

KaTeX for Dex bug

After (fixed):

KaTeX for Dex fix

### DIFF
--- a/examples/latex.dx
+++ b/examples/latex.dx
@@ -28,6 +28,17 @@
 
 '$$\begin{rcases} a &\text{if } b \\\\ c &\text{if } d \end{rcases} \Rightarrow \dots$$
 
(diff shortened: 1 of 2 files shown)


dan-zheng:
Merging. I think this basically concludes LaTeX rendering features and bug fixes.

PR #469, opened 2021-01-17
List some commandline flags and examples

oxinabox:
I struggle to remember exactly how to do thing.
Also there is no help for the REPL flag.
also I didn't know about `-h` to list the main flags.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,19 @@ alias dex="stack exec --stack-yaml=stack-macos.yaml dex -- --lib-path lib"
   * Execute script: `dex script examples/pi.dx`
   * Live-updated notebook display `dex web examples/pi.dx` (html) or `dex watch
     examples/pi.dx` (terminal).
+    
+### Command-line Flags:
+Listed here are some commonly used flags and examples.
+For a full list run `dex -h`.
+
+ - `--prelude [file]` start by loading the given file instead of the normal prelude.
+     - e.g. `dex --prelude=myprelude.dx repl` starts the dex repl with `myprelude.dx` loaded.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -68,6 +68,19 @@ alias dex="stack exec --stack-yaml=stack-macos.yaml dex -- --lib-path lib"
...
+For a full list run `dex -h`.
+
+ - `--prelude [file]` start by loading the given file instead of the normal prelude.
+     - e.g. `dex --prelude=myprelude.dx repl` starts the dex repl with `myprelude.dx` loaded.
+     - e.g. `dex --prelude=/dev/null web lib/prelude.dx` starts dex without any prelude loaded, so you can edit the `prelude.dx` file and view it in the web notebook.
+
+#### REPL only:
+ - `--prompt [string]` use the given string as the prompt instead of `>=> `
+    - e.g. `dex repl --prompt dex:` starts the REPL with the prompt being `dex:`
+    - e.g. `dex repl --prompt "$(printf '\x1B[32m\x02>=> \x1B[m\x02')"` starts the REPL with the prompt being `>=> ` in green text
```
apaszke, line 83: Is that really such an important flag? I've never really used it

oxinabox: Colouring the prompt massively enhances readability for me because I can easily see where one statement starts.
But it's a personal preference thing.
I personally am a huge fan of colours in the terminal.

I think we should list it here as it's not listed anywhere else.
We probably should add it it `dex -h` I guess

axch:
I'm going to close this, given that it hasn't been merged in all this time.  Feel free to reopen if you disagree.

The `--prompt` flag is (technically) discoverable with `dex -r repl`; and I don't think there is consensus that we want to maintain a flag list in the README.  Perhaps we could autogenerate a documentation page with all the flags for all the execution modes?

PR #473, opened 2021-01-20
Add `nil` empty List.

dan-zheng:
Use `nil` instead of `AsList _ []` or `(mempty: List a)`.


dan-zheng:
After chatting with @dougalm, we realized that this `nil` addition has broader implications.
Maybe there's room for more improvement and upside. Check out the discussion thread: https://github.com/google-research/dex-lang/discussions/476

axch:
Closing this PR as abandoned.  If we decide to adopt `nil` in the future, it won't be hard to implement.

PR #479, opened 2021-01-21
Add examples/nn to the CI

srush:
The NN example script was added without passing the CI. This adds it to the CI. 

(Dependent on https://github.com/google-research/dex-lang/pull/472 for the MNist file downloads.)


srush:
Updated this to use Fashion MNist and provide more literate comments describing the main ideas.

Issue #486, opened 2021-01-24
Support LaTeX in static rendered HTML


dan-zheng:
https://github.com/google-research/dex-lang/pull/444 added LaTeX rendering for `dex web` via KaTeX.

However, static HTML pages [(example)](https://google-research.github.io/dex-lang/examples/pi.html) created via `dex script --outfmt html` do not yet have KaTeX support:

<img width="500" alt="Screen Shot 2021-01-24 at 6 44 17 PM" src="https://user-images.githubusercontent.com/5590046/105647499-2ed72100-5e74-11eb-9a57-eb3ef290b762.png">

We ought to fix this, and strive for feature parity between `dex web` (dynamic) and `dex script --outfmt html` (static) in the future.

PR #487, opened 2021-01-25
Support LaTeX in static rendered HTML.

dan-zheng:
Static HTML pages created via `dex script --outfmt html` now support rendered LaTeX (via KaTeX), in addition to `dex web` dynamic notebooks.

Example:

Screen Shot 2021-01-24 at 6 49 26 PM

---

Resolves https://github.com/google-research/dex-lang/issues/486.
Happy to regenerate `gh-pages` after this is merged.


apaszke (changes requested):
I wonder why can't we just reuse the same JavaScript file for both static and dynamic pages? I understand that there are slight differences (e.g. the dynamic pages listen for the events), but couldn't we just have a top-level variable that picks which mode we're in? There is a large set of behaviors (LaTeX rendering, TOC generation, collapsible sections) that I expect will be the same in both modes.

static/dynamic.html
```diff
@@ -9,15 +9,16 @@
...
+  <!-- Do dynamic webpage rendering on-load. -->
+  <script defer src="https://cdn.jsdelivr.net/npm/katex@0.12.0/dist/contrib/auto-render.min.js" integrity="sha384-mll67QQFJfxn0IYznZYonOWZ644AWYC+Pt2cHqMaRhXVrursRwvLnLaebdGIlYNa" onload="render(RENDER_MODE.DYNAMIC);" crossorigin="anonymous"></script>
 </head>
 
 <body>
 
 <div id="main-output"></div>
 
-</body>
+<script src="/index.js""></script>
```
dan-zheng, line 20: Note: I moved this `<script>` element within `<body>` because `<script>` is technically only valid within `<body>`. Chrome applies this change automatically upon loading the page, making the source consistent seems good.

dan-zheng (commented):
> I wonder why can't we just reuse the same JavaScript file for both static and dynamic pages?

Done: I created a unified `static/index.js` function exposing a `render(renderMode)` function. This is quite cleaner, thanks for the suggestion.

apaszke (approved):

Issue #291, opened 2020-11-23
Automate generation of Github Pages


oxinabox:
I looked at https://google-research.github.io/dex-lang/prelude.html
and saw it still showed `Float` as `Real`
I looked at the **gh-pages** branch and saw it had not been updated since August 10th.

I think something needs to be setup so that on a PR being merged  HTML renderings are updated in the **gh-pages** branch.
I am sure GitHub actions can do this.

apaszke:
Yeah, I agree this should be handled by GitHub actions, but we didn't get to it just yet. It would be a great contribution if you're up for it!

apaszke:
@dougalm might have more details as to how he generated those pages. IIRC it involved some manual modifications, but we can just check in a diff file that will be applied.

dougalm:
The only manual modification is changing the parameters for the ray tracer. The checked-in version uses a low resolution so you see a result quickly if you run it.

Some of the examples (e.g. the ray tracer) take a while to run on a beefy machine, so they might not be a perfect fit for GH actions. Maybe we should set up GH actions for things like the prelude and lightweight examples and continue running the computationally intensive examples manually. That way we also wouldn't have to worry about applying a diff file.

PR #467, opened 2021-01-16
Add navigation sidebar for `dex web`

oxinabox:
closes #400 

As one can probably tell this was:
 1.  written from scratch without reference to someone else implementation
 2. Basically my first time writing javascript
 
It gets the job done, but I am sure it could be way better.
The CSS also could use some work
e.g. right now if it is very long don't know what happens, 
and it will happily go over the top of the main content.

screenshot

PR #491, opened 2021-01-28
Automate HTML doc generation via GitHub Actions.

dan-zheng:
Add GitHub Action workflow to run `make docs` and push to `gh-pages` branch upon pushes to main branch.
This should greatly alleviate the need to manually update `gh-pages` branch.

---

Verified this workflow on my fork, it seems to work well.

`gh-pages` deployment options:

```yaml
    - name: Deploy to GitHub Pages
      uses: "JamesIves/github-pages-deploy-action@3dbacc7e69578703f91f077118b3475862cb09b8" # 4.1.0
      with:
          token: ${{ secrets.GITHUB_TOKEN }}
          branch: gh-pages # The branch the action should deploy to.
          folder: doc      # The folder the action should deploy.
          clean: false     # If true, automatically remove deleted files from the deploy branch.
          commit-message: Updating gh-pages from ${{ github.sha }}
```

The `clean` option is set to `false`, so ad-hoc added files ([like HTML redirection pages](https://github.com/google-research/dex-lang/pull/392)) won't be removed on deployments.

Resolves https://github.com/google-research/dex-lang/issues/291.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -8,6 +8,7 @@ on:
 
 jobs:
   build:
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
         os: [ubuntu-18.04, macos-latest]
@@ -18,8 +19,6 @@ jobs:
           - os: ubuntu-18.04
             install_deps: sudo apt-get install llvm-9-tools llvm-9-dev pkg-config
             path_extension: /usr/lib/llvm-9/bin
-
-    runs-on: ${{ matrix.os }}
 
     steps:
     - name: Checkout the repository
@@ -46,7 +45,8 @@ jobs:
         key: ${{ runner.os }}-${{ hashFiles('**/*.cabal', 'stack*.yaml') }}
         restore-keys: ${{ runner.os }}-
 
-      # See https://github.com/actions/cache/issues/445
+    # This step is a workaround.
+    # See issue for context: https://github.com/actions/cache/issues/445
     - name: Remove cached Setup executables
       run: rm -rf ~/.stack/setup-exe-cache
       if: runner.os == 'macOS'
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -48,10 +48,10 @@ jobs:
       run: make docs
 
     - name: Deploy to GitHub Pages
-      uses: "JamesIves/github-pages-deploy-action@3.7.1"
+      uses: "JamesIves/github-pages-deploy-action@3dbacc7e69578703f91f077118b3475862cb09b8" # 4.1.0
       with:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          BRANCH: gh-pages # The branch the action should deploy to.
-          FOLDER: doc      # The folder the action should deploy.
-          CLEAN: false     # If true, automatically remove deleted files from the deploy branch.
-          COMMIT_MESSAGE: Updating gh-pages from ${{ github.sha }}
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -19,10 +20,8 @@ jobs:
             install_deps: sudo apt-get install llvm-9-tools llvm-9-dev pkg-config
             path_extension: /usr/lib/llvm-9/bin
 
-    runs-on: ${{ matrix.os }}
-
     steps:
-    - name: Checkout the repository
+    - name: Check out the repository
```
oxinabox, line 24: isn't `Checkout` here a referring to `git checkout`? and thus the spelling is correct?
Not that it is a big deal either rway

.github/workflows/docs.yaml
```diff
@@ -0,0 +1,57 @@
...
+    - name: Generate docs
+      run: make docs
+
+    - name: Deploy to GitHub Pages
+      uses: "JamesIves/github-pages-deploy-action@3.7.1"
+      with:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          BRANCH: gh-pages # The branch the action should deploy to.
+          FOLDER: doc      # The folder the action should deploy.
+          CLEAN: false     # Automatically remove deleted files from the deploy branch
```
oxinabox, line 56:
```suggestion
          CLEAN: false     # false => Don't automatically remove deleted files from the deploy branch
```

oxinabox (commented):
🎉

.github/workflows/ci.yaml
dan-zheng: Thanks. I don't feel strongly, consistently changed back to `Checkout` everywhere.

.github/workflows/docs.yaml
dan-zheng: Thanks for pointing out the ambiguous comment! [(Copied from here.)](https://github.com/JamesIves/github-pages-deploy-action#getting-started-airplane)

Improved:
```yaml
CLEAN: false     # If true, automatically remove deleted files from the deploy branch.
```

oxinabox:
It would be good to do this, so we can benifit from #467  in the files linked in the readme. Especially the gigantic prelude and also the tutorial.

.github/workflows/docs.yaml
```diff
@@ -0,0 +1,57 @@
...
+        restore-keys: ${{ runner.os }}-
+
+    - name: Build
+      run: make build
+
+    - name: Generate docs
+      run: make docs
+
+    - name: Deploy to GitHub Pages
+      uses: "JamesIves/github-pages-deploy-action@3.7.1"
```
apaszke, line 51: Can you please pin the version of this action to a specific commit? This is the Google requirement for any third-party packages, since tags can be changed to point to arbitrary other content. See [this](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsuses) for syntax.

apaszke (changes requested):
LGTM, but we need to pin a concrete commit of the third-party GitHub action to use it.

.github/workflows/docs.yaml
dan-zheng: Good idea, thanks. Done.

apaszke (approved):

apaszke:
Thanks and sorry for the sloppiness!

oxinabox:
This looks like it failed the first time it ran 
https://github.com/google-research/dex-lang/runs/2298443642?check_suite_focus=true

apaszke:
Looks like it works now! But I'm still not seeing navigation 🤔

PR #501, opened 2021-02-13
simple completion

joaogui1:
This adds a simple completion to dex, now if you type "dex<tab>" bash will show web, repl, watch and script as completion options, after selecting one of them bash will resume normal folder/file completion. 
Should we add instructions to setting the completion on README or should we set the completions using the makefile?


joaogui1:
@dougalm @apaszke can you check this PR?

apaszke (approved):
This LGTM, although I would appreciate a link to something that would let me verify its correctness. I don't really know how bash completions work. Also, could you please move this to `misc/`?

Finally, sorry for the delay. We've been in paper writing mode for the last few weeks, which had the unfortunate effect of all the delays on GitHub.

joaogui1:
No worries about the delay!
I'm not sure which link could I send you (I used a couple guides to write it, but I don't have them anymore), couldn't you test this branch using nix or docker?
Also moved it to misc!

axch:
I don't know much about bash completions either, but is there any reason not to merge this?  @apaszke, @dougalm?  There is precedent for not-mechanically-tested stuff in `misc/`, e.g. `dex.el`.

joaogui1:
I can try fixing it over the weekend, do you have any examples of such parser?

apaszke:
I found [this](https://www.gnu.org/software/bash/manual/html_node/A-Programmable-Completion-Example.html), but it's not great. I think the idea would be to look through `COMP_WORDS` to see if the mode has already been selected in another argument.

joaogui1:
Got it, I can take a stab at it during the weekend

joaogui1:
Ok this was harder than I imagined, as all bash-related things are, the current version stops autocompleting after repl and looks for ".dx" files in case of script, watch and web. Another option presented by the unused function would be to run find and get all .dx files that are descendants of the current dir.
What are your thoughts on the current state of it?

misc/dex-completion.bash
```diff
@@ -0,0 +1,26 @@
+#/usr/bin/env bash
+
+_get_dx_files()
```
apaszke, line 3: This seems unused

apaszke (approved):
Awesome, thanks for updating it! This looks good to me, once the dead function is removed. 

I'd also suggest not focusing overly on the first word, since some flags are independent of the selected mode and can be specified before it. But it's minor, since in the current instance that will just degrade the quality of completions, without causing any incorrect suggestions.

misc/dex-completion.bash
joaogui1: yep, it was there in case we wanted to move to the second option, removed it!

joaogui1:
@apaszke done :)

PR #502, opened 2021-02-24
[Prelim] Implementation of a mini probabilistic programming language

srush:
This implements a version of a PPL through a probabilistic monad. This is roughly similar to https://hackage.haskell.org/package/probability and is based on this old paper https://web.engr.oregonstate.edu/~erwig/papers/PFP_JFP06.pdf . I wanted to target something that does more advanced inference as well (like https://dl.acm.org/doi/10.1145/3236778) but not sure if the type system is going to let us cleanly separate out the inference method into a type class.  Thought it might be fun would start with discrete variables and enumeration. 

image

(Dex is super fun for discrete variables)

### DIFF
--- a/examples/probability.dx
+++ b/examples/probability.dx
@@ -1,0 +1,194 @@
+
+' ## A Mini Probabilistic Programming Language 
+
+' This example is a Dex-style implementation of discrete probability
+ monad presented in Probabilistic Functional Programming in Haskell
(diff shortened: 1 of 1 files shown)


srush:
Huh, now that I think about it, this is maybe too haskell-y a way to do this. Going to try again with Accum and Split.

Issue #506, opened 2021-03-02
`make tests` segfaults if Fashion MNIST files are missing


perimosocordiae:
After installing dependencies and running `make` successfully, I tried `make tests` and encountered a segfault in `tests/tutorial.dx`:

Output truncated to the last few lines:
```
misc/check-quine examples/chol.dx stack exec         dex -- script --allow-errors
OK
misc/check-quine examples/fft.dx stack exec         dex -- script --allow-errors
OK
misc/check-quine examples/tutorial.dx stack exec         dex -- script --allow-errors
misc/check-quine: line 31: 220729 Segmentation fault      (core dumped) ${@:2} $1 > $tmpout 2> $errout
make: *** [makefile:122: run-examples/tutorial] Error 139
```

After looking at the source a bit I found the instructions to download the Fashion MNIST files. Once those were in place, the segfault no longer happened.

Platform: Ubuntu 20.10 64-bit

srush:
This is my fault. @dougalm is there a way to check if a file is exists?

apaszke:
I don't think we have any way to do that right now. We should add a function to the standard library

srush:
Great, I see that this is merged. 

@apaszke Any preferences on the right way to gate a section of code? Putting it in an `if` breaks top-level print. I could just create random images if the file doesn't exist, or just throw an error to prevent segfault?

apaszke:
Actually `openFile` now throws an error, so at least it shouldn't crash the whole compiler. But, if you wanted to, we could have an `if hasFile` that raises a more friendly message.

srush:
Oh that works. So maybe we should just close this issue.

perimosocordiae:
The segfault is indeed gone, and the error message indicates which file is missing, so I'm happy to close this now.

PR #507, opened 2021-03-08
Add a hasFile function to the prelude

perimosocordiae:
Very bare bones, just checks if the result of `fopen` is non-null.

See gh-506 for motivation. I'm open to better name suggestions.

### DIFF
--- a/lib/prelude.dx
+++ b/lib/prelude.dx
@@ -1408,6 +1408,13 @@ def writeFile (f:FilePath) (s:String) : {IO} Unit =
 def readFile (f:FilePath) : {IO} String =
   withFile f ReadMode \stream. fread stream
 
+def hasFile (f:FilePath) : {IO} Bool =
+  stream = fopen f ReadMode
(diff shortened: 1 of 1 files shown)


apaszke (approved):
This is great, thank you! I'm happy to merge this as is, but could you also update `withFile` to crash the program if you ever try to use a file that doesn't exist instead of returning a null pointer?

perimosocordiae:
Sure! I had `withFile` call `error` if the stream pointer is null, but that required moving the read/write functions a bit further down the prelude. I split the change into two commits to make it easier to see what was actually changed.

apaszke (commented):
Awesome, thanks! In that case, could you just move the whole section related to files down? Or are there any dependencies that force it to be interleaved like that?

perimosocordiae:
Done. The combined diff looks a little funky, but the file I/O functions are back together.

apaszke (approved):
Thanks!

PR #508, opened 2021-03-09
Bump stackage to use ghc-8.10.4

junjihashimoto:
Fixed Haskeline's MonadException to use ghc8.10.4.

### DIFF
--- a/stack-macos.yaml
+++ b/stack-macos.yaml
@@ -4,7 +4,7 @@
 # license that can be found in the LICENSE file or at
 # https://developers.google.com/open-source/licenses/bsd
 
-resolver: lts-16.26
+resolver: lts-17.5
 
 packages:
 - .
--- a/stack-macos.yaml
+++ b/stack-macos.yaml
@@ -10,8 +10,8 @@ packages:
 - .
 
 extra-deps:
-  - github: apaszke/llvm-hs
-    commit: a9a74be1a7c15f3d21b2fffd35a425002ae7736f
+  - github: junjihashimoto/llvm-hs
+    commit: 7bf182ba66c1622779f7f0c424f13c07a06bf55a
     subdirs:
       - llvm-hs
       - llvm-hs-pure
--- a/dex.cabal
+++ b/dex.cabal
@@ -1,4 +1,4 @@
-cabal-version:       2.2
+cabal-version:       3.0
 
 -- Copyright 2019 Google LLC
 --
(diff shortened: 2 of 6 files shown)


junjihashimoto:
Before merging this pr, we should be upgrade llvm-hs to support macos.

dan-zheng:
> Before merging this pr, we should fix llvm-hs to support macos.

Dex supports building with macOS via `make`, which invokes `stack` commands with [`--stack-yaml=stack-macos.yaml`](https://github.com/google-research/dex-lang/blob/bfc4c27b453cf9f3b68736f28678c315ca958932/makefile#L18).

I wonder if that works for you? Also wonder how you encountered the LLVM-related build error in the screenshot.

junjihashimoto:
I wonder if llvm-hs's `cc-options: -std:c++11` is not passed to c++ compiler.
`cc-options` should be `cxx-options`?
https://github.com/llvm-hs/llvm-hs/blob/llvm-9/llvm-hs/llvm-hs.cabal#L239

stack-macos.yaml
```diff
@@ -4,14 +4,14 @@
...
+resolver: lts-17.5
 
 packages:
 - .
 
 extra-deps:
-  - github: apaszke/llvm-hs
-    commit: a9a74be1a7c15f3d21b2fffd35a425002ae7736f
+  - github: junjihashimoto/llvm-hs
+    commit: 90f47b9881e6c4fc4d1870e44b8c1075d8fcdc23
```
apaszke, line 14: Why this change?

apaszke (commented):
Hey Junji! What's the reason for the bump? Do you have any references to issues I should look at?

stack-macos.yaml
junjihashimoto: This dump is to fix [build errors](https://github.com/google-research/dex-lang/runs/2069515204
) for llvm-hs on macos.

There are two problems and it seems that the c++ default version and its settings are not working.

First problem is that the default version of c++ to build llvm-hs is different for linux and macos.
linux uses gcc, and macos uses clang on xcode.

gcc uses c++-14 by default.
https://gcc.gnu.org/projects/cxx-status.html
xcodes may use c++-98 by default.

Second problem is that ghc-8.8 had a bug in handling c++ options.
https://gitlab.haskell.org/ghc/ghc/-/issues/16477

The bump of llvm-hs seems to have solved these problems, because llvm-hs can be compiled.
Maybe the hash of the source I used is not good. The build of the dex package fails.
image

junjihashimoto:
@apaszke 
> What's the reason for the bump? 

I just wanted to use both ghc-8.10.4 and its haskell-language-server to read dex-codes.
You don't have to bump right away. 
Sorry for bothering you.

axch:
Closing as obsolete, since Dex is now on GHC 8.10.7.

PR #509, opened 2021-03-11
Add setup.py for python bindings.

zhangqiaorjc:


python/dex/api.py
```diff
@@ -7,11 +7,10 @@
 import ctypes
 import pathlib
 import atexit
+from pkg_resources import resource_filename
 from typing import List
 
-here = pathlib.Path(__file__).parent.absolute()
-
-lib = ctypes.cdll.LoadLibrary(here / 'libDex.so')
+lib = ctypes.cdll.LoadLibrary(resource_filename('dex', 'libDex.so'))
```
apaszke, line 13: I do wonder if we need to use `pkg_resources` here, given that we really want to search within the package. A relative path might still be good enough, but I guess this is still more robust anyway 🤷‍♂️

python/setup.py
```diff
@@ -0,0 +1,14 @@
...
+setup(
+  name='dex',
+  version='0.0.1',
+  description='A research language for typed, functional array processing',
+  license='BSD',
+  author='Adam Paszke',
+  author_email='apaszke@google.com',
+  packages=find_packages(),
+  package_data={'dex': ['libDex.so']},
+  install_requires=['numpy'],
```
apaszke, line 12: Install shouldn't require numpy right? It should only be a dependency.

apaszke (approved):
Wohooo looks great! Thanks! I only wonder if we really should declare numpy as an install dependency.

python/setup.py
```diff
@@ -0,0 +1,14 @@
...
+
+setup(
+  name='dex',
+  version='0.0.1',
+  description='A research language for typed, functional array processing',
+  license='BSD',
+  author='Adam Paszke',
+  author_email='apaszke@google.com',
+  packages=find_packages(),
+  package_data={'dex': ['libDex.so']},
```
apaszke, line 11: BTW maybe we should check that this file exists either here at the top level, or by overriding some commands (e.g. the `build` command). We could give people a helpful error message that says they should run `make build-python` if it's not here. But it's fine to leave this for a follow-up.

python/setup.py
zhangqiaorjc: Looks like native_function.py imports numpy.

Also JAX used install_requires instead of a requirements.txt. Maybe this is fine?

python/dex/api.py
zhangqiaorjc: seems that pkg_resources is a safer approach

https://stackoverflow.com/questions/39104/finding-a-file-in-a-python-module-distribution
http://peak.telecommunity.com/DevCenter/PythonEggs#accessing-package-resources

```
zhangqiaorjc@qiao-gpu-1:~$ tree .local/lib/python3.8/site-packages/dex/
.local/lib/python3.8/site-packages/dex/
├── __init__.py
├── __pycache__
│   ├── __init__.cpython-38.pyc
│   ├── api.cpython-38.pyc
│   └── native_function.cpython-38.pyc
├── api.py
├── interop
│   ├── __init__.py
│   ├── __pycache__
│   │   ├── __init__.cpython-38.pyc
│   │   └── jax.cpython-38.pyc
│   └── jax.py
├── libDex.so
└── native_function.py
```

python/setup.py
zhangqiaorjc: Good suggestion! I added a check at the top of the file.

python/dex/api.py
apaszke: Cool, let's keep that!

apaszke (approved):
Thanks Qiao! This is awesome!

PR #511, opened 2021-03-20
experiment with highlighting on github

mattjj:
By adding this .gitattributes file, we can tell GitHub to highlight .dx files in this repo as any language supported by [github/linguist](https://github.com/githublinguist). [These linguist docs](https://github.com/github/linguist/blob/master/docs/overrides.md#using-gitattributes) explain how it works.

I started with Haskell and [it looks okay](https://github.com/google-research/dex-lang/tree/ea2a92731e9cda94354ce45ce4adb0e667cb913b/examples). Here's an example with GitHub's dark theme:

highlighting

But maybe misleading highlighting is worse than no highlighting, and we should just close this PR? Thoughts? Has this been discussed already?

Once there are 200+ unique `:user/:repo` repositories using Dex [we can add our own highlighting definitions to Linguist](https://github.com/github/linguist/blob/master/CONTRIBUTING.md#adding-a-language).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.dx linguist-language=haskell


apaszke:
Where did you get that example? It looks like Dex pre-last-summer (we haven't had `type` for quite a while!) Can you try out the highlighting on a newer example?

Overall I do quite like the idea. I turn on Haskell highlighting any time I edit Dex files anyway. It is slightly incorrect, but still is a net win IMO.

mattjj:
> Where did you get that example? It looks like Dex pre-last-summer (we haven't had type for quite a while!) Can you try out the highlighting on a newer example?

Ah, it was just examples/mandelbrot.dx, but I think something went wrong with my fork and it was based on an old commit. Maybe an issue with the master vs main branch name? Anyway, it should be fixed now. I updated the screenshot in the OP too.

Luckily I think you can [click here](https://github.com/google-research/dex-lang/tree/ea2a92731e9cda94354ce45ce4adb0e667cb913b/examples) to see all the examples syntax highlighted (browsing the code on this PR). (That's what the "it looks okay" link in the OP does.)

axch (approved):

PR #515, opened 2021-04-04
Query for GPU architecture before CUDA compilation

sharadmv:
This change uses a CUDA query to get the architecture instead of using a hardcoded one. I wasn't sure of the best place to put the call to CUDA (it requires a `cuInit`) so we should definitely iterate on this to find the best solution.

### DIFF
--- a/src/lib/dexrt.cpp
+++ b/src/lib/dexrt.cpp
@@ -380,6 +380,20 @@ void dex_ensure_has_cuda_context() {
   }
 }
 
+void dex_get_cuda_architecture(int device, char* arch) {
+  int majorVersion, minorVersion;
+  cuInit(0);
+  CUdevice dev;
+  CHECK(cuDeviceGet, &dev, device);
+  cuDeviceGetAttribute(&majorVersion, CU_DEVICE_ATTRIBUTE_COMPUTE_CAPABILITY_MAJOR, device);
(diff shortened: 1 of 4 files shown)


src/lib/dexrt.cpp
```diff
@@ -380,6 +380,16 @@ void dex_ensure_has_cuda_context() {
...
 }
 
+void dex_get_cuda_architecture(int device, char* arch) {
+  int majorVersion, minorVersion;
+  cuInit(0);
+  CUdevice dev;
+  CHECK(cuDeviceGet, &dev, device);
+  cuDeviceGetAttribute(&majorVersion, CU_DEVICE_ATTRIBUTE_COMPUTE_CAPABILITY_MAJOR, device);
+  cuDeviceGetAttribute(&minorVersion, CU_DEVICE_ATTRIBUTE_COMPUTE_CAPABILITY_MINOR, device);
+  sprintf(arch, "sm_%d%d", majorVersion, minorVersion);
```
apaszke, line 390: Can you please use `snprintf` to ensure there will be no overflows? Also good to check that both major and minor are between 0 and 9.

apaszke (commented):
Nice! LGTM once the `sprintf` is made safer

src/lib/dexrt.cpp
sharadmv: Done!

apaszke (approved):

apaszke (approved):

PR #516, opened 2021-04-04
Use CUDA_PATH instead of hardcoded CUDA path

sharadmv:


src/lib/LLVMExec.hs
```diff
@@ -358,6 +361,7 @@ compileCUDAKernel logger (LLVMKernel ast) = do
         usePTXAS <- maybe False (=="1") <$> lookupEnv "DEX_USE_PTXAS"
         if usePTXAS
           then do
+            ptxasPath <- (flip (++) $ "/bin/ptxas") <$> cudaPath
```
apaszke, line 364: wouldn't `(++"/bin/ptxas")` work?

sharadmv: Ah cool, didn't know you could use concat like that.

apaszke (approved):

src/lib/LLVMExec.hs
apaszke: Yeah you can do that with any binary operator. `(+2)` and `(2*)` are also valid!

PR #520, opened 2021-04-12
fix navbar not showing up in static HTML docs

oxinabox:
noted in https://github.com/google-research/dex-lang/pull/491#issuecomment-816813240
The problem was that the static documentation doesn't use the HTML from `static/dynamic.html` as a base.
Which means it was missing the `nav` bar item that the navigation stuff gets inserted into.
So it wasn't showing up.

My fix was to move the creation of that element also into the javascript.

### DIFF
--- a/static/index.js
+++ b/static/index.js
@@ -79,6 +79,14 @@ function updateNavigation() {
     }
 
     var navbarEle = document.getElementById("navbar")
+    if (navbarEle === null) {  // create it
+        navbarEle = document.createElement("div");
+        navbarEle.id="navbar";
+        navOuterEle = document.createElement("nav")
+        navOuterEle.appendChild(navbarEle);
+        document.body.prepend(navOuterEle);
+    }
+
     navbarEle.innerHTML = ""
     var structure = extractStructure()
     navbarEle.appendChild(navItemList(structure));
--- a/static/style.css
+++ b/static/style.css
@@ -42,7 +42,7 @@ nav ol {
   padding-left: 1rem;
 }
 
-main {
+#main-output {
   max-width: var(--main-width);
 }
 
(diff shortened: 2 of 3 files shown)


static/index.js
```diff
@@ -79,6 +79,14 @@ function updateNavigation() {
     }
 
     var navbarEle = document.getElementById("navbar")
+    if (navbarEle === null) {  // create it
+        navbarEle = document.createElement("div");
+        navbarEle.id="navbar";
+        navOuterEle = document.createElement("nav")
+        navOuterEle.appendChild(navbarEle);
+        document.body.prepend(navOuterEle);
```
oxinabox, line 87: it seems like javascript is a surprisingly verbose language when it comes to trying to define HTML elements.
This is the code to make `<nav><div id="navbar"></div></nav>`.
I am probably missing some better way to do it.

static/style.css
```diff
@@ -42,7 +42,7 @@ nav ol {
   padding-left: 1rem;
 }
 
-main {
+#main-output {
```
oxinabox, line 45: another difference betweem the static page created by `make docs` and the `dynamic.html` used by `dex web`
is that static uses as `div` and the dynamic uses a `main`.
But both have the id `#main-output`.
So this just makes there styling consistent.

apaszke: I mean, we can also just change the tag to always be `main`. But this is good enough

static/index.js
apaszke: Ugh. A comment might be useful, but it's fine without it too.

apaszke (approved):

apaszke:
Thanks Lyndon!

static/style.css
oxinabox: Yeah, I was going to, but I couldn't fund where the tag was defined in `RenderHtml.hs`

PR #527, opened 2021-04-14
Recursively serial copy and parallel copy for tables if possible

sharadmv:
Previously, we would try to parallel copy at the top level of `copyAtom` and fall back to `serialCopy` if it wasn't possible. This would error for `BoxedRef`s which don't work with `parallelCopy`. This changes `copyAtom` to only try to `parallelCopy` when it sees a table and recursively serialCopy otherwise.

### DIFF
--- a/src/lib/Imp.hs
+++ b/src/lib/Imp.hs
@@ -696,67 +696,69 @@ copyAtom :: Dest -> Atom -> ImpM ()
 copyAtom topDest topSrc = do
   backend <- asks impBackend
   device  <- asks curDevice
-  let optimizedCopy =
-        if backend == LLVMCUDA && device == CPU
-          then tryParallelCopy else Nothing
-  fromMaybe (serialCopy topDest topSrc) optimizedCopy
+  let canParallelize = backend == LLVMCUDA && device == CPU
+  copyRec canParallelize topDest topSrc
   where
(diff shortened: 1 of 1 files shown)


src/lib/Imp.hs
```diff
@@ -696,67 +696,69 @@ copyAtom :: Dest -> Atom -> ImpM ()
...
+            Nothing -> Nothing
+            Just [] -> Nothing  -- No point in parallelizing a scalar store
+            Just idxTys -> do
+              let (flatIndex, indexBundleDesc) = mkBundleTy idxTys
+              return $ void $ fromBuilder $ do
+                buildForAnn ParallelFor (Bind $ "pari" :> flatIndex) \pari -> do
+                  indices <- unpackBundle pari indexBundleDesc
+                  destElem <- foldM indexDest dest indices
+                  srcElem <- foldM app src indices
+                  emitOp $ PrimEffect destElem $ MPut srcElem  -- XXX: Not sure if this is well typed?
```
apaszke, line 751: Actually couldn't this just be `copyRec False destElem srcElem`?

apaszke:
Might be better to switch the commit authors to our Google email addresses to clear up the confusion with CLA bot.

sharadmv:
@googlebot I fixed it.

apaszke (approved):

PR #541, opened 2021-05-19
Kernel regression example

niklasschmitz:
Coming from #539 this adds a minimal kernel ridge regression example featuring a generic gram matrix, while borrowing the synthetic data and CG solver from the basis function regression example.

### DIFF
--- a/examples/kernelregression.dx
+++ b/examples/kernelregression.dx
@@ -1,0 +1,58 @@
+import plot
+
+-- Conjugate gradients solver
+def solve (mat:m=>m=>Float) (b:m=>Float) : m=>Float =
+  x0 = for i:m. 0.0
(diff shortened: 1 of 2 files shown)


apaszke (commented):
This is great, thank you! Can you please add this example to the `Makefile` to make sure that we're not going to break it in the future? Look for names of other examples to see how.

niklasschmitz:
I've added it to the `makefile`, and `make tests` runs without breaking (after manually `wget`-ing the required Fashion MNIST files for `tutorial.dx`)

apaszke (approved):

apaszke:
Sorry for the slowness, but it's a bit of a busy time for us. Thanks for the PR, this is great!

PR #551, opened 2021-06-15
Adds README section for setting up HLS

harryaskham:
I've not been able to test this out on Linux (LLVM is uniquely borked on my machine) so may need to revisit this once folk have tried it out.


apaszke (approved):
Thanks!

PR #555, opened 2021-06-16
Fix examples/nn.dx by fixing `grad maximum` and import statement.

### LINKED ISSUE #542: Unable to differentiate the maximum function
Context:

https://github.com/google-research/dex-lang/blob/68e07022921d898a3962337ae3154b2c101f4b48/examples/nn.dx#L141 fails at run time. Likely due to differentiating `logsoftmax` which calls `maximum` in prelude. 

Error message:
```
AD of Accum effect only supported when the base monoid is (0, +) on Float
CallStack (from HasCallStack):
  error, called at src/lib/Autodiff.hs:279:7 in dex-0.1.0.0-Aq6fzL7q3Hb4ctvwgnY3wO:Autodiff
```
Here is a snippet to reproduce the error. 
```
def myfunc (w: Float) : Float =
    arr = for i:(Fin 10). 42.0
    maximum for i. w * arr.i

(grad myfunc) 2.0
```

Potential solution:
@duvenaud and @apaszke suggested that `reduce` cause the issue and replacing `yieldAccum` may resolve the issue. 
https://github.com/google-research/dex-lang/blob/68e07022921d898a3962337ae3154b2c101f4b48/lib/prelude.dx#L737-L743

LenaMartens:
Fixed `grad maximum` by reverting `reduce` to no longer use
`yieldAccum`. Reduce can be implemented with the accumulation effect,
but AD does not currently support this effect for reductions other than
add.

Also fixes the include statement to use import instead.

Fixes #542

### DIFF
--- a/examples/nn.dx
+++ b/examples/nn.dx
@@ -1,6 +1,6 @@
 ' # Neural Networks
 
-include "plot.dx"
+import plot
 
 ' ## NN Prelude
 
--- a/lib/prelude.dx
+++ b/lib/prelude.dx
@@ -744,12 +744,8 @@ def reindex (ixr: b -> a) (tab: a=>v) : b=>v = for i. tab.(ixr i)
 -- `combine` should be a commutative and associative, and form a
 -- commutative monoid with `identity`
 def reduce (identity:a) (combine:(a->a->a)) (xs:n=>a) : a =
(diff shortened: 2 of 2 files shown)


apaszke (approved):
Thank you Lena!

PR #560, opened 2021-06-17
Add special case to fix plot errors for constant matrices.

siegeln:
Currently, when a constant matrix is passed to matshow, it sometimes displays non-constant output, and uses colors even though it should be grayscale (screenshot from the tutorial below). This is likely due to the division by zero, which may result in undesired behavior when converting to 8 bit.

This PR checks for the 0 range case explicitly, and sets all pixels to the middle color if this is the case.

Example of erroneous plotting:
\
\
f7c9cdSjqHPmick


apaszke (approved):
This was a great find, thanks!

PR #561, opened 2021-06-17
Caps CUDA architecture to 7.5, the maximum supported by LLVM 9

harryaskham:


src/lib/dexrt.cpp
```diff
@@ -391,7 +391,13 @@ void dex_get_cuda_architecture(int device, char* arch) {
     printf("Invalid CUDA architecture version: %d.%d", majorVersion, minorVersion);
     std::abort();
   }
-  snprintf(arch, 5, "sm_%d%d", majorVersion, minorVersion);
+  if (majorVersion > 7 || (majorVersion == 7 && minorVersion > 5)) {
+    printf("CUDA architecture version (%d.%d) unsupported by LLVM 9, capping at 7.5",
```
apaszke, line 395: We probably don't want to print here, because it will keep repeatedly clobbering the output. I'd remove it or at least gate it behind a static variable to make sure that this only happens once.

harryaskham: Sorry, missed this reply. Have removed the printf (it was indeed messing up the output)

apaszke (approved):

apaszke:
Thank you!

PR #567, opened 2021-06-19
Add instances for Add, Mul, and VSpace of functions.

duvenaud:
I'm using these for the SDE demo, where I add and scale drift functions.  I also like the idea of maintaining the analogy between tables and functions as closely as possible.

Pleasingly, these instances also work for functions of multiple arguments.

Also added an instance of `Arb` for `Bool`.

### DIFF
--- a/lib/prelude.dx
+++ b/lib/prelude.dx
@@ -152,6 +152,11 @@ instance Add Unit
   sub = \x y. ()
   zero = ()
 
+instance [Add a] Add (n->a)
+  add = \f g. \x. (f x) + (g x)
+  sub = \f g. \x. (f x) - (g x)
+  zero = \x. zero
+
 '#### Mul
 Things that can be multiplied.
 This defines the `Mul` [Monoid](https://en.wikipedia.org/wiki/Monoid), and its operator.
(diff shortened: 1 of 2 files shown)


apaszke (approved):
This LGTM, although I have mixed feelings about this. But I guess it makes sense mathematically, so why not 🤷‍♂️ One potential downside is that many functions that require `[VSpace a]` also assume that `a` is a representable piece of data (e.g. because they want to create references to it), but this will not work with functions.

duvenaud:
Yes, I'm not totally sure either.  I imagine that one failure mode of liberal typeclass instance definition might be that eventually every operation is defined on every type and so fewer errors are caught.  But maybe part of Dex's raison d'etre is as an academic language exploring the analogy between functions and arrays.

apaszke:
I think this is good to merge, but it has conflicts at this point 😕

oxinabox:
> One potential downside is that many functions that require [VSpace a] also assume that a is a representable piece of data (e.g. because they want to create references to it), but this will not work with functions.

I think at some point we need a separate type class for this.

and that this PR makes sense;

apaszke:
@oxinabox I agree, this extra constraint should be called `Data` and should be an interface that is provided by the compiler. This is why I don't think there are any blockers for this PR, except for conflict resolution. Ideally it would also get rebased on top of the `safe-names-dev` branch, because we have changed the syntax a bit since it was written.

duvenaud:
Yeah, I'm waiting to rebase a few PRs like this until the other tests pass on the dev branch.

duvenaud:
I rebased to `safe-names-dev` and fixed the merge conflicts.

apaszke (approved):

apaszke:
Thanks!

PR #569, opened 2021-06-20
Moved linear algebra from examples to lib, and split out linalg tests.

duvenaud:
I'm not sure if prelude should `import linalg`, but thought that linear algebra makes more sense in `lib` than `examples`. 

I also added `Add` instances for triangular tables.  Before I add any more, I wanted to check if there's a general way to specify instances for tables with dependent index types.


apaszke (approved):

apaszke (approved):
It would be nice to have a linear algebra example still, but this is good for now.

PR #576, opened 2021-06-23
Add jvp and transpose to JAX interop.

DHHZ:


DHHZ:
Added the unary transpose functionality to the PR.

DHHZ:
Just realised this was missing handling for the special `Zero` case in JAX, moving to draft while I fix it.

DHHZ:
I've updated the transpose implementation to work with non-unary functions. 

The implementation is pretty clunky, requiring a separate compiled primitive for each input argument, but it seems to work. If we get tuple types working then this logic can be cleaned up.

python/dex/interop/jax.py
```diff
@@ -189,6 +192,206 @@ def dex_call_batched(batched_args, batched_dims, func_atom):
...
+      module=env)
+
+  # Materialize jax.ad.Zero values into actual arrays of zeros.
+  tangents_no_zeros = [
+      jax.lax.zeros_like_array(arg) if type(tan) is jax.ad.Zero else tan
+      for arg, tan in zip(arg_values, arg_tangents)
+  ]
+
+  # TODO: Once we can return tuple values from Dex to Python, we might want to
+  # use a single primitive for both forward and linearized evaluation.
```
apaszke, line 264: FWIW we will always want to use two primitives, because only the second one is transposable. 

_But_, this way of doing it is a little inefficient, because `evaluate_linearized` will evaluate the full primal expression again, doubling the cost. What we really should do is linearize the primal code, but instead of returning a result and a function, we would return a result and a closure of the function (i.e. all array intermediates it needs for gradient evaluation). Then, the second primitive would take a lambda-lifted linearized function, apply it to the closure and only then to the tangent arguments. But this is not something we want to do in this PR anyway. This is fine for now.

python/dex/interop/jax.py
```diff
@@ -189,6 +192,206 @@ def dex_call_batched(batched_args, batched_dims, func_atom):
...
+  #   snd linearized (u0 u1 u2)
+  # ```
+  evaluate_linearized = eval(
+      r"\ " + arg_string("x") + " " + arg_string("u") + "."
+      f"\n  linearized = linearize jax_func_uncurried {tuple_string('x')}" +
+      f"\n  snd linearized {tuple_string('u')}",
+      module=env)
+
+  # Materialize jax.ad.Zero values into actual arrays of zeros.
+  tangents_no_zeros = [
```
apaszke, line 258: FWIW a better way to do this would be to skip the arguments with zero tangents from the call to linearize. For example, if you have `x0` and `x1`, and `x1` has a zero tangent, then you could just compute:
```
\x0 x1 u0.
  linearized = linearize (\x0. jax_func x0 x1) x0
  snd linearized u0
```

This would be more efficient, but what you have here is fine for now. Can you only please add a TODO?

python/dex/interop/jax.py
```diff
@@ -189,6 +192,206 @@ def dex_call_batched(batched_args, batched_dims, func_atom):
...
+  # use a single primitive for both forward and linearized evaluation.
+  return (
+      primitive(func_atom)(*arg_values),
+      primitive(evaluate_linearized)(*arg_values, *tangents_no_zeros),
+  )
+
+jax.interpreters.ad.primitive_jvps[dex_call_p] = dex_call_jvp
+
+
+# === transpose ===
```
apaszke, line 273: (For the future: when you have self-contained changes, it's better to split them into separate PRs. This makes reviewing much easier.)

python/dex/interop/jax.py
```diff
@@ -189,6 +192,206 @@ def dex_call_batched(batched_args, batched_dims, func_atom):
...
+
+  # Add func_atom as a variable `linearized` in the context.
+  env = api.insert(func_atom.module, api.as_cstr("linearized"), func_atom)
+
+  tangent_input_indices = [
+      i for i, t in enumerate(tangents) if jax.ad.is_undefined_primal(t)
+  ]
+  tangent_constant_indices = [
+      i for i, t in enumerate(tangents) if not jax.ad.is_undefined_primal(t)
+  ]
```
apaszke, line 310: Sorry, what are those? Why would a tangent ever be an undefined primal?

python/dex/interop/jax.py
```diff
@@ -189,6 +192,206 @@ def dex_call_batched(batched_args, batched_dims, func_atom):
...
+  # Form a lambda which partially evaluates `linearized` at the constant primal
+  # and tangent values, with the remaining arguments (which are
+  # `UndefinedPrimal`s / linear inputs) combined into a tuple, and
+  # then transpose the lambda.
+  #
+  # For a three-input primal function with constant input for the tangent
+  # parameter at index 1, the evaluated string should look like:
+  # ```
+  # \ x0 x1 x2 u1 ct.
+  #   transposeLinear (\(t0, t2). linearized x0 x1 x2 t0 u1 t2) ct
```
apaszke, line 324: What are `u`s here?

python/dex/interop/jax.py
```diff
@@ -189,6 +192,206 @@ def dex_call_batched(batched_args, batched_dims, func_atom):
...
+
+  # x0 x1 x2 t0 u1 t2
+  linearized_inputs = (
+      arg_string("x", range(num_primals)) + " " + linearized_tangent_inputs)
+
+  # \ x0 x1 x2 u1 ct.
+  #   transposeLinear (\(t0, t2). linearized x0 x1 x2 t0 u1 t2) ct
+  transposed = eval(
+      r"\ " + transposed_atom_params + ". transposeLinear (\ " +
+      linear_lambda_params + ". linearized " + linearized_inputs + ") ct",
```
apaszke, line 348: Could you use fstrings instead of string concatenation? That would make this code so much more readable.

apaszke (commented):
JVP looks good. For transpose I'd appreciate some answers before I complete the review. I'd also be best to split them into two PRs to simplify the review.

python/dex/interop/jax.py
DHHZ: Good point! Returning a closure would be a neat trick.

DHHZ:
Thanks for the comments! I think both parts could be simplified with tuple return types, so I'll move this back to draft until that PR goes in.

python/dex/interop/jax.py
DHHZ: Done.

DHHZ:
Cleaned up, removed the transpose op and addressed the comments, PTAL.

python/dex/interop/jax.py
```diff
@@ -189,6 +192,84 @@ def dex_call_batched(batched_args, batched_dims, func_atom):
...
+  env = api.insert(func_atom.module, api.as_cstr("jax_func"), func_atom)
+
+  # `linearize` only seems to work properly for functions which take a single
+  # input argument, so we uncurry `func_atom` to make it into this form. The
+  # evaluated string for three function arguments should look like:
+  # ```
+  # \ (x0, x1, x2). jax_func x0 x1 x2
+  # ```
+  uncurried = eval(
+      f"\\ {tuple_string('x')}. jax_func {arg_string('x')",
```
apaszke, line 234: isn't this missing a closing brace? We don't have Python wired up in CI 🤦‍♂️

python/tests/jax_test.py
```diff
@@ -81,6 +81,50 @@ def test_vmap_jit_nonzero_index(self):
     np.testing.assert_allclose(
         jax.jit(jax.vmap(add_two, in_axes=1, out_axes=1))(x), x + 2.0)
 
+  def test_vjp(self):
```
apaszke, line 84: this tests the jvp, not vjp

python/tests/jax_test.py
```diff
@@ -81,6 +81,50 @@ def test_vmap_jit_nonzero_index(self):
...
+    u = jnp.linspace(0.1, 0.3, num=10)
+    v = jnp.linspace(2.0, -5.0, num=10)
+
+    output_dex, tangent_dex = jax.jvp(f_dex, (x, y), (u, v))
+    output_jax, tangent_jax = jax.jvp(f_jax, (x, y), (u, v))
+
+    np.testing.assert_allclose(output_dex, output_jax)
+    np.testing.assert_allclose(tangent_dex, tangent_jax)
+
+  def test_vjp_jit(self):
```
apaszke, line 103: also a jvp test

apaszke (commented):
Did the tuple PR actually simplify anything here? Or is it mostly useful for transpose?

python/tests/jax_test.py
DHHZ: Yep, done.

python/tests/jax_test.py
DHHZ: Done

DHHZ:
> Did the tuple PR actually simplify anything here? Or is it mostly useful for transpose?

No, I thought it might but I was wrong, it's just useful for the transpose. We could have used it to return the both primal and tangent computation using a single compiled function call, but as you said in a previous comment it wouldn't have been transposable in the same way.

apaszke (approved):
Looks great now. Thank you!

PR #580, opened 2021-06-25
Add pair result types to export

DHHZ:
Small patch to add pair types to the function signature exporter. 

All the other pieces to translate tuple return types to Python were already in place.

### DIFF
--- a/src/lib/Export.hs
+++ b/src/lib/Export.hs
@@ -143,7 +143,7 @@ prepareFunctionForExport env nameStr func = do
           elemTy' <- substBuilder (b@>Var i) elemTy
           createDest (idx <> Nest (Bind i) Empty) elemTy'
         return (Con $ TabRef destTab, exportResult)
-      PairTy a b -> do
+      PairTy a b | idx == Empty -> do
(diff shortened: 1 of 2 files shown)


src/lib/Export.hs
```diff
@@ -143,6 +143,10 @@ prepareFunctionForExport env nameStr func = do
           elemTy' <- substBuilder (b@>Var i) elemTy
           createDest (idx <> Nest (Bind i) Empty) elemTy'
         return (Con $ TabRef destTab, exportResult)
+      PairTy a b -> do
```
apaszke, line 146: Can you add a check that makes sure that pair types only appear on the outermost level of recursion? I think it's enough to add `| idx == Empty` after `PairTy a b` (to add an extra guard that we're not recursing through any table arrows).

apaszke (approved):
Wow that looks great. LGTM but please add a check

src/lib/Export.hs
DHHZ: Yep, done.

apaszke:
Perfect, thank you!

PR #581, opened 2021-06-25
Neural network layers example

RobertStanforth:


apaszke (commented):
While I think that having both approaches is nice, it would be great to deduplicate the files a little bit. Maybe let's only keep one instance of the training loop?

PR #583, opened 2021-06-25
Add transpose primitive to JAX interop.

DHHZ:
I found the JAX transpose mechanism quite hard to follow, I've tried adding a bunch of comments to explain but please let me know if anything doesn't make sense, or doesn't fit with your understanding of transpose in JAX.

### DIFF
--- a/python/dex/interop/jax.py
+++ b/python/dex/interop/jax.py
@@ -271,5 +271,128 @@ def dex_call_jvp(arg_values, arg_tangents, func_atom):
 
 jax.interpreters.ad.primitive_jvps[dex_call_p] = dex_call_jvp
 
-# TODO
-# jax.interpreters.ad.primitive_transposes[self.primitive] = ...
+# === transpose ===
+
+# alias to avoid confusion around overloading of "primal".
+_is_linear_input = jax.ad.is_undefined_primal
+
+
+def dex_call_evaluate_linearized_transpose(cotangents, *args, func_atom):
+  """Evaluates the transpose of a function atom.  """
+
+  # `func_atom` is assumed to be of the form of `evaluate_linearized` from
+  # `dex_call_jvp`, applied to a some function atom, called `f`, say.
+  # Concretely, if `f` has three primal arguments, `func_atom` should look like:
+  # ```
+  # \ x0 x1 x2 u0 u1 u2.
+  #   intermediate_linearized = linearize f (x0, x1, x2)
+  #   snd intermediate_linearized (u0 u1 u2)
+  # ```
+  # In particular, its arguments are assumed to be `num_primals` primal inputs,
+  # followed by `num_primals` tangent inputs.
+
+  assert len(args) % 2 == 0
+  num_primals = len(args) // 2
+
+  primals, tangents = args[:num_primals], args[num_primals:]
+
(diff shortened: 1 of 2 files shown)


python/dex/interop/jax.py
```diff
@@ -271,5 +271,127 @@ def tuple_string(prefix):
...
+  # Add `func_atom` as a variable `linearized` in the context.
+  env = api.insert(func_atom.module, api.as_cstr("linearized"), func_atom)
+
+  # Form lists of the indices in `tangents` which correspond to linear inputs
+  # (which we are expected to transpose w.r.t.) and constant inputs (which we
+  # are not).
+  tangent_input_indices = [
+      i for i, t in enumerate(tangents) if _is_linear_input(t)
+  ]
+  tangent_constant_indices = [
```
apaszke, line 327: Is that ever non-empty? I don't think so? Or when you say "constant inputs" do you mean the zeros that you've instantiated in the JVP? If yes then let's make the comment more precise.

python/dex/interop/jax.py
```diff
@@ -271,5 +271,127 @@ def tuple_string(prefix):
...
+
+  # Form a lambda which partially evaluates `linearized` at the constant primal
+  # and constant tangent values, with the remaining arguments (i.e. the linear
+  # input tangents) combined into a tuple, and then transpose the lambda.
+  #
+  # For a three-input primal function with constant input for the tangent
+  # parameter at index 1, the evaluated string should look like:
+  # ```
+  # \ x0 x1 x2 u1 ct.
+  #   transposeLinear (\(t0, t2). linearized x0 x1 x2 t0 u1 t2) ct
```
apaszke, line 344: This works, although one downside of this trick is that it effectively makes all Dex primitives do checkpointing as if they were decorated with `jax.remat`.

apaszke (commented):
Looks very good! I only have one minor question and a potential challenge to efficiency, but this is pretty much ready to merge.

python/dex/interop/jax.py
DHHZ: It's the latter, added a further comment to clarify.

The linear / constant input terminology I was using comes from [here](https://jax.readthedocs.io/en/latest/notebooks/How_JAX_primitives_work.html#transposition), if there's a better way to state this please feel free to suggest something!

python/dex/interop/jax.py
DHHZ: Right, I see. It seems to me that in order to avoid this we would need to do the closure trick you talked about in the previous PR, that is to extract all intermediate arrays from the computation of `linearized` in the JVP op in order to re-use them here. I'm not sure how we would extract these intermediate arrays with the current version of the Dex foreign API.

Is there a way we could do this? Or am I missing something and is there another way to avoid the checkpointing behaviour?

apaszke: Yes, this improvement would require that API and I don't think that we can emulate it in any way.

python/dex/interop/jax.py
apaszke: Looks great, thanks!

apaszke (approved):
Awesome, thanks for adding the comment.

PR #584, opened 2021-06-26
Removed unnecessary Eq constrains from linear algebra functions.

duvenaud:
They were getting in the way of easy polymorphism over `VSpace`, and I noticed they were unnecessary.


apaszke (approved):

apaszke:
Thanks David! We should really make `Eq` a superclass of `IdxSet` at some point, since you can always compare for equality after casting to ordinals.

PR #586, opened 2021-06-30
CKY CRF example

srush:
Preliminary implementation of CKY CRF with constraints for @duvenaud 

Wrote this code a while ago, now trying to clean it up to run under the state effect and use helpers to abstract some of the nasty indexing functions. 

Would really like to make this a monoid, but that will require some List acrobatics. Not sure if that is worth it.

### DIFF
--- a/examples/cky.dx
+++ b/examples/cky.dx
@@ -1,0 +1,94 @@
+
+' Implementation of the CKY Algorithm
+
+length = 10 
+Label = Fin 20
+WordPos = Fin 10
+
+' ## Span manipulation
+
+' Shift an index x in space i... to an index in j... (assumed possinble) 
+
+def shift (i: a) ?-> (j: a) ?-> (x:(i<..)) : (j<..) =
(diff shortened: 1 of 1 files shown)


apaszke (commented):
The `Add` instance for indices is questionable, but I'll allow it 😛 Would you rather merge it now or once the grad stuff is fixed?

axch:
I don't know whether this is still under active development, but while trying to reproduce #587 I ported the code to current Dex syntax.  The main difference is that function arguments are now "required" as far as the function definition is concerned, but one can pass `_` for an argument that can be inferred at the call site.  Here is that port, in case it's helpful.
[cky.dx.txt](https://github.com/google-research/dex-lang/files/8721964/cky.dx.txt)

apaszke:
@axch does the example complete without errors now? If so, we should submit it!

srush:
Neat. I can do a review if that is helpful. I am hoping to finish some of these up this summer. Got a bit busy during the term and haven't caught up on language changes

PR #588, opened 2021-07-02
Safer names - second installment

dougalm:


src/lib/SaferNames/NameCore.hs
```diff
@@ -0,0 +1,294 @@
...
+instance Ord (Name s n) where
+
+instance Show (Name s n) where
+  show (UnsafeMakeName rawName) = show rawName
+
+
+-- === unsafe coercions ===
+
+-- Sometimes we need to break the glass. But at least these are slightly safer
+-- than raw `unsafeCoerce` because at the checks the kind
```
danieldjohnson, line 291: typo here

src/lib/SaferNames/Name.hs
```diff
@@ -4,267 +4,113 @@
...
+      -> NameMap v (i:=>:i') o          -- names first lookup up here
+      -> Env v i' o
+
+newtype EnvFrag (v::E->E) (i::S) (i'::S) (o::S) =
+  EnvFrag (NameMap v (i:=>:i') o)
+
+-- Scopes carry
+--    1. The full set of names (suitable for generating fresh names)
+--    2. A no-shadowing guarantee (via `Distinct`)
+--    3. The payload associated with each name's static data parameter
```
danieldjohnson, line 57: Where does the payload appear? I assumed that would be an `s`? but I don't see one in the type signature? Is this just because it's unsafe and we have to do a runtime check?

src/lib/SaferNames/Name.hs
```diff
@@ -4,267 +4,113 @@
...
+(<>>) (Env f frag) (EnvFrag frag') = Env f $ concatNameMaps frag frag'
+
+infixl 1 <.>
+(<.>) :: EnvFrag v i1 i2 o -> EnvFrag v i2 i3 o -> EnvFrag v i1 i3 o
+(<.>) (EnvFrag env) (EnvFrag env') = EnvFrag (concatNameMaps env env')
+
+emptyEnvFrag :: EnvFrag v i i o
+emptyEnvFrag = EnvFrag emptyNameMap
+
+lookupEnvFrag :: EnvFrag v i i' o -> Name s i' -> Either (Name s i) (v s o)
```
danieldjohnson, line 85: understanding check: we either prove that this name wasn't in the fragment (`Left name'`) or we have a value from the fragment `Right (lookupNameMap m name) :: v s o`

src/lib/SaferNames/Name.hs
```diff
@@ -4,267 +4,113 @@
...
+  Env :: (forall s. Name s i -> v s o)  -- fallback function
+      -> NameMap v (i:=>:i') o          -- names first lookup up here
+      -> Env v i' o
+
+newtype EnvFrag (v::E->E) (i::S) (i'::S) (o::S) =
+  EnvFrag (NameMap v (i:=>:i') o)
+
+-- Scopes carry
+--    1. The full set of names (suitable for generating fresh names)
+--    2. A no-shadowing guarantee (via `Distinct`)
```
danieldjohnson, line 56: So we're forbidding shadowing in scopes, presumably by rewriting stuff? So this is kinda an invariant we (hopefully provably) are maintaining as we go?

src/lib/SaferNames/Name.hs
```diff
@@ -4,267 +4,113 @@
...
 
 -- === common scoping patterns ===
 
 data Abs (binder::B) (body::E) (n::S) where
   Abs :: binder n l -> body l -> Abs binder body n
 
-data FreshAbs (binder::B) (body::E) (n::S) where
-  FreshAbs :: FreshExt n l -> binder n l -> body l -> FreshAbs binder body n
-
 data NestPair (b1::B) (b2::B) (n::S) (l::S) where
```
danieldjohnson, line 103: What's an example of a `NestPair`?

danieldjohnson (commented):
Started trying to understand the name core. I feel like I'm still missing a lot of pieces of understanding. It might help to have more documentation or examples of what the different type parameters mean and what the "shadowing"/"extending"/"freshness" claims refer to.

src/lib/Err.hs
```diff
@@ -97,6 +103,21 @@ layout :: LayoutOptions
 layout = if unbounded then LayoutOptions Unbounded else defaultLayoutOptions
   where unbounded = unsafePerformIO $ (Just "1"==) <$> lookupEnv "DEX_PPRINT_UNBOUNDED"
 
+-- === zippable class ===
+
+-- These are in this module because they need a way to throw errors. Is there a
+-- monad type class for the maybe-like things, with `throw :: m a` we could
+-- use instead?
```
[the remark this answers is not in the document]
dougalm: MonadFail seems like it's intended for may-fail patterns in do-notation specifically but it certainly has the right type. Maybe we should just repurpose it.

src/lib/SaferNames/NameCore.hs
```diff
@@ -0,0 +1,294 @@
...
+
+import qualified Env as D
+
+-- `S` is the kind of "scope parameters". It's only ever used as a phantom type.
+-- It represents a list of names, given by the value of the singleton type
+-- `NameSet n` (`n::S`). Names are tagged with a scope parameter, and a name of
+-- type `Name n` has an underlying raw name that must occur in the corresponding
+-- `Scope n`. (A detail: `NameSet n` actually only carries a *set* of names, not
+-- a list, because that's all we need at runtime. But it's important to remember
+-- that it conceptually represents a list. For example, a `NameSet n` and a `NameSet
```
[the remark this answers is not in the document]
dougalm: Yes, left-to-right order corresponds to outside-to-inside scope order. So in `\x. \y. \z. \x. ...`, the `...` has the scope `[x, y, z, x]`. Names are looked up from right to left (i.e. the usual lexical scoping rule). When I say "exposed name", later, I mean the rightmost name.

src/lib/SaferNames/NameCore.hs
```diff
@@ -0,0 +1,294 @@
...
+  UnsafeMakeNameSet (s1 <> s2)
+
+-- TODO: we reuse the old `Name` to make use of the GlobalName name space while
+-- we're using both the old and new systems together.
+-- TODO: something like this instead:
+--    type Tag = T.Text
+--    data RawName = RawName Tag Int deriving (Show, Eq, Ord)
+type RawName = D.Name
+
+data Name (s::E)  -- static information associated with name
```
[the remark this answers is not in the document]
dougalm: Yes, static information is what we currently call `BinderInfo`. It includes type info , but also things like the flavor of arrow if it's a lambda-bound variable, or the actual rhs of the let binding if it's let-bound. `s` has kind `S -> *` because it's parameterized by a name space. For example, if the static information is a type, then it may contain free variables.

src/lib/SaferNames/NameCore.hs
```diff
@@ -0,0 +1,294 @@
...
+          (n::S)  -- scope parameter
+  where
+    UnsafeMakeName :: Typeable s => RawName -> Name s n
+
+data NameBinder (s::E)  -- static information for the name this binds (note
+                         -- that `NameBinder` doesn't actually carry this data)
+                 (n::S)  -- scope the binder lives in
+                 (l::S)  -- scope within the binder's scope
```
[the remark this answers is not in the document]
dougalm: Yes, exactly. The `Injectable` instance for `NameBinder` means that you can get a `NameSet (n:=>:l)` from a `NameBinder s n l`.

src/lib/SaferNames/NameCore.hs
```diff
@@ -0,0 +1,294 @@
...
+         -> (forall l. Distinct l => NameBinder s n l -> Name s l -> a) -> a
+withFresh (UnsafeMakeNameSet scope) cont =
+  cont @UnsafeMakeDistinctS (UnsafeMakeBinder freshName) (UnsafeMakeName freshName)
+  where
+    freshName = freshRawName "v" scope
+
+freshRawName :: D.Tag -> S.Set RawName -> RawName
+freshRawName tag usedNames = D.Name D.GenName tag nextNum
+  where
+    nextNum = case S.lookupLT (D.Name D.GenName tag bigInt) usedNames of
```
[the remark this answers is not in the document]
dougalm: Actually this is unchanged from the old implementation. It's just a way to generate a `(String,Int)` pair that doesn't occur in the give `Set (String, Int)`.

src/lib/SaferNames/NameCore.hs
```diff
@@ -0,0 +1,294 @@
...
+-- containing more in-scope variables. For example, when we turn this:
+--   let foo = \x. x + z
+--   in \y. foo y
+-- into this:
+--   \y. (\x. x + z) y
+-- The expression `\x. x + z`, initially in the scope `[z]`, gets injected into
+-- the scope `[z, y]`. To prove that we can do the injection as an coercion,
+-- it's sufficient to know (1) that the new scope contains distinct names, and
+-- (2) it extends the old scope. These are the `Distinct l` and
+-- `(InjectableB b, b)` conditions below.
```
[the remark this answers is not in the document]
dougalm: Exactly. It's sufficient but not necessary. Until yesterday, I'd been working with the weaker condition I'd called `FreshExt`, which says that you can inject a name from `n` to `l` if you know that `l` extends `n` with fresh names. But it turned out that in practice, we always met the stronger condition that the names in `l` were unique in an absolute sense, not just a relative one. And the single-parameter `Distinct n` is a bit easier to work with than the two-parameter `FreshExt n l`, so I switched to that.

src/lib/SaferNames/NameCore.hs
```diff
@@ -0,0 +1,294 @@
...
+            => Distinct l => b n l -> e n -> e l
+injectNames _ x = unsafeCoerceE x
+
+-- `injectNames` is the only function we actualy use in practice. The rest of
+-- this section is just about proving that it's valid to use `injectNames`, by
+-- offering tools to implement `Injectable` instances.
+
+-- To prove that a data type is injectable by coercion, we have to show that
+-- it's injectable without using coercion. `injectionProof` is never actually
+-- called at run time, but it should still be implemented to prove that it's
```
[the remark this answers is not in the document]
dougalm: Interesting idea. Definitely worth thinking about. Though tbh I'd been planning to leave most of the instance implementations as `undefined` until someone feels like filling them out to be sure.

src/lib/SaferNames/NameCore.hs
```diff
@@ -0,0 +1,294 @@
...
+                  -> (forall l'. ObservablyFresh l l' -> b n' l' -> a)
+                  -> a
+
+  -- Not sure if this belongs here, but it doesn't make sense in `HasNamesB`
+  -- either. Maybe we need a separate type class?
+  boundNames :: b n l -> NameSet (n:=>:l)
+
+-- an `ObservablyFresh n l` means that scope `l` is a superset of scope `n` that
+-- doesn't shadow any of the names exposed by `n`.
+data ObservablyFresh (n::S) (l::S) = UnsafeMakeObservablyFresh
```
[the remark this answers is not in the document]
dougalm: Right, it's a weaker condition that's handy for the recursive calls.

By "expose a name" I mean "present the name as the rightmost/innermost/lexically-available version of that name". What do I mean by "version"? Good question. You've caught me fudging things here and I'd like to find a way to be more precise about this.

I've said that a scope is just a list of names, but the `ObservablyFresh` predicate only makes sense if we consider names to have some kind of "identity" alongside their actual value. It's easy to describe if we imagine that all the variables happen to have different types. For example:

```
-- l is observably fresh wrt n
n = [x:Int,         y:Float]
l = [x:Int, y:Bool, y:Float, z:Int]

-- l not observably fresh wrt n. The "new" `y:Bool` shadows the old `y:Float`.
n = [x:Int, y:Float]
l = [x:Int, y:Float, y:Bool, z:Int]

```

This comes up if we imagine inlining `foo` in the following expression:

```
x:Int = 1
foo = (\y:Float. x + y)
y:Bool = True
z:Int = 2
foo
```

Before inlining, the `x + y` expression is in the scope `[x:Int, y:Float]`. After inlining, it's in the larger (deeper) scope `[x:Int, y:Bool, z:Int, y:Float]`. But it's ok, because even though the new scope adds an extra non-unique name, `y:Bool`, it's shadowed by the one we care about, the `y:Float`.

Does that make sense? I'd love a way to describe this precisely without resorting to unique identities or types that happen to be different.

src/lib/SaferNames/NameCore.hs
```diff
@@ -0,0 +1,294 @@
...
+
+instance InjectableE (Name s) where
+  injectionProofE _ name = unsafeCoerceE name
```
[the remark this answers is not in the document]
dougalm: yeah, this is just the unsafe base case. The instances defined in Syntax.hs are the ones that we implement safely. I'll add a comment.

src/lib/SaferNames/NameCore.hs
```diff
@@ -0,0 +1,294 @@
...
+
+-- Here's the reasoning informally. Let's say `x` is the name of the binder.
+-- The scopes are related like this:
+--   l  = n  ++ [x]
+--   l' = n' ++ [x]
+-- if n' is observably fresh wrt n then l' is observably fresh wrt l. This is because every
+-- exposed name in l' is either
+--   (1) x itself, which is exposed in both l and l', thus l' doesn't shadow it, or
```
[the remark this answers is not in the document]
dougalm: yep, you're completely right (and see comment above too). If you have any suggestions for pinning down this idea of `x` being "the same" `x` I'd love to hear them.

src/lib/SaferNames/NameCore.hs
```diff
@@ -0,0 +1,294 @@
...
+-- === environments ===
+
+-- The `NameMap` type is purely an optimization. We could do everything using
+-- the safe API by defining:
+--    type NameMap v i o = (NameSet i, forall s. Name s i -> v s o)
+-- Instead, we use this unsafely-implemented data type for efficiency, to avoid
+-- long chains of case analyses as we extend environments one name at a time.
+
+data NameMap
+  (v::E -> E)  -- env payload, as a function of the static data type
```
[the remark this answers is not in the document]
dougalm: Oh, by "payload" I just mean "the thing you get when you look up a name in the env". In the case of a substitution, it's an `Atom`. Or it might be a name. Or a type. What's new is that the type of the thing you get when you query an env can now depend on the type parameter associated with the name. So you can have different sorts of names, like data definition names and let/lambda-bound names, that give you different things when you look them up in an env.

src/lib/SaferNames/NameCore.hs
```diff
@@ -0,0 +1,294 @@
...
+extendNameMap (UnsafeMakeNameMap m1 s1) (UnsafeMakeNameMap m2 s2) =
+  UnsafeMakeNameMap (m2 <> m1) (s2 <> s1)
+
+nameMapNames :: NameMap v i o -> NameSet i
+nameMapNames (UnsafeMakeNameMap _ s) = UnsafeMakeNameSet s
+
+-- === handling the dynamic/heterogeneous stuff for Env ===
+
+data EnvVal (v::E->E) (n::S) where
+  EnvVal :: TypeRep s -> v s n -> EnvVal v n
```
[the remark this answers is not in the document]
dougalm: Ha, you're right that `n::S` and `e::E` are reasonable enough but `v :: E -> E` starts to get a bit silly. There are only two instantiations of `v:: E -> E` so far, and they might clarify things. There's the identity map, which is used by `Scope` in `Name.hs`. It just says that if you have a `Name s n` you can query the scope to get a (newtype-wrapper) `s n`. Concretely, in the core IR we have `Name TypedBindeInfo n` for ordinary let/lambda-bound names, and `Name DataDef n` for data definitions. But I imagine we might want to start having names for modules, for pointers, for transformation specializations and so on. Then there's `SubstVal` which plays a GADT trick to say that you have two cases -- either the `s` parameter of the name you're looking up matches a particular type, in which case you get, say, an `Atom`, or else it doesn't, in which case you merely get a new name.

src/lib/SaferNames/Name.hs
dougalm: It's `s n` for any `Name s n`. Because it's universally quantified, the type doesn't appear in the definition of `Scope`.

src/lib/SaferNames/Name.hs
dougalm: yep, exactly

src/lib/SaferNames/Name.hs
dougalm: yes, that's right. And having to meet a `Distinct n` constraint whenever we produce a new scope is the mechanism that should keep us honest.

src/lib/SaferNames/Name.hs
dougalm: I'm not sure if we have any in our ASTs. `NestPair b1 b2 n l` is to `Nest b n l` as `(a1, a2)` is to `[a]`, so it seems appropriate to have it for completeness. At various times I've found it useful to create a `NestPair` on the fly to use its instances.

src/lib/SaferNames/NameCore.hs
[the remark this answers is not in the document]
danieldjohnson: Ah, that makes sense. Might be helpful to include that explanation as a comment?

src/lib/SaferNames/NameCore.hs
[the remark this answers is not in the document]
danieldjohnson: I think I get what you mean. I'm a bit worried about the encoding, though. It seems like whether something is observably fresh depends on a particular interpretation of `n` and `l`, but that interpretation isn't checked by the Haskell compiler. So I'd be a bit concerned that maybe we'd misinterpret the object as proving some property that doesn't actually hold.

I can't immediately think of a better way to encode this, though...

Maybe it would make sense at least for understanding purposes if we thought of a nameset as an ordered *mapping*, instead of an ordered list? Or at least the type-level metadata for such a mapping?

src/lib/SaferNames/NameCore.hs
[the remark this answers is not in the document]
danieldjohnson: Thanks, that's helpful. Does the `E -> E` pattern appear in lots of other places, or is it just in `EnvVal`? I wonder if it would be worth having an alias for it with documentation saying what some examples look like? Or just have more examples listed here?

src/lib/SaferNames/Name.hs
danieldjohnson: Ah, so this is how you say there are two different kinds of binders that are being composed. And I guess `NestPair b1 b2` is itself a `B`?

(I wonder if we could have more descriptive names for `E` and `B`? But I guess it's nice to not have to type it everywhere...)

src/lib/SaferNames/NameCore.hs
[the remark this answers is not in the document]
danieldjohnson: Ok, fleshing out that idea a bit more. It sounds like `Name`s already "contain" some other static information `s`. And that presumably uniquely identifies what it means to be the "same"? Or at least it could, in principle, if we chose `s` correctly. So would it be correct to say:
- there's always exactly one mapping from names to their static information (for any `v`?)
- the `n :: S` phantom type parameter uniquely identifies this mapping, so that if you have a `NameMap v n` it must be the "right" one (carefully maintained using unsafe stuff). Essentially, we say that every NameMap is semantically a singleton object that gives us run-time access to the mapping described by `n`.
- an `ObservablyFresh n l` is a proof that, if you have a `NameMap v l`, and a `NameMap v n`, then the second is ALWAYS a sub-mapping of the first; equivalently, you can pretend your `NameMap v l` is the singleton instance of `NameMap v n` without changing the results of any lookup.
- a `NameSet n` is just a cheap representation of the keys of the `NameMap`, but it's not the fundamental object we are reasoning about. Maybe it's isomorphic to the singleton `NameMap UnitV n` or something that maps every name to the unit. Of course, it's got to follow the same guarantees above, but that's kind of meaningless because every payload is identical so ObservablyFresh doesn't mean anything important here.

I think this is true based on what you told me in person? But maybe it makes sense for this to be the explicit formalism we are operating under, as well.

(On the other hand, I'm not sure I know what it means for this interpretation of a name set to be "distinct". Maybe I'm still missing something, or maybe this interpretation isn't quite right either.)

src/lib/SaferNames/NameCore.hs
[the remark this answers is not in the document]
danieldjohnson: Actually, I'm still a bit confused about this. Why do we need `InjectableE` or `InjectableB` in the definition below? Isn't it enough to just know that every name is distinct, and that there exists some binder witnessing the extension?

src/lib/SaferNames/Name.hs
```diff
@@ -4,288 +4,206 @@
...
+instance BindsNames Ext where
+  toExtVal = id
+
+toNameSet :: BindsNames b => b n l -> NameSet (n:=>:l)
+toNameSet b = fromExtVal $ toExtVal b
+
+inject :: BindsNames b => InjectableE e => Distinct l => b n l -> e n -> e l
+inject ext x = injectNames (toNameSet ext) x
+
+-- like inject, but uses the SopeReader monad for its `Distinct` proof
```
apaszke, line 137: nit: typo in ScopeReader

src/lib/SaferNames/Name.hs
```diff
@@ -4,288 +4,206 @@
...
+fromConstAbs :: ScopeReader m => MonadErr1 m => BindsNames b => HasNamesE e
+             => Abs b e n -> m n (e n)
+fromConstAbs (Abs b e) = do
+  let t = newNameTraversal \name ->
+             case projectName (toNameSet b) name of
+               Left name' -> Just name'
+               Right _    -> Nothing
+  scope <- askScope
+  case traverseNamesE scope t e of
+    Nothing -> throw EscapedNameErr ""
```
apaszke, line 150: Add more context? Otherwise we'll have a great time debugging those. Also, it seems a little inefficient that we traverse the expression every time we want to use this, because I expect that we'll be doing this all over the place. If we don't have a better idea then let's at least mark this as a potential hot spot.

src/lib/SaferNames/Name.hs
```diff
@@ -4,288 +4,206 @@
...
+  withFreshM ann \b -> do
+    body' <- injectM b body
+    return $ Abs b body'
+
+-- === type classes for traversing names ===
+
+data NameTraversal m i o where
+  NameTraversal :: (forall s. Name s i -> m (Name s o)) -- monadic function for free vars
+                -> EnvFrag Name i i' o                  -- renaming local vars
+                -> NameTraversal m i' o
```
apaszke, line 168: The fact that you use `i` in the `data` declaration, and the `i'` for the same thing in the constructor definition (and `i` to mean something else) is _extremely_ confusing. Can you please rename those?

apaszke: Also can you please add docs for what `i` and `o` are supposed to be? Are they the scopes of free variables? "renaming local vars" is also insufficient for me to understand what's going on. Is `i'` the local scope? Why is it wrapped as an existential thing in a traversal?

apaszke: If so, perhaps rename `i'` to `l` for local? Or just use longer names here? No one forces us to use one-letter aliases and while it's justifiable in many function signatures, they do have special meaning here

src/lib/SaferNames/Syntax.hs
```diff
@@ -107,43 +100,53 @@ data Expr n =
...
 
-data DataConDef n = DataConDef RawName (Abs (Nest Binder) UnitE n)
+-- As above, the `RawName` is just for pretty-printing
+data DataConDef n = DataConDef RawName (EmptyAbs (Nest Binder) n)
                     deriving Show
 
-data NamedDataDef n = NamedDataDef (Name n) (DataDef n)
-                      deriving Show
+data Block n where
+  Block :: Type n -> Nest Decl n l ->  Expr l -> Block n
```
apaszke, line 135: What's this type? Is it the return type? Why can't we query the result expr for that anymore?

src/lib/SaferNames/Syntax.hs
```diff
@@ -154,135 +157,62 @@ type Con n = PrimCon (Atom n)
...
-  { fromSourceNameMap :: M.Map SourceName (Name n)}
+  { fromSourceNameMap :: M.Map SourceName (Name UnitE n)}
 
 data Module n where
   Module :: IRVariant
          -> Nest Decl n l
-         -> BindingsFrag l l'
+         -> ScopeFrag l l'
          -> SourceNameMap l'
          -> Module n
```
apaszke, line 176: I know this was checked in before, but can you please document the fields here? I don't remember what they mean anymore

src/lib/SaferNames/Syntax.hs
```diff
@@ -107,43 +100,53 @@ data Expr n =
...
                 deriving (Show)
 
-data DataConRefBinding n l = DataConRefBinding (Binder n l) (Atom n)
-                             deriving Show
+type AtomName   = Name TypedBinderInfo
 
-type AltP (e::E) = Abs (Nest Binder) e :: E
-type Alt = AltP Block                  :: E
+data Binder (n::S) (l::S) =
+  (:>) (NameBinder TypedBinderInfo n l) (Type n)  deriving Show
```
apaszke, line 118: I'm a bit confused as to why do we put the type both in the binder and in the `TypedBinderInfo`. Is it just to mimic how we used to construct `Bindings` before this change?

src/lib/SaferNames/Syntax.hs
```diff
@@ -154,135 +157,62 @@ type Con n = PrimCon (Atom n)
...
-type IndexStructure = Nest Binder
-
-getBaseMonoidType :: Scope n -> Type n -> Type n
-getBaseMonoidType scope ty = case ty of
-  TabTy i b -> case projectNamesL scope (i@>()) b of
-    Just b' -> b'
-    Nothing -> error "Can't use a dependent table as a monoid"
-  _         -> ty
+data EvaluatedModule (n::S) where
+  EvaluatedModule :: ScopeFrag n l -> SourceNameMap l -> EvaluatedModule n
```
apaszke, line 179: Please document the fields here as well

src/lib/SaferNames/Syntax.hs
```diff
@@ -154,135 +157,62 @@ type Con n = PrimCon (Atom n)
...
 
 -- === effects ===
 
-data EffectRow n = EffectRow (S.Set (Effect n)) (Maybe (Name n))
-                   deriving Show
+data EffectRow n = EffectRow (S.Set (Effect n)) (Maybe (AtomName n))
```
apaszke, line 183: Should the extensions of effect rows really be atom names?

src/lib/SaferNames/Syntax.hs
```diff
@@ -154,135 +157,62 @@ type Con n = PrimCon (Atom n)
...
 
-(-->) :: Type n -> Type n -> Type n
-a --> b = Pi (Abs (Ignore:>a) (WithArrow PureArrow b))
+nonDepPiType :: ScopeReader m => Arrow -> Type n -> EffectRow n -> Type n -> m n (Type n)
+nonDepPiType arr argTy eff resultTy =
+  toConstAbs (TypedBinderInfo argTy PiBound) (PairE eff resultTy) >>= \case
+    Abs b (PairE eff' resultTy') ->
+      return $ Pi $ PiType arr (b:>argTy) eff' resultTy'
+
+(?-->) :: ScopeReader m => Type n -> Type n -> m n (Type n)
```
apaszke, line 205: Why are those functions monadic now? It seems to me that if you have two types in the same scope and want to build a non-dependent lambda then you should just be able to do it. Is it because we no longer have the `Ignore` binders and we have to use the scope to come up with a fresh name?

src/lib/SaferNames/Syntax.hs
```diff
@@ -436,147 +340,249 @@ instance Semigroup (SourceNameMap n) where
...
+  injectionProofE = undefined
+
+instance HasNamesE DataConDef where
+  traverseNamesE = undefined
+
+instance SubstE AtomSubstVal DataConDef where
+  substE = undefined
+
+instance InjectableE Atom where
+  injectionProofE = undefined
```
apaszke, line 375: Are you planning to fill those in before merging?

src/lib/SaferNames/Name.hs
```diff
@@ -316,168 +234,339 @@ data ListE (e::E) (n::S) = ListE { fromListE :: [e n] }
...
+type MonadKind2 = S -> S -> * -> *
+
+type Monad1 (m :: MonadKind1) = forall (n::S)        . Monad (m n  )
+type Monad2 (m :: MonadKind2) = forall (n::S) (l::S) . Monad (m n l)
+
+type MonadErr1 (m :: MonadKind1) = forall (n::S)        . MonadErr (m n  )
+type MonadErr2 (m :: MonadKind2) = forall (n::S) (l::S) . MonadErr (m n l)
+
+type MonadFail1 (m :: MonadKind1) = forall (n::S)        . MonadFail (m n  )
+type MonadFail2 (m :: MonadKind2) = forall (n::S) (l::S) . MonadFail (m n l)
```
apaszke, line 301: Can we unify `MonadErr` and `MonadFail`?

src/lib/SaferNames/Name.hs
```diff
@@ -316,168 +234,339 @@ data ListE (e::E) (n::S) = ListE { fromListE :: [e n] }
...
+lookupSubstM :: SubstReader v m => Name s i -> m i o (v s o)
+lookupSubstM name = (!name) <$> askSubst
+
+-- === alpha-renaming-invariant equality checking ===
+
+type AlphaEq e = AlphaEqE e  :: Constraint
+
+checkAlphaEq :: AlphaEqE e => MonadErr1 m => ScopeReader m
+             => e n -> e n -> m n ()
+checkAlphaEq e1 e2 = do
```
apaszke, line 399: We never needed to check alpha-equality before this change. Why now?

src/lib/SaferNames/Name.hs
```diff
@@ -316,168 +234,339 @@ data ListE (e::E) (n::S) = ListE { fromListE :: [e n] }
...
-  env' = injectEnvNamesL ext env
-  in RecEnv $ env' <>> frag
+withDistinct :: ScopeReader m => (Distinct n => m n a) -> m n a
+withDistinct cont = askScope >>= \case Scope _ -> cont
 
-injectEnvNamesL :: HasNamesE e => FreshExt n l -> Env i (e n) -> Env i (e l)
-injectEnvNamesL ext env = fromEnvE $ injectNamesL ext $ EnvE env
+-- === subst monad ===
+
+class ScopeReader2 m => SubstReader (v::E->E) (m::MonadKind2) | m -> v where
```
apaszke, line 331: At this point I am completely lost 😕  I need more docs/explanation

apaszke (commented):
How about we split off `Type.hs` and `PPrint.hs` from this PR and submit them afterwards? I think it should be self-contained without that and will certainly be quite a bit smaller. I tried going over `Name.hs` but as usual the 5-arrow kinds flew over my head. Can you please add some more explanations of the new monadic stuff inline?

src/lib/SaferNames/NameCore.hs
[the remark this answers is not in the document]
dougalm: Yes, it's stale. `(InjectableB, b)` became `NameSet (n:=>:l)`. @danieldjohnson the `InjectableE` condition is still necessary because not everything with kind `S -> *` can be injected. It's the usual contravariance issue with subtyping. A map from names `Name n` isn't a valid map if you extend `n` to `l`.

src/lib/SaferNames/NameCore.hs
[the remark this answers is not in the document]
dougalm: Yes, that sounds right to me. I'll add a longer comment.

src/lib/SaferNames/NameCore.hs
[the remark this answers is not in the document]
dougalm: It's really to handle two important sorts of envs: static info envs, and substitution envs. The static info envs just use the identity function on `E`, giving a payload that's exactly the type specified by the `s` parameter in `Name s n`. The substitution envs use `SubstVal`, which allow them to apply a substitution only to names with the correct `s` parameter, and a renaming otherwise.

src/lib/PPrint.hs
```diff
@@ -184,8 +184,6 @@ instance PrettyPrec e => PrettyPrec (PrimTC e) where
     _ -> prettyExprDefault $ TCExpr con
 
 instance PrettyPrec e => Pretty (PrimCon e) where pretty = prettyFromPrettyPrec
-instance Pretty (PrimCon Atom) where pretty = prettyFromPrettyPrec
```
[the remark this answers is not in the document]
dougalm: I think it was just redundant and I was trying to make it simpler

src/lib/SaferNames/Bridge.hs
```diff
@@ -49,14 +67,14 @@ instance HasSafeVersionE D.Atom where
...
     D.Var v -> S.Var $ toSafeE v
-    D.Lam (D.Abs b (arr, body)) ->
-      S.Lam $ S.Abs (toSafeB b) (S.WithArrow (fmap toSafeE arr) (toSafeE body))
-    D.Pi  (D.Abs b (arr, body)) ->
-      S.Pi  $ S.Abs (toSafeB b) (S.WithArrow (fmap toSafeE arr) (toSafeE body))
-    D.DataCon dataDef params con args ->
-      S.DataCon (toSafeE dataDef) (map toSafeE params) con (map toSafeE args)
-    D.TypeCon dataDef params ->
-      S.TypeCon (toSafeE dataDef) (map toSafeE params)
+    -- D.Lam (D.Abs b (arr, body)) ->
```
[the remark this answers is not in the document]
dougalm: Yes, this is one of the things I'm struggling with. I think I need to bite the bullet and start carrying a proper Safe<->Dangerous name map at the top level. Let's discuss.

src/lib/SaferNames/PPrint.hs
```diff
@@ -0,0 +1,203 @@
...
+      ls = LabeledItems $ case i of
+            0 -> M.empty
+            _ -> M.singleton label $ NE.fromList $ fmap (const ()) [1..i]
+    RecordTy items -> prettyExtLabeledItems items (line <> "&") ":"
+    VariantTy items -> prettyExtLabeledItems items (line <> "|") ":"
+    ACase e alts _ -> prettyPrecCase "acase" e alts
+    DataConRef _ _ _ -> undefined
+    BoxedRef ptr size (Abs b body) -> atPrec AppPrec $
+      "Box" <+> p b <+> "<-" <+> p ptr <+> "[" <> p size <> "]" <+> hardline <> "in" <+> p body
+    ProjectElt _ _ -> undefined
```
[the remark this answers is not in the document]
dougalm: Yes, happy to fill these out

src/lib/SaferNames/Name.hs
dougalm: I can't think of an alternative to traversing the expression. We definitely need to be conscious of its cost when we use it. I'll make a note.

Regarding context, I was planning to at least add the leaked names themselves to the context. Other than that, I think it's best to add context from higher up in the call stack, where we have more information about what's going on. A function like this is so general that it's not clear what more context we could add.

dougalm: Actually, adding the name to the context was an easy switch from `Maybe` to `Either Err`. Hooray for monads!

src/lib/SaferNames/Name.hs
dougalm: I use this naming convention a lot, so maybe it's worth writing a note about it. It looks something like this:
```
                             | before/above binders | after/below binders
input  (i.e. pre-traversal)  |         i            |        i'
output (i.e. post-traversal) |         o            |        o'
```
Does that picture help?

dougalm: The local vars / free vars thing is a distraction. It happens to be the case that we composition of function-and-env splits along a free/local vars boundary -- we start with a function that we want to apply to free vars, and then we rename vars as we traverse the expression and go under binders -- but that's not actually important.

src/lib/SaferNames/Syntax.hs
dougalm: Yes, it's the return type. The expr lives in the name space under the decls, so querying its type can leak names! We could use `fromConstAbs` to check for leaks, but that then requires traversing the block. This is just a cached result as an optimization (and also a sort of proof that there are no leaks). I'll write a note.

src/lib/SaferNames/Syntax.hs
dougalm: A `NameBinder TypedBinderInfo n l` doesn't actually carry a `TypedBinderInfo`. It just promises that the name it binds is a `Name TypedBinderInfo l`, which can be queried in a `Scope l` to give a `TypedBinderInfo l`. We store an actual `Type n` alongside it so that when we go under a binder, we can add the type to the scope to fulfill that promise.

src/lib/SaferNames/Syntax.hs
dougalm: Good question. I tried playing with a different flavor of name but `AtomName` ended up making the most sense. Unlike special names like `Name DataDef`, the name of an effect row extension can be bound by an ordinary lambda or pi binder.

src/lib/SaferNames/Syntax.hs
dougalm: Yes, exactly. It's the price we pay for dropping `Ignore`. I wrote about it a bit in the commit message to af82c443b5709f3677ef05efbc14a64cdb48bbea. We can always make a "scopeless" variant that queries the names in the types to create a fresh one, but that has performance implications.

src/lib/SaferNames/Syntax.hs
dougalm: Sure. Happy to.

src/lib/SaferNames/Name.hs
dougalm: I'm not sure what you mean. `MonadFail` is a standard haskell thing we need to implement if we want failable patterns in do-notation. And `MonadErr` carries our own custom error types.

src/lib/SaferNames/Name.hs
dougalm: We used to check alpha-equality as part of ordinary equality, via the `Eq` instance for `Abs`. You could do it that way with the new system too, though it's a bit more awkward because you have to conjure up a scope by querying free variables. But I decided against it for performance reasons. In the both the old system and the new one, doing alpha equality by implementing `Eq` means traversing the body many times when you have nested abstractions.

src/lib/SaferNames/Name.hs
dougalm: Sorry :/ . Let's chat and figure out which are the tricky parts that require extra docs

src/lib/SaferNames/NameCore.hs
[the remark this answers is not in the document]
danieldjohnson: Hm, ok. Should that be a third condition, then? Or is the issue that this comment is only talking about expressions, but the implementation is talking about any `E`? Maybe an example of an `E` that isn't `InjectableE` would be useful

src/lib/SaferNames/Syntax.hs
```diff
@@ -436,147 +349,392 @@ instance Semigroup (SourceNameMap n) where
...
+showParens :: Show a => a -> String
+showParens x = "(" <> show x <> ")"
+
+instance Show (DataDef n) where
+  show (DataDef name bs cons) =
+    "DataDef" <++> showParens name <++> showParens bs <++> showParens cons
+
+instance SubstE AtomSubstVal (Name DataDef) where
+  substE name =
+    lookupSubstM name >>= \case
```
danieldjohnson, line 371: Style question: I probably would have written
```
do
  Rename name' <- lookupSubstM name
  return name'
```
here. Is there a reason to prefer explicit bind with a lambda case?

src/lib/SaferNames/Name.hs
```diff
@@ -316,201 +268,364 @@ data ListE (e::E) (n::S) = ListE { fromListE :: [e n] }
...
+runScopeReaderT scope m = runReaderT (runScopeReaderT' m) scope
+
+runScopeReaderM :: Scope n -> ScopeReaderM n a -> a
+runScopeReaderM scope m = runIdentity $ runScopeReaderT scope m
+
+lookupScopeM :: Typeable s => ScopeReader m => Name s n -> m n (s n)
+lookupScopeM v = flip lookupScope v <$> askScope
+
+withDistinct :: ScopeReader m => (Distinct n => m n a) -> m n a
+withDistinct cont = askScope >>= \case Scope _ -> cont
```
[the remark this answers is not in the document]
dougalm: Yes, it should have a `ScopeReader` instance. I think it just turned out that so far I've used `SubstReaderT` instead so I didn't notice.

src/lib/SaferNames/Syntax.hs
dougalm: Good question. Definitely the `do` version is much more readable. The reason I started using lambda case when unpacking GADTs is that it seemed to work better at bringing the GADT's type classes and existential types into scope. But more recently I've noticed that `<-` sometimes does too. So maybe it's just `let`-within-`do` that doesn't? I should probably just learn what the rules are so I can make sure I only use the lambda case style when it's absolutely necessary.

src/lib/Util.hs
```diff
@@ -241,3 +242,22 @@ measureSeconds m = do
...
+  zipWithZ :: MonadFail m => (a -> b -> m c) -> f a -> f b -> m (f c)
+
+zipWithZ_ :: Zippable f => MonadFail m => (a -> b -> m c) -> f a -> f b -> m ()
+zipWithZ_ f xs ys = zipWithZ f xs ys >> return ()
+
+zipErr :: MonadFail m => m a
+zipErr = fail "zip error"
+
+forMZipped :: MonadFail m => [a] -> [b] -> (a -> b -> m c) -> m [c]
+forMZipped xs ys f
```
apaszke, line 258: If we defined `Zippable` for lists, wouldn't that give us `forMZipped` through `zipWithZ`?

src/lib/Util.hs
```diff
@@ -241,3 +242,22 @@ measureSeconds m = do
   ans <- m
   t2 <- liftIO $ getCPUTime
   return (ans, (fromIntegral $ t2 - t1) / 1e12)
+
+-- === zippable class ===
+
+class Zippable f where
+  zipWithZ :: MonadFail m => (a -> b -> m c) -> f a -> f b -> m (f c)
```
apaszke, line 249: perhaps call this `zipWithM` since it's a monadic version of `zipWith`?

src/lib/SaferNames/PPrint.hs
```diff
@@ -0,0 +1,237 @@
...
+-- https://developers.google.com/open-source/licenses/bsd
+
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE QuantifiedConstraints #-}
+{-# LANGUAGE ConstraintKinds #-}
+{-# LANGUAGE KindSignatures #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE IncoherentInstances #-}  -- due to `ConRef`
```
apaszke, line 14: Do we have that in the original PPrint too?

src/lib/SaferNames/PPrint.hs
```diff
@@ -0,0 +1,237 @@
...
+            _ -> M.singleton label $ NE.fromList $ fmap (const ()) [1..i]
+    RecordTy items -> prettyExtLabeledItems items (line <> "&") ":"
+    VariantTy items -> prettyExtLabeledItems items (line <> "|") ":"
+    ACase e alts _ -> prettyPrecCase "acase" e alts
+    DataConRef _ params args -> atPrec AppPrec $
+      "DataConRef" <+> p params <+> p args
+    BoxedRef ptr size (Abs b body) -> atPrec AppPrec $
+      "Box" <+> p b <+> "<-" <+> p ptr <+> "[" <> p size <> "]" <+> hardline <> "in" <+> p body
+    ProjectElt idxs v ->
+      atPrec AppPrec $ "ProjectElt" <+> p idxs <+> p v
```
apaszke, line 151: Those things are very hard to understand, which is why we use `prettyProjection` for this. In general I am a bit afraid that we will end up loosing a few really good features during this change by missing small things like that.

apaszke: Though I guess at least for printing we might have some golden outputs in the tests...

src/lib/SaferNames/NameCore.hs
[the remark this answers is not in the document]
apaszke: We're still missing a comment here

src/lib/SaferNames/Syntax.hs
```diff
@@ -436,147 +355,509 @@ instance Semigroup (SourceNameMap n) where
...
+  traverseNamesB (DataConRefBinding b ref) cont = do
+    ref' <- traverseNamesE ref
+    traverseNamesB b \b' -> cont $ DataConRefBinding b' ref'
+
+instance SubstB AtomSubstVal DataConRefBinding where
+  substB (DataConRefBinding b ref) cont = do
+    ref' <- substE ref
+    substB b \b' -> cont $ DataConRefBinding b' ref'
+
+instance InjectableE Atom where
```
apaszke, line 423: I'd still like to better understand the utility of those `InjectableE` instances, but it shouldn't block this

src/lib/SaferNames/Name.hs
```diff
@@ -4,288 +4,245 @@
...
+  toExtVal :: b n l -> Ext n l
+
+instance BindsNames Ext where
+  toExtVal = id
+
+toNameSet :: BindsNames b => b n l -> NameSet (n:=>:l)
+toNameSet b = fromExtVal $ toExtVal b
+
+inject :: BindsNames b => InjectableE e => Distinct l => b n l -> e n -> e l
+inject ext x = injectNames (toNameSet ext) x
```
apaszke, line 142: At some point we might want to decorate `injectNames` with `{-# INLINE #-}` to force GHC to inline the definition and make it realize that it doesn't actually have to compute the first argument anywhere.

src/lib/SaferNames/Name.hs
```diff
@@ -4,288 +4,245 @@
...
+
+tryProjectName :: MonadErr m => NameSet (n:=>:l) -> Name s l -> m (Name s n)
+tryProjectName names name =
+  case projectName names name of
+    Left name' -> return name'
+    Right _    -> throw EscapedNameErr (pprint name)
+
+-- TODO: seems silly to have to have the static annotation here. I think we need
+-- to have a finer-grained scope reader class hierarchy so we're not forced to
+-- supply static information when it's not needed.
```
apaszke, line 165: What static information? `s`?

apaszke (approved):
Looks good. Each time I read through all those lines I take something away a little more :) But I can't say that I'm fully following what's going on. The new comments are great though!

src/lib/SaferNames/Bridge.hs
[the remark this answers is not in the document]
dougalm: Yes, that's right

src/lib/Util.hs
dougalm: Yep, good point

src/lib/Util.hs
dougalm: `zipWithM` is already a thing. This is almost a generalization of that though, except that it throws an error if there's a length mismatch. So we could hide `zipWithM` and replace it with ours.

src/lib/SaferNames/PPrint.hs
dougalm: yep

src/lib/SaferNames/PPrint.hs
dougalm: Definitely something to be aware of. As you say, the golden outputs should at least catch the user-facing stuff.

src/lib/SaferNames/Name.hs
dougalm: Laziness doesn't help use there?

src/lib/SaferNames/Name.hs
dougalm: Yeah, `s`

PR #606, opened 2021-07-20
Simplify variants away

apaszke:
We'd like simplification to heavily limit the subset of our language
that has to be consumed by backends and optimization passes, and
variants are equivalent to the newly added n-ary sum types. This adds a
post-simplification pass that desugars them in the local computation.
Global atoms in bindings will still use record and variant types, to
make the compilation unit boundary well-typed.


axch:
IIRC, the purpose of the PR was recently achieved, and also this particular code is likely stale enough to be easier to rewrite if needed than resurrect.  Closing the PR on those grounds, but please reopen if my assessment is wrong.

Issue #587, opened 2021-07-01
Gradients with State effect and slice indices


srush:
In the CKY implementation in https://github.com/google-research/dex-lang/pull/586 when I try to take a gradient I get the error -> 

Can't differentiate wrt type (Ref h' ((tmp77<..) => Float32))
CallStack (from HasCallStack):
  error, called at src/lib/Autodiff.hs:402:23 in dex-0.1.0.0-1c1KJ1laoEjKvvidxV1QZ9:Autodiff

I think this is an okay thing to do. I've been able to use State and grad together before. But I think the slice indexing breaks it.

chenzizhao:
Thanks @apaszke, for reviewing this issue today! 
Normal indices give a similar error: Can't differentiate wrt type (Ref h' ((Fin 3) => Float32))

```
def myfunc2 (mat0: a=>a=>Float) (x:Float) : Float = 
    mat = yieldState mat0 \m.
        i = 0@a
        m!i!i := x
    sum for i. sum for j. mat.i.j

mat0 : (Fin 3)=>(Fin 3)=>Float = zero
myfunc2 mat0 7.0
-- (grad (myfunc2 mat0)) 7.0
```

apaszke:
@chenzizhao Thanks for making the broken program smaller! I managed to fix that in #593, but it seems like it's still not enough to make the CKY example run. I'll keep digging...

apaszke:
I think I drilled down to the core of the CKY issue and it turns out to be something that we knew is not implemented, but weren't checking for it. A fix that should at least give a reasonable error message is coming in #594.

axch:
Looks like this issue has since been fixed.  The simple repro by @chenzizhao remains functional, and I also hack-ported the CKY example to current Dex syntax, and it runs to completion there too.  (I have no idea whether the results are any good, but at least the program executes now.)

PR #611, opened 2021-07-25
Add a dependent pair type

apaszke:
Short-term this lets us implement telescopes thanks to which we can
remove the constructor hoisting hack we've been using. It should also
make it possible to do AD of triangular for-loops, which was requested
in #587, but this is not part of this patch.

Longer-term we should be able to use it to get rid of DataConRef and
implement all ADTs as coerced sum, product and dependent pair types.

### DIFF
--- a/src/lib/Autodiff.hs
+++ b/src/lib/Autodiff.hs
@@ -356,6 +356,7 @@ linearizeAtom atom = case atom of
     wrt <- ask
     return (atom, buildLam i TabArrow \i' ->
                     rematPrimal wrt $ linearizeBlock (i@>i') body)
+  DepPair _ _ _   -> notImplemented
   DataCon _ _ _ _ -> notImplemented  -- Need to synthesize or look up a tangent ADT
   Record elems    -> Record <$> traverse linearizeAtom elems
   Variant t l i e -> Variant t l i <$> linearizeAtom e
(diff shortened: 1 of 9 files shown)


duvenaud:
This might be a silly question, but why not just make all pairs dependent?  My beginner's understanding is that they generalize non-dependent pairs.

cristianurlea:
> why not just make all pairs dependent?

More often than not, adding "computational power"  (to a language) makes it more difficult to optimize. As an example: general recursion is strictly more powerful than primitive recursion, but that also makes it more difficult to reason about.

If one makes all pairs dependent, the compiler must then assume that there is a dependency between the pair's components and either evaluate both in sequence or run a potentially expensive analysis pass to prove that the components are independent. 

In contrast, a non-dependent pair type informs the compiler that there is no functional dependency between the components and this means that they can be evaluated in parallel. Even better, if a component is never used, it need not be evaluated.

duvenaud:
Thanks for the explanation.  Although I'm not sure I agree that it's hard to tell if there's a dependency.  It seems like it would just require checking whether the name of the first element appeared in the type of the second element.

However, I grant that in the compiler internals it might make sense to keep these separate after that point for the reasons you gave.  My suggestion was meant mainly for the user-facing syntax level.

apaszke:
Yeah the reason for keeping this separate is precisely what @cristianurlea said. While we don't expect to extract fine-grained parallelism from regular product types, they are just much easier to work with in the compiler, and so we separate the hard dependent cases which require much more thought. The optionality of names is also going to go away soon with Dougal's patches and we were never great about maintaining the invariant that unused type-binders are set to `Ignore`. So checking whether a pair was dependent or not was relatively expensive anyway. But we are only using this new constructor sparingly which should make it clear when the dependency is real.

PR #619, opened 2021-08-16
Port the parser and source-name renamer to SaferNames.

axch:
With apologies to Donald Knuth, beware of bugs in this code, for I have only proved it [type-]correct, not tested it.

Things that remain to be done:
- Fill in class instances GHC complains about, notably to implement SaferNames.SourceRename.Renamer
- Get the top level right in SaferNamer.SourceRename
- Port type inference to SaferNames
- Adjust Toplevel.evalUModule and friends to parse, rename, and type-check source code using both the old and new style, and check that they produce the same results on the extant corpus of Dex programs (and ideally the same errors on near misses, if there is a corpus of those available).

One particular bug this may find: there may be a `setMayShadow` call or two missing somewhere in the renamer.

### DIFF
--- a/src/lib/SaferNames/Name.hs
+++ b/src/lib/SaferNames/Name.hs
@@ -345,7 +345,11 @@ data SubstVal (sMatch::E) (atom::E) (s::E) (n::S) where
   SubstVal :: atom n   -> SubstVal s      atom s n
   Rename   :: Name s n -> SubstVal sMatch atom s n
 
-withFreshM :: (ScopeReader m, ScopeExtender m, Typeable s, InjectableE s, HasNameHint hint)
+-- TODO Presumably this wanted the constraints Typeable s and
+-- InjectableE s only because it used `withFresh`; but I don't know
+-- why `withFresh` wanted them, so I don't know whether `withFreshM`
+-- may want to add them back.
+withFreshM :: (ScopeReader m, ScopeExtender m, HasNameHint hint)
            => hint
            -> (forall o'. NameBinder s o o' -> m o' a)
            -> m o a
--- a/src/lib/SaferNames/Name.hs
+++ b/src/lib/SaferNames/Name.hs
@@ -18,9 +18,9 @@ module SaferNames.Name (
   Scope, Bindings, ScopeFrag (..), BindingsFrag, SubstE (..), SubstB (..),
   E, B, HasNamesE, HasNamesB, BindsNames (..), RecEnvFrag (..),
   BindsOneName (..), BindsNameList (..),
-  Abs (..), Nest (..), NestPair (..),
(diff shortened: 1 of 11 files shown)


src/lib/SaferNames/NameCore.hs
```diff
@@ -115,7 +115,8 @@ data NameBinder (s::E)  -- static information for the name this binds (note
                 (l::S)  -- scope under the binder (`l` for "local")
   = UnsafeMakeBinder { nameBinderName :: Name s l }
 
-withFresh :: (InjectableE s, Typeable s, Distinct n)
+-- TODO Why did this want the constraints Typable s and InjectableE s?
```
dougalm, line 118: Good catch. withFresh used to need them because the constructor for Name used to require Typeable/Injectable constraints. I changed that in 097de2023402fc2ac3bf5f7fd3e98c182da5c9b1 but I missed updating this case.

src/lib/SaferNames/SourceRename.hs
```diff
@@ -0,0 +1,368 @@
...
+-- -- === renaming patterns ===
+
+-- We want to ensure that pattern siblings don't shadow each other, so we carry
+-- the set of in-scope siblings' names along with the normal renaming env.
+
+type SiblingSet = S.Set SourceName
+
+-- TODO It seems this monad actually has no way for a downstream action
+-- to check for membership in the sibling set produced by an upstream action?
+-- Do we already have an ask/extend type of monad for this sort of thing?
```
dougalm, line 301: Yes, good observation. We should make the NameGen instance take a constraint that the underlying monad gives you read-access to the sibling set, like we do here: https://github.com/google-research/dex-lang/blob/dev/src/lib/SourceRename.hs#L313

dougalm (approved):
This looks excellent. Thank you! No need for any follow-ups on your part. I'll merge it as-is and take it from here.

src/lib/SaferNames/Name.hs
```diff
@@ -111,6 +112,10 @@ pattern EmptyAbs bs = Abs bs UnitE
 -- wraps an EnvFrag as a kind suitable for SubstB instances etc
 newtype RecEnvFrag (v::E->E) n l = RecEnvFrag { fromRecEnvFrag :: EnvFrag v n l l }
 
+-- Proof object that a given scope is void
+data IsVoidS n where
```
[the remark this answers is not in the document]
axch: I think so.

src/lib/SaferNames/NameCore.hs
axch: Perfect.  I was wondering, because Injectable looks like a pretty magic typeclass, and I didn't want to accidentally weaken the safety of the system by removing a semantic constraint.

PR #625, opened 2021-08-20
Call Dex code from Julia

oxinabox:
Based on the python interface stuff in https://github.com/google-research/dex-lang/tree/main/python

This is basically ready for review.
Though you might want to wait for the README to be written.

# TODO

- [x] write CI.yaml for this
- [x] make looking up variables from context work (idk why they don't)
- [x] scalar outputs
- [x] make function declarations work
- [x] make a string macro so can compile dex functions at julia parse time
- [x] Add flag to dex_func to make the named functions declared as const
- [x] Add support for named functions declared as `foo = \...`
- [x] Doc string for `dex_func`
- [x] README

# Maybe
-  array outputs from atoms (not possible? https://github.com/google-research/dex-lang/pull/625#issuecomment-903129851)
- [x]  work out what we are doing with values returned from Dex, as an API level. A single `Atom` type doesn't feel great, maybe we should do like PyCall does and eagerly convert things to the equiv julia type. (though what we do now is consistent with  the python interface)

# Probably follow up PRs
-  Optimize `NativeFunction` to not to do any unnesc allocations during call, but to do them all during Construction, so that `dex_func` can have basically 0 allocations at run-time (except for array flipping and maybe the named_size dict)
-  make working with arrays not generate copies, by using julia's C ordered array types
-  make said string macro also generate `ChainRulesCore.rrule`s that hood into Dex's AD, so AD  Just Works (tm)
-  Register as a julia package in General registry (can't be done by me)


oxinabox:
I am a bit stuck on how to get array outputs,
because the following python doesn't work,

```python
>>> import dex
>>>
... m = dex.Module("""
... x = 2.5
... y = [2, 3, 4]
... """)
>>>
>>> result = dex.api.CAtom()
>>> dex.api.toCAtom(m.y, result)
0

>>> result.value  # expected CRectArray here
<dex.api.Lit object at 0x7f378548c440>

>>> result.value.value
0
```

I thought it would work because this tagged union in python exist:
https://github.com/google-research/dex-lang/blob/01cbb9f9fb3eeaa4656a3f192a143efdf7bd69e2/python/dex/api.py#L24-L29

OTOH this code exist which I think make's it clear it shouldn't
https://github.com/google-research/dex-lang/blob/01cbb9f9fb3eeaa4656a3f192a143efdf7bd69e2/src/Dex/Foreign/Serialize.hs#L76
So I guess the code  api.py is dead.
But I can't find the real code that works for this.
I think it might be in `native_function.py`

julia/README.md
```diff
@@ -0,0 +1,121 @@
...
+julia> juliaize(m.x)
+42
+
+julia> typeof(juliaize(m.x))
+Int32
+```
+
+It is not presently possible to `juliaize` arrays (but you can get them as the output of functions, see below).
+
+To convert functions into something you can execute, use `NativeFunction`.
```
oxinabox, line 74:
```suggestion
To convert functions `Atoms` into something you can execute, use `NativeFunction`.
```

julia/README.md
```diff
@@ -0,0 +1,121 @@
...
+"126"
+
+julia> m.addTwo
+"\\n:Int32 ?->\n  \\x:((Fin n) => Float32).\n    for i:(Fin n).\n       tmp:((Add Float32) ?=> Float32 -> Float32 -> Float32) = (+) Float32\n       tmp1:(Float32 -> Float32 -> Float32) = tmp instance1\n       tmp2:Float32 = x i\n       tmp3:(Float32 -> Float32) = tmp1 tmp2\n      tmp3 2."
+```
+
+
+## Atoms: `juliaize`, `NativeFunction`
+
+As shown above `evaluate` and the contents of a `DexModule` are returned as atoms.
```
oxinabox, line 52:
```suggestion
`evaluate` and the contents of a `DexModule` are returned as `Atom`s.
```

julia/README.md
```diff
@@ -0,0 +1,121 @@
...
+
+```julia
+julia> typeof(m.x)
+DexCall.Atom
+
+julia> typeof(m.addTwo)
+DexCall.Atom
+```
+
+To convert scalars into julia typed scalars used `juliaize`
```
oxinabox, line 63:
```suggestion
To convert scalars into julia typed scalars used `juliaize`.
```

julia/src/api_types.jl
```diff
@@ -0,0 +1,67 @@
...
+
+
+struct CLit <: TaggedUnion{Tuple{Int64, Int32, Int8, Float64, Float32}}
+    tag::UInt64
+    payload::NTuple{8, UInt8}  # actually the Union, needs to be big enough to hold largest which is an Int64
+end
+
+mutable struct CAtom <: TaggedUnion{Tuple{CLit, CRectArray}}
+    tag::UInt64
+    payload::NTuple{3, UInt64}  # actually the union needs to be big genough to hold largest which is the CRectArray
```
oxinabox, line 50:
```suggestion
    payload::NTuple{3, UInt64}  # actually the Union, needs to be big genough to hold largest which is the CRectArray
```

julia/src/api_types.jl
```diff
@@ -0,0 +1,67 @@
...
+    CAtom() = new()  # incomplete initialization because will initialize in a ccall
+end
+
+function CAtom(atm::Ptr{HsAtom})
+    result = Ref(CAtom())
+    success = to_CAtom(atm, result)
+    iszero(success) && throw_from_dex()
+    return result[]
+end
```
oxinabox: I need to test this locally, but I think this can be done with:
```suggestion
struct CAtom <: TaggedUnion{Tuple{CLit, CRectArray}}
    tag::UInt64
    payload::NTuple{3, UInt64}  # actually the Union, needs to be big enough to hold largest which is the CRectArray
end

function CAtom(atm::Ptr{HsAtom})
    result = Ref{CAtom}()
    success = to_CAtom(atm, result)
    iszero(success) && throw_from_dex()
    return result[]
end
```

julia/src/api_types.jl
```diff
@@ -0,0 +1,67 @@
...
+
+"""
+    juliaize(x)
+
+Get the corresponding Julia type from some output of Dex.
```
oxinabox, line 64:
```suggestion
Get the corresponding Julia object from some output of Dex.
```

julia/src/native_function.jl
```diff
@@ -0,0 +1,226 @@
...
+    finally
+        free_function_signature(sig_ptr)
+    end
+    #== TODO
+    func_type = ctypes.CFUNCTYPE(
+        ctypes.c_int64,
+        *(arg.type.arg_ctype for arg in self.argument_signature),
+        *(res.type.ref_ctype for res in self.result_signature))
+    self.callable = func_type(ctypes.cast(ptr, ctypes.c_void_p).value)
+    ==#
```
oxinabox, line 76:
```suggestion
```

julia/src/native_function.jl
```diff
@@ -0,0 +1,219 @@
...
+end
+function NativeFunctionSignatureJL(x::NativeFunctionSignature)
+    return NativeFunctionSignatureJL(unsafe_string.((x.arg, x.res, x._ccall))...)
+end
+function NativeFunctionSignatureJL(ptr::Ptr{NativeFunctionSignature})
+    return NativeFunctionSignatureJL(unsafe_load(ptr))
+end
+
+"given a Binder for a result_signature returns the julia type that will be returned by the funcion"
+result_type(x::Binder) = result_type(x.type)  # don't care about size etc
```
oxinabox, line 178:
```suggestion
result_type(x::Binder) = result_type(x.type)
```

julia/src/native_function.jl
```diff
@@ -0,0 +1,219 @@
...
+    end
+
+    type = Either{Any}(Any[array_type, scalar_type])
+
+    implicit = parser("?"=>true) | false
+
+    arg_sig = map(implicit * name * ":" * type) do (i, n, _ , t)
+        Binder(n, t, i)
+    end
+    arg_sigs = join(Repeat(arg_sig),",")
```
oxinabox, line 216:
```suggestion
    size_ele = Numeric(Int) | name
    sizes = join(Repeat(size_ele),",")
    array_type = map(scalar_type * "[" * sizes * "]") do (T, _, sz, _)
        ArrayBuilder{T}(sz)
    end
    type = Either{Any}(Any[array_type, scalar_type])
    implicit = parser("?"=>true) | false
    arg_sig = map(implicit * name * ":" * type) do (i, n, _ , t)
        Binder(n, t, i)
    end
    arg_sigs = join(Repeat(arg_sig),",")
```

julia/src/native_function.jl
```diff
@@ -0,0 +1,213 @@
...
+
+# CombinedParsers.jl based parser:
+@with_names begin
+    name = map(Symbol, !re"arg\d+")
+    scalar_type = Either{Any}(Any[
+        parser("f32"=>Float32),
+        parser("f64"=>Float64),
+        parser("i32"=>Int32),
+        parser("i64"=>Int64),
+        parser("i8"=> Int8),
```
oxinabox, line 198:
```suggestion
        parser("i8"=>Int8),
```

apaszke:
Yeah the serialization of arrays into `CAtom` is not implemented yet, sorry. At the moment the only efficient (pointer-based) exchange for arrays is implemented in the native function interface.

julia/src/native_function.jl
```diff
@@ -0,0 +1,277 @@
...
+    end
+
+    results_ptrs_and_thunks = map(f.result_signature) do builder
+        create(builder.type, named_sizes)  
+    end
+
+    result_ptrs = first.(results_ptrs_and_thunks)
+    result_thunks = Tuple(last.(results_ptrs_and_thunks))
+
+    args = (implict_values..., arg_ctypes..., result_ptrs...)
```
apaszke, line 160: The assumption that all the implicit values come before arg_ctypes is a little sketchy and doesn't necessary have to be true. Can you follow the behavior of Python bindings and just use the full argument signature to constructor the argument tuple?

apaszke (commented):
Wow this is super cool! Thanks! I hardly know any Julia, so can't do a super thorough review, but this looks very nice. My main concern is the construction of the argument vector in native function that seems too restrictive.

Also, doesn't Julia use LLVM for compilation too? I wonder if it would be possible to lower Dex functions to LLVM modules and link those together with the larger Julia programs. That would give you essentially no FFI boundary but would all get compiled into a single binary blob!

apaszke:
And about the mixup around C/Fortran array orders, we could add a flag to the JIT that would make it generate code that expects Fortran ordered arrays for I/O.

julia/README.md
```diff
@@ -0,0 +1,136 @@
...
+
+You can use it to define either named functions.
+Both in long form:
+```julia
+julia> dex_func"""
+              def myTranspose (n: Int) ?-> (m: Int) ?->
+                              (x : (Fin n)=>(Fin m)=>Float) : (Fin m)=>(Fin n)=>Float =
+                  for i j. x.j.i
+              """
+(::NativeFunction{Matrix{Float32}}) (generic function with 1 method)
```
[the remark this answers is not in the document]
oxinabox: Yeah, we need to remember the return type to make it type stable.
Possibly i could instead stick the pointer into the type parameter, and then let specialization and type inference work it out, but that is less useful to the user.
Possibly in future will stick the input types as type-parameters also.
(or the whole `Binder` objects)

julia/src/api.jl
```diff
@@ -0,0 +1,65 @@
+const libdex = joinpath(dirname(@__DIR__), "deps", "libDex.so")
+isfile(libdex) || error("libDex not found in $libdex. Please run `Pkg.build()`")
+
+
+##########################################################################################
+# Global State
+
+# These can only be called once in life-time of program, can not re-init after fini etc
```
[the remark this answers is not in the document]
oxinabox: Errors for me when I was doing it.
But I might have been doing something else wrong.

I will amend that comment

julia/src/evaluate.jl
```diff
@@ -0,0 +1,79 @@
+"""
+    Atom
+A wrapped DexLang value.
+On its own not good for much -- just displays as a string.
```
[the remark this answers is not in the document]
oxinabox: Right now I don't have that exposed anywhere.
Possibly should, but would need thinking and working out what different parts of API are poked.
Where is example of that in the python code?

julia/src/api.jl
```diff
@@ -0,0 +1,65 @@
...
+lookup(ctx, str) = @ccall libdex.dexLookup(ctx::Ptr{HsContext}, str::Cstring)::Ptr{HsAtom}
+
+print(atm) = unsafe_string(@ccall libdex.dexPrint(atm::Ptr{HsAtom})::Cstring)
+
+compile(ctx, atm, jit=JIT) = @ccall libdex.dexCompile(jit::Ptr{HsJIT}, ctx::Ptr{HsContext}, atm::Ptr{HsAtom})::Ptr{NativeFunctionObj}
+unload(f, jit=JIT) = NO_FREE[] || @ccall libdex.dexUnload(jit::Ptr{HsJIT}, f::Ptr{NativeFunctionObj})::Nothing
+
+get_function_signature(f, jit=JIT) = @ccall libdex.dexGetFunctionSignature(jit::Ptr{HsJIT}, f::Ptr{NativeFunctionObj})::Ptr{NativeFunctionSignature}
+
+# This disagrees with the code in python/src/api.py, but matches the code in src/Dex/Foreign
```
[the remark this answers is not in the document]
oxinabox:
```suggestion
```

julia/src/evaluate.jl
```diff
@@ -0,0 +1,79 @@
...
+
+However scalar values can be converted to julia objects using [`juliaize`](@ref),
+and functions can be made exectuable using [`NativeFunction`](@ref).
```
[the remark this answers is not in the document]
oxinabox: yep, i have not implemented that, yet.
I feel like it is something to think about later as part of:

> work out what we are doing with values returned from Dex, as an API level. A single Atom type doesn't feel great, maybe we should do like PyCall does and eagerly convert things to the equiv julia type.

which similarly might also implement overloads of `convert` and `promote`

julia/src/native_function.jl
```diff
@@ -0,0 +1,277 @@
...
+            assigment = Expr(:const, assigment)
+        end
+        return assigment
+    end
+end
+
+
+"holds information on how to map an argument argment in a signature into a julia type"
+struct Binder
+    name::Symbol
```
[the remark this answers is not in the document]
oxinabox: More or less.
It is the type that represents as identifier in the julia AST.
It is interned; and is freely convertable to and from `String`.
But it itself doesn't implement string operations like `startwith` etc.

julia/src/native_function.jl
```diff
@@ -0,0 +1,277 @@
...
+    type  # Intentionally left unrestricted for now. No restriction seems useful, it is either one of half a dozen scalar types or an ArrayBuilder
+    implicit::Bool
+end
+
+"Creates an `Array{T,N}``, if all elements of size are present then that array is initialized to hold `size`"
+struct ArrayBuilder{T,N}
```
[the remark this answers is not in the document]
oxinabox: Yeah, it specializes on them and allows dispatch on them.
You can see when dispatch on them happens further below, where it has separate dispatches to make it error if you are trying to use with array of the wrong element type

julia/src/native_function.jl
```diff
@@ -0,0 +1,277 @@
...
+end
+ArrayBuilder{T}(size) where T = ArrayBuilder{T,length(size)}(size)
+
+
+
+"""
+    NativeFunction{R}
+
+Callable Julia wrapper of some native Dex function.
+Returns type from calling it is is `R`.
```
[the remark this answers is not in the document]
oxinabox:
```suggestion
Return type from calling it is is `R`.
```

julia/src/native_function.jl
```diff
@@ -0,0 +1,277 @@
...
+
+    named_sizes = Dict{Symbol, Int}()
+    arg_ctypes = map(f.explicit_argument_signature, args) do binder, arg
+        to_ctype!(named_sizes, binder.type, arg)
+    end
+
+    implict_values = map(f.implict_argument_signature) do binder
+        # Right now the Exported API only supports implicts that are sizes
+        # and now all of them will have been determined by the preverious `to_ctype!` step
+        @assert binder.type == Int32
```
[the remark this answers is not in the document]
oxinabox: This is not important.
It is just an assertion about my belief about the world.
But AFAICT is is always true for things that an be called in the API?
And the other code is invalid if it isn't true?
If it isn't true I would like a example that  shows it is not to add to the tests.

julia/src/native_function.jl
oxinabox: I could.
Can you show me a case that doesn't follow this so I can add it to tests?
It was true for all the cases in the python tests.
And making that assumption simplified the code.
But I could totally do it the way python does.

julia/src/native_function.jl
```diff
@@ -0,0 +1,277 @@
...
+    _ccall_dex_func(f.c_func_ptr, args...)
+
+    if length(result_thunks) === 1
+        only(result_thunks)()
+    else
+        map(t->t(), result_thunks)
+    end
+end
+
+# Work around needing to have the return types as literals, by defining a generated function that does.
```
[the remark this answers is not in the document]
oxinabox: Noone should be expected to follow what this code does.
Like even people who have been using julia for years shouldn't.
It's pretty gross.

The sensible version of this code is:
````julia
ccall(f_ptr, Int64, typeof.(args), args...)
```
What this code does is call a function that has a C API, with return type Int64, and with argument types that are exactly the types of the `args`
The reason it does it this way is because `ccall` is not a function but a built-in language intrinsic that is largely resolved at compile time, and that needs the type information given to it as literals.
`typeof.(args)` is not a literal which is a problem -- even though it can actually be resolved at compile-time -- the information exists.
So we use a generated function in order to get at the information.
Docs on those are at https://docs.julialang.org/en/v1/manual/metaprogramming/#Generated-functions
The short version though is that they take in arguments at compile time and they only get their types, then they return a AST which is what is run at runtime.
So what we are doing is building the AST that is for the ccall, which has the types all present as literals.


I will add a comment to the code to explain all this.
And an issue on JuliaLang/julia about how much this sucks.

oxinabox:
> And about the mixup around C/Fortran array orders, we could add a flag to the JIT that would make it generate code that expects Fortran ordered arrays for I/O.

That would be great. Much faster than copies.
And cleaner than pushing code to use non-default julia array types

oxinabox:
> Also, doesn't Julia use LLVM for compilation too? I wonder if it would be possible to lower Dex functions to LLVM modules and link those together with the larger Julia programs. That would give you essentially no FFI boundary but would all get compiled into a single binary blob!

Yeah, I think that should be possible with `llvmcall` or LLVM.jl
I think beyond scope of this PR but I will talk to some people and find-out how feasible it is.

julia/src/native_function.jl
```diff
@@ -0,0 +1,277 @@
...
+3-element Vector{Float32}:
+ 2.0
+ 4.0
+ 8.0
+```
+"""
+macro dex_func_str(str, flags="")
+    m = something(
+        match(r"^def ([a-zA-Z0-9_]+)", str),  # `def foo`
+        match(r"^([a-zA-Z0-9_]+)\s*=\s*[\\]", str),  # `foo = \x`
```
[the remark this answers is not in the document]
oxinabox: I have on good authority that the correct regex for julia identifies is
https://gist.github.com/pfitzseb/1e5cb6c52b196bb556983959c1aef02b
we **could** accept that

julia/src/api.jl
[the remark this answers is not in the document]
apaszke: Oh ok, in that case let's leave it as is! Probably no one wants to do that anyway

julia/src/native_function.jl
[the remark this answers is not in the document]
apaszke: At the moment it might be true, but I expect that there will come a time where it won't (e.g. Int64 might be another valid option). I'd rather not hard-code any assumptions that are unnecessary and we don't make in Dex itself.

julia/src/native_function.jl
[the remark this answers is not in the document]
apaszke: Thank you, that's super helpful.

julia/src/native_function.jl
apaszke: I expect that if you did this then it might fail:
```
def f (a : Float) (n : Int) ?-> (b : (Fin n)=>Float) : Float = a + sum b
```

apaszke:
Yeah the LLVM stuff is definitely beyond the scope, but it would be super neat. It would effectively eliminate any language boundary and allow LLVM to cross-optimize the call boundary.

julia/src/native_function.jl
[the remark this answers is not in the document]
oxinabox:
```suggestion
```

julia/src/evaluate.jl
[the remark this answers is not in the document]
apaszke: [Right here](https://github.com/google-research/dex-lang/blob/main/python/tests/api_test.py#L30-L37).

julia/src/native_function.jl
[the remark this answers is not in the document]
oxinabox: Resolved by relaxing it and then letting the parser decide for itself if it is an error

.github/workflows/julia-ci.yaml
```diff
@@ -0,0 +1,57 @@
...
+
+    - name: Setup Haskell Stack
+      uses: actions/setup-haskell@v1
+      with:
+        enable-stack: true
+        stack-no-global: true
+        stack-version: 'latest'
+    
+    - name: Setup Julia
+      uses: julia-actions/setup-julia@2d6959702913c06ee8c6997dadafe089ec18549f  # v1.6.1
```
oxinabox, line 32:
```suggestion
      uses: julia-actions/setup-julia@ee66464cb7897ffcc5322800f4b18d449794af30  # v1.6.1
```

julia/src/native_function.jl
oxinabox: You are correct that that doesn't have the implict first
```
julia> DexCall.NativeFunctionSignatureJL(sig_ptr)
DexCall.NativeFunctionSignatureJL("arg0:f32,?arg1:i32,arg2:f32[arg1]", "arg3:f32", "arg0,arg1,arg2,arg3")
```

But for some reason the code works, and I can't seem to workout why
I can't be just getting lucky with pointers can I?
The following is with a show statement so I can see what is being passed to the `_dex_ccall`
```julia
julia> dex_func"def f (a : Float) (n : Int) ?-> (b : (Fin n)=>Float) : Float = a + sum b"
       f(100f0, [10f0, 2f0, 0.1f0, 0.1f0])
args = (implict_values..., arg_ctypes..., result_ptrs...) = (4, 100.0f0, Ptr{Float32} @0x00007fda2c47b590, Ptr{Nothing} @0x00007fda298e2300)
112.2f0
```

oxinabox: I have fixed it anyway. Because I have no idea why that worked.
Now it only makes the assumption that return types are at the end, and that arguments go into the ccall the same order as in the the argument_signature.
Which is an assumption made in the python code here also:
https://github.com/google-research/dex-lang/blob/eb52530e3666ed1d09ac913a8edfce290548264b/python/dex/native_function.py#L107-L110

apaszke: Yes this is the right assumption to make. Our API is undocumented (of course), but those are the semantics of native function export.

julia/src/native_function.jl
```diff
@@ -136,31 +133,35 @@ function NativeFunction(atom::Ptr{HsAtom}, ctx=PRELUDE, jit=JIT)
...
-        throw(ArgumentError("wrong number of inputs, expected: $(length(f.explicit_argument_signature) )"))
-    end
-
     named_sizes = Dict{Symbol, Int}()
-    arg_ctypes = map(f.explicit_argument_signature, args) do binder, arg
-        to_ctype!(named_sizes, binder.type, arg)
+    num_explict = 0
+    c_args = map(f.argument_signature) do binder
+        if binder.implicit
+            -one(binder.type)  # we will do implicts in a second pass, once named_sizes is filled out
```
apaszke, line 140: Note that it is possible to handle the arguments in a single loop, but you'd have to iterate over them in reverse. The scoping rules for arguments are left-to-right, so if an implicit can be inferred from an argument, it has to come after it. So in the reverse loop it would get found first and would fill in the right fields in `named_sizes`.

apaszke (approved):
Looks great! Let me know if you want to make the change below. It's totally optional, but it might simplify the code a little.

julia/src/native_function.jl
oxinabox: Ok cool!
lets add a note and then it can be changed later.
I will save that for a make that in a follow up PR, when I optimize this heavily.

```suggestion
            -one(binder.type)  # we will do implicts in a second pass, once named_sizes is filled out
            # Note: because scoping rules are left-to-right if we reversed the list we could do this in a single pass
```

PR #628, opened 2021-08-25
Safe names type checker

dougalm:


src/lib/SaferNames/Type.hs
```diff
@@ -0,0 +1,968 @@
...
+-- MonadAtomSubst.
+class ( MonadFail2 m, Monad2 m, MonadErr2 m, EnvGetter Name m
+      , BindingsReader2 m, BindingsExtender2 m)
+     => Typer (m::MonadKind2) where
+  declareEffs :: EffectRow o -> m i o ()
+  extendAllowedEffect :: Effect o -> m i o () -> m i o ()
+  withAllowedEff :: EffectRow o -> m i o a -> m i o a
+
+-- This fakes MonadErr by just throwing a hard error using `error`. We use it
+-- to skip the checks (via laziness) when we just querying types.
```
apaszke, line 75: Interesting! Have you verified that it actually skips the checks?

src/lib/SaferNames/Type.hs
```diff
@@ -0,0 +1,968 @@
...
+  BaseTypeRef p -> do
+    (PtrTy (_, b)) <- getTypeE p
+    return $ RawRefTy $ BaseTy b
+  TabRef tabTy -> do
+    TabTy b (RawRefTy a) <- getTypeE tabTy
+    return $ RawRefTy $ TabTy b a
+  ConRef conRef -> case conRef of
+    ProdCon xs -> RawRefTy <$> (ProdTy <$> mapM typeCheckRef xs)
+    IntRangeVal     l h i -> do
+      l' <- substM l
```
apaszke, line 382: Can you please remind me why does type checking now need a substitution? I briefly recall us discussion that but I already forgot 😕

apaszke (approved):
Far from a comprehensive review, but LGTM after a skim

src/lib/Err.hs
```diff
@@ -140,3 +144,5 @@ instance Pretty ErrType where
     ZipErr            -> "Zipping error"
     EscapedNameErr    -> "Escaped name"
     ModuleImportErr   -> "Module import error"
+
+sCompilerErr (Right x) = Right x
```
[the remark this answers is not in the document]
dougalm: yep, thanks

src/lib/SaferNames/Type.hs
dougalm: No, I should try using Debug.Trace to make sure

src/lib/SaferNames/Type.hs
dougalm: Unless you have the `Distinct n` constraint, shadowing is possible. So you can end up with expressions like this:
```
\(x:a). \(a:Type). x
```
The type of `x` should be the outer `a`, not the inner one.

Previously, we sort-of-mostly tried to enforce the no-shadow invariant everywhere, so you had to deshadow when looking up a substitution, for example. We also previously had types cached alongside variable occurrences, so the above example would have been impossible.

PR #632, opened 2021-08-25
Demo: Gaussian process regression

chenzizhao:
Pending #617 

This is a GP regression demo inspired by https://github.com/google/jax/blob/main/examples/gaussian_process_regression.py and #541 The main difference from #541 is to use a PSD solver from lib/linalg instead of a gradient solver. 

Q: add to examples/kernalregression instead of creating a new file?

Screenshot from 2021-08-25 12-50-51


apaszke (approved):

apaszke:
Thank you @chenzizhao!

Issue #638, opened 2021-08-26
ParseFailureException when running dex for the first time


AlexDikelsky:
Hello, I am starting to use dex. When I tried to run

    $ stack exec dex -- --lib-path lib repl

I got the error
```
Compiler bug!
Please report this at github.com/google-research/dex-lang/issues

ParseFailureException "Unknown attribute kind (70) (Producer: 'LLVM12.0.0' Reader: 'LLVM 9.0.1')"
```

I was doing this on a brand new virtual machine, so here are the relevant commands I ran:
```
wget -wO- https://get.haskellstack.org/ | sh
sudo apt-get install llvm-9-dev
sudo apt-get install clang
sudo apt-get install libpng-dev
curl --proto '=https' --tlsv1.2 -sSf https://get-ghcup.haskell.org | sh
git clone https://github.com/google-research/dex-lang
cd dex-lang
make install
stack exec dex -- --lib-path lib repl
```

I am on Ubuntu
```
$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description: Ubundu 21.04
Release: 21.04
Codename: hirsute
```

on commit `b995589...`

I think I installed `stack` twice in the installation at different steps, but I don't know if that would cause this problem.

I saw #585, and my clang version is
```
Ubuntu clang version 12.0.0-3ubuntu1~21.04.1
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```

AlexDikelsky:
Should be fixed by #639

Issue #585, opened 2021-06-27
Unknown attribute kind (70)


oxinabox:
New laptop (ubuntu this time), new failures.
Freshly built clone from current `main` errors as soon as I start the REPL.
```
$ stack exec -- dex repl                                                                17:56:47
Compiler bug!
Please report this at github.com/google-research/dex-lang/issues

ParseFailureException "Unknown attribute kind (70) (Producer: 'LLVM12.0.0' Reader: 'LLVM 9.0.1')"
```
I guess it is mixing LLVM versions somewhere.

oxinabox:
Fixed by using Clang v11 rather than Clang v12

apaszke:
We should do some stricter version checking in the makefile 😕

axch:
Looks like this is fixed-enough by #639, so I'm going to close it.

PR #639, opened 2021-08-27
Check that the clang compiler is compatible with LLVM-9 in the makefile

AlexDikelsky:
This should provide an error message when there is an incompatibility between clang and llvm versions. See #638 and #585 for more discussion. 

`-dumpversion` isn't documented in `clang++ --help`, but is listed in the [reference](https://clang.llvm.org/docs/ClangCommandLineReference.html).

### DIFF
--- a/makefile
+++ b/makefile
@@ -49,7 +49,7 @@ endif
 
 possible-clang-locations := clang++-9 clang++-10 clang++-11 clang++
 
-CLANG := $(shell if [[ $$DEX_LLVM_HEAD -eq 1 ]] ; \
+CLANG := $(shell if [[ "$$DEX_LLVM_HEAD" -eq 1 ]] ; \
 	then echo "clang++"; \
 	else \
 		for clangversion in $(possible-clang-locations) ; do \
--- a/makefile
+++ b/makefile
@@ -47,26 +47,29 @@ ifneq (,$(PREFIX))
 STACK_BIN_PATH := --local-bin-path $(PREFIX)
 endif
 
+possible-clang-locations := clang++-9 clang++-10 clang++-11 clang++
+
+CLANG := $(shell for clangversion in $(possible-clang-locations) ; do \
+	if [[ $$(command -v "$$clangversion" 2>/dev/null) ]]; \
+	then echo "$$clangversion" ; break ; fi ; done)
+
+
 ifeq (1,$(DEX_LLVM_HEAD))
 ifeq ($(PLATFORM),Darwin)
-	$(error LLVM head builds not supported on macOS!)
+$(error LLVM head builds not supported on macOS!)
 endif
 STACK_FLAGS := $(STACK_FLAGS) --flag dex:llvm-head
 STACK := $(STACK) --stack-yaml=stack-llvm-head.yaml
+else \
+ifeq (,$(CLANG))
+$(error "Please install clang++-9")
+else
+clang-version-compatible := $(shell $(CLANG) -dumpversion | awk '{ print(gsub(/^((9\.)|(10\.)|(11\.)).*$$/, "")) }')
+ifneq ($(clang-version-compatible),1)
+$(error "Please install clang++-9")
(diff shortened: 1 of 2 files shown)


AlexDikelsky:
I changed the makefile so that the clang executable doesn't have to be in `clang++` (mentioned in #638). Should clang 9 be the recommended version? Any of 9, 10, and 11 seem to work.

I changed the readme too because if you use `apt-get install clang` in Ubuntu 21.04, you get clang 12.

makefile
```diff
@@ -55,6 +55,19 @@ STACK_FLAGS := $(STACK_FLAGS) --flag dex:llvm-head
...
+	then echo "$$clangversion" ; break ; fi ; done)
+possible-clang-locations = clang++-9 clang++-10 clang++-11 clang++
+clang-version-compatibile = $(shell $(CLANG) -dumpversion | awk '{ print(gsub(/^((9\.)|(10\.)|(11\.)).*$$/, "")) }')
+check-clang-version:
+ifeq (, $(CLANG))
+	$(error "Please install clang++-9")
+endif
+ifneq ($(clang-version-compatibile), 1)
+	$(error "Please install clang++-9")
+endif
```
apaszke, line 69: Are those ifs really part of the `check-clang-version` rule? I thought that the recipe of a single target always has to be indented by tabs. If they're not then we might be able to remove that target.

apaszke: Oh and also please make sure that we _don't_ check this when the env variable [`DEX_LLVM_HEAD`](https://github.com/google-research/dex-lang/blob/main/makefile#L50-L56) is set.

apaszke (commented):
Very nice, thank you!

makefile
AlexDikelsky: I believe that the ifs are part of the rule (There's some conversation on this [here](https://unix.stackexchange.com/questions/220447/checking-environment-variables-value-in-makefile), and I also looked at it [here](https://gist.github.com/AlexDikelsky/f5e498c4a5257f4ab12838dfa7a9dd61)).

The reason I have the rule in the first place is so that you can still run `make clean` without having the right clang version, but I'm not sure if that is important. In [3589d7f](https://github.com/google-research/dex-lang/pull/639/commits/3589d7fc4648011990c48b50cbb537e1d2bbd5ec) I removed the rule, but I think adding it back in would be fairly simple.

apaszke (approved):
Very nice! Thanks for adding the checks. Hopefully it will save others some time in the future 😕

makefile
apaszke: Today I learned! Thanks for investigating this.

apaszke: Adding the rule sounds good to me and might be nicer. This looks ready to merge, but I can wait for you to add that if you'd prefer. Just let me know.

AlexDikelsky: Thanks!

Before this is merged, I want to change the `CLANG :=` to respect `DEX_LLVM_HEAD`, since I think it will try to use clang++-9 even if that is set.

makefile
```diff
@@ -47,13 +47,34 @@ ifneq (,$(PREFIX))
...
+
+CLANG := $(shell if [[ "$$DEX_LLVM_HEAD" -eq 1 ]] ; \
+	then echo "clang++"; \
+	else \
+		for clangversion in $(possible-clang-locations) ; do \
+		if [[ $$(command -v "$$clangversion" 2>/dev/null) ]] ; \
+		then echo "$$clangversion" ; break ; \
+		fi ; \
+		done ; \
+	fi)
```
apaszke, line 60: Instead of putting the conditional on DEX_LLVM_HEAD couldn't we set `CLANG` to be `clang++` by default and override its value using the `for` loop you have here in the `else` case of the `ifeq (1,$(DEX_LLVM_HEAD))` conditional?

apaszke (approved):

apaszke:
Thank you!

Issue #637, opened 2021-08-25
Scoping bug in inference for cases with dependent data constructors


dougalm:
Minimal repro:
```
def listElt (x:List Int) : Int =
  case x of AsList _ y -> y.(0@_)

> assertion failure ():
> [n:Int32, elements:((Fin n) => Int32)] != [pat:Int32, y:((Fin n) => Int32)]
```

The problem is that we flatten the nest of binders [here](https://github.com/google-research/dex-lang/blob/bcc8b1b258c6b8c50bb5e2b44bc2cee2800911c1/src/lib/Inference.hs#L587), ignoring the possibility of dependence between them. I don't know how this *ever* worked, really. It's coming up now because #629 exposes it in vega-plotting.dx. It's the sort of thing that'll be ruled out by the safe-names system anyway, and inference is the next pass I'm going to try to tackle, so it should be fixed soon.

sharadmv:
I'm not sure if this is exactly the same bug but I ran into a similar compiler error with this:
```
import linalg

data Distribution =
    Gaussian n:Type (n => Float) (LowerTriMat n Float)

cov =  [[4., 2.], [2., 4.]]

def ones : (n => Float) =
  for i. 1.

def gaussianLogProb (mean: n => Float) (covChol: LowerTriMat n Float) (x: n => Float): Float =
  1.


def logProb (dist: Distribution) (x:n => Float) : Float =
  case dist of
    Gaussian n mean covChol -> gaussianLogProb mean covChol x
```

Error:
```
Compiler bug!
Please report this at github.com/google-research/dex-lang/issues

assertion failure ():
[n:Type, tmp11:(n => Float32), tmp12:((tmp12:n) => (..tmp12) => Float32)] != [n1:Type, mean:(n => Float32), covChol:((tmp:n) => (..tmp) => Float32)]

CallStack (from HasCallStack):
  assertEq, called at src/lib/Type.hs:365:39 in dex-0.1.0.0-9xSbpA6uKrYCn2f8QfR66i:Type

checking decl:
logProb:((n:Type) ?-> Distribution -> (n => Float32) -> Float32) =
  \n:Type ?->
    \dist:Distribution x:(n => Float32).
      case dist of
        n1 mean covChol ->
          logProb1:((n => Float32)
                    -> ((tmp:n) => (..tmp) => Float32)
                    -> (n => Float32) -> Float32) = gaussianLogProb n
          logProb2:(((tmp:n) => (..tmp) => Float32)
[...]

PR #640, opened 2021-08-29
Fix MNIST loading in tutorial

mkiefel:
Each pixel value in the Fashion MNIST dataset is stored as unsigned
byte.  Previously the loading code of the tutorial would map values
equal and above 128 (which are interpreted as negative numbers by the
signed datatypes) to the wrong integer value (e.g., '\128' -> 256,
should be 128, '\129' -> 255, should be 129).

Testing: Compared the rendering of the first shoe image displayed in the
browser with a Python reference implementation.

### DIFF
--- a/examples/tutorial.dx
+++ b/examples/tutorial.dx
@@ -317,7 +317,7 @@ Full = Fin ((size Batch) * (size IHeight) * (size IWidth))
 def pixel (x:Char) : Float32 =
      r = W8ToI x
      IToF case r < 0 of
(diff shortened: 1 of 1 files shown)


srush (approved):
Oh yes, @duvenaud pointed this out of well. Meant to send a PR.

apaszke (approved):
Nice catch, thanks!

apaszke:
@dougalm Is it expected that #637 might show up here?

PR #641, opened 2021-08-30
Safe-names version of Builder

dougalm:


src/lib/SaferNames/Name.hs
```diff
@@ -91,7 +94,7 @@ class Monad1 m => ScopeReader (m::MonadKind1) where
   getScope :: m n (WithDistinct Scope n)
 
 class ScopeReader m => ScopeExtender (m::MonadKind1) where
-  extendScope :: Distinct l => ScopeFrag n l -> m l r -> m n r
+  extendScope :: Distinct l => ScopeFrag n l -> (Ext n l => m l r) -> m n r
```
apaszke, line 97: What does `Ext n l` mean?

src/lib/SaferNames/Builder.hs
```diff
@@ -0,0 +1,296 @@
...
+            -> (forall l. Ext n l => m l (Block l, a))
+            -> m n (Atom n, a)
+buildLamAux ty arr fEff fBody = undefined
+
+buildLam :: (EffectsReader m, BindingsReader m)
+         => Arrow
+         -> Type n
+         -> (forall l. Ext n l => Atom l -> m l (Block l))
+         -> m n (Atom n)
+buildLam _ _ _ = undefined
```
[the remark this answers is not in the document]
dougalm: Yes, I'll try to finish most of the implementation before merging. But none of it will really get tested until we use it in inference, which will be the next PR.

src/lib/SaferNames/Name.hs
dougalm: It means that `l` is a (not-necessarily-fresh) extension of `n`. See 8ffde6164c07c020b3831d620b12a1dcf157dce6

apaszke: Is that not implied by the existence of `ScopeFrag n l`? How can you have a scope fragment when the target is not an extension?

apaszke: Ah I guess that this `Ext` is in a negative position, so you're only adding more evidence to the callback. Fair enough

PR #647, opened 2021-09-10
Safe names builder

dougalm:
Safe-names version of builder monad. We handle continuation-accepting functions like `buildLam` using a new predicate, `Emits`, which gives permission to emit decls into a mutable name space. For example, the following type gives has `Emits` constraint on `l` but not on `n`, which means that the body of the lambda may produce decls, but building the lambda itself doesn't.

```
buildLam
  :: Builder m
  => Type n                                                        -- argument type
  -> (forall l. (Emits l, Ext n l) => AtomName l -> m l (Atom l))  -- lambda body builder
  -> m n (Atom n)                                                  -- result, a lambda term

```


apaszke (approved):
Not a thorough review, but it looks good after a quick skim. Might be good to go over it the next time we talk, unless it's really mostly threading `Emits` and some bookkeeping changes (`Distinct` not being parameterized by any value).

Issue #497, opened 2021-02-08
How to input a Double?


oxinabox:
Literals seem to be `Float`.
as far as i can tell the [parser.dx code](https://github.com/google-research/dex-lang/blob/a373663ce4fb4b88d7f304f0bb2849eb518bb4c9/lib/parser.dx) doesn't support `Double` (or `Float`)

danieldjohnson:
I believe you've linked to the prototype of pure-Dex userspace parsing, which isn't actually used when parsing normal Dex code. But indeed the Haskell parser implementation that parses Dex code seems to hard-code all non-integer numeric literals as floats:

https://github.com/google-research/dex-lang/blob/a373663ce4fb4b88d7f304f0bb2849eb518bb4c9/src/lib/Parser.hs#L209

(as does the type inference pass:)

https://github.com/google-research/dex-lang/blob/a373663ce4fb4b88d7f304f0bb2849eb518bb4c9/src/lib/Inference.hs#L295

oxinabox:
> I believe you've linked to the prototype of pure-Dex userspace parsing, which isn't actually used when parsing normal Dex code.

Yes, as a work around to not being able to input them directly I thought I might use `parse Double "3.1415926535"`.

> But indeed the Haskell parser implementation that parses Dex code seems to hard-code all non-integer numeric literals as floats:

Yes that's the main issue

PR #652, opened 2021-09-19
WIP: Connecting to ChainRulesCore for JuliaLang AD compat

oxinabox:
Right now this is a very early WIP.
It has a proof of concept that runs, and shows roughly the API bits that need to line up.
I am just creating this PR so I don't lose track of this.

### DIFF
--- a/julia/src/evaluate.jl
+++ b/julia/src/evaluate.jl
@@ -60,8 +60,8 @@ julia> m.y
 "84"
 ```
 """
-function DexModule(source::AbstractString)
-    ctx = dex_eval(PRELUDE, source)
+function DexModule(source::AbstractString, parent_ctx=PRELUDE)
+    ctx = dex_eval(parent_ctx, source)
     ctx == C_NULL && throw_from_dex()
     m =  DexModule(ctx)
     finalizer(m) do _m
--- a/julia/test/chainrules.jl
+++ b/julia/test/chainrules.jl
@@ -1,0 +1,50 @@
+using Pkg: Pkg
+Pkg.activate(dirname(@__DIR__))
+using Test
+using DexCall
+using DexCall: Atom, insert
+
+double_dex = evaluate(
+    raw"\x:Float. 2.0 * x"
+)
+
+function frule_2(darg::Atom, primal_f::Atom, arg::Atom)
+    env = primal_f.ctx
+    env = insert(env, "primal_f", primal_f.ptr)
+    env = insert(env, "darg", darg.ptr)
+    env = insert(env, "arg", arg.ptr)
+
+    m = DexModule(raw"""
+    (primal_out, pushforward) = linearize primal_f arg
+    tangent_out = pushforward darg
+    """,
+    env
+    )
+    return m.primal_out, m.tangent_out
+end
+@test frule_2(evaluate("1.5"), double_dex, evaluate("4.0")) .|> juliaize == (8f0, 3f0)
+
+
+function rrule_1(primal_f::Atom, arg::Atom)
+    env = primal_f.ctx
+    env = insert(env, "primal_f", primal_f.ptr)
+    env = insert(env, "arg", arg.ptr)
+
+    m = DexModule(raw"""
+    (primal_out, pushforward) = linearize primal_f arg
+    pullback = transposeLinear pushforward
+    """,
+    env
+    )
+
+    dex_pullback = m.pullback
+    function pullback(d_out::Atom)
+        return dex_pullback(d_out)
+    end
+    return m.primal_out, pullback
+end
+
+val, pb = rrule_1(double_dex, evaluate("4.0"))
+@test juliaize(val) == 8f0
+
+@test juliaize(pb(evaluate("1.5"))) == 3f0


julia/README.md
```diff
@@ -87,6 +87,8 @@ julia> typeof(convert(Int64, m.x))
 Int64
 ```
 
+The inverse of `juliaize` is `dexize`, it is currently very limited.
```
oxinabox, line 90: I haven't given it proper docs yet, because it only does Float32.
It's mostly just for testing purposes.
We need a proper API for this.

julia/src/api_types.jl
```diff
@@ -55,11 +55,3 @@ function CAtom(atm::Ptr{HsAtom})
...
     return result[]
 end
-
-"""
-    juliaize(x)
-
-Get the corresponding Julia object from some output of Dex.
-"""
-juliaize(x::CAtom) = bust_union(x)
-juliaize(x::Ptr{HsAtom}) = juliaize(CAtom(x))
```
oxinabox, line 65: Moved this so it was with the other definition of `juliaize`.
Sorry for the noise

julia/src/evaluate.jl
```diff
@@ -13,9 +13,37 @@ end
...
 Base.convert(::Type{T}, atom::Atom) where {T<:Number} = convert(T, juliaize(atom))
 
+"""
+    dexize(x)
+
+Get the corresponding Dex object from some output of Julia.
+
+NB: this is currently a hack that goes via string processing.
+"""
+function dexize(x::Float32, _module=PRELUDE, env=_module)
```
oxinabox, line 33: I suspect we do not need to take a `_module` argument

julia/src/chainrules.jl
```diff
@@ -0,0 +1,93 @@
...
+    env = insert(env, "x", x.ptr)
+
+    m = DexModule(raw"""
+    (y, pushforward) = linearize f x
+    pullback = transposeLinear pushforward
+    """,
+    env
+    )
+
+    # It is important that we close over `m` as otherwise the env may be destroyed by GC
```
oxinabox, line 72: This is no longer the case since right now that finalizers is commented out.
And before this is merged we may way move to working out how to attach the finalizer to the context more directly.

julia/src/evaluate.jl
```diff
@@ -60,12 +88,18 @@ julia> m.y
...
     m =  DexModule(ctx)
     finalizer(m) do _m
-        destroy_context(getfield(_m, :ctx))
+        # TODO: Undo commenting this out. But for now this causes a lot of problems.
+        # DexModule will often go out of scope, while a Atom attached to that context still
+        # exists. Possibly we need to make the ctx a mutable struct everywhere, and then
+        # attach the finalizer there.
+        #(also will let us delete some manual destroys in other palces)
+        
+        #destroy_context(getfield(_m, :ctx))
```
oxinabox, line 102: This needs resolving

julia/src/DexCall.jl
```diff
@@ -1,14 +1,16 @@
...
     using CombinedParsers.Regexp
 
-    export evaluate, DexError, DexModule, juliaize, NativeFunction, @dex_func_str
+    export evaluate, DexError, DexModule, dexize, juliaize, NativeFunction, @dex_func_str
 
     include("api_types.jl")
     include("api.jl")
     include("evaluate.jl")
     include("native_function.jl")
+    include("chainrules.jl")
```
oxinabox, line 13:
```suggest
include("chainrules.jl")
```

julia/README.md
apaszke: Sure, we can extend it if you'd like. Just let me know what would be helpful to have.

julia/src/evaluate.jl
apaszke: The `_module` and `env` parameters are a bit of a hack that I added for Python bindings. The rough idea was that `module` is the module an output atom declares itself to be defined in, while `env` is the scope that's really used to evaluate the expression. This is used in the `__call__` implementation where we temporarily extend the prelude with new names that refer to arguments, but then we want to pretend that the result is still defined in the original module that doesn't have those dummies. But now that I think about it, it's only well defined for non-dependent functions, so we should find a different workaround...

apaszke: In any case, I think you can safely ignore `env` and just use the `_module`.

julia/README.md
oxinabox: I guess it is something like `create_literal` that works like `insert` except instead of taking an `Atom` it takes a C compatible value for a Int/Float32/Float64/Array.
Possibly `ctypes` doesn't allow that directly, so maybe it needs to be wrapped into a tagged union?
I guess maybe accepting the same tagged union that comes out of the atom's pointer makes sense.
(Except right now that doesn't support arrays)

apaszke: Wouldn't a function that converts a `CAtom` into an `Atom` be sufficient? That's what I would imagine. And then we can add more cases to `CAtom` if that would be useful.

apaszke: Cool, which cases would you like to have? I'm happy to add them for you

oxinabox: `Float32`, `Float64` (particularly since can't input those as literals https://github.com/google-research/dex-lang/issues/497)
Arrays would be nice, but given we can't currently convert `Atom` to `CAtom` for arrays anyway, that doesn't matter so much.
Integer types would be nice, for completeness, but not particularly interesting for AD.

apaszke: Added that in #657

PR #654, opened 2021-09-20
Make `Fallible`, a class for throwing errors but not catching them.

dougalm:
Our go-to error-handling class had been `MonadError Err` which gives you both
`throwError` and `catchError`. But we never actually use `catchError` to recover
from errors. We just use it to add more error context and then re-raise.
Meanwhile, `catchError` in its full generality can't be implemented by some of
the monads we care about. In particular, it doesn't play well with our
name-scoping system because the value you throw might contain names that don't
make any sense in the handler's context.

In this change, we add the `Fallible` class to model this restricted error
handling. It just has methods for throwing and adding context.

There's a second potential use for `catchError` that we don't currently exploit:
continuing after an error long enough to collect further error information. Adam
had the insight that this looks like an applicative-but-not-monadic pattern: if
you want to sequence two computations where the second computation doesn't
depend on the result of the first, you can run the first and then run the second
even if the first failed. You get to collect all the errors from both
computations before you finally crash. For example, if we're doing type
inference on a decl with a user-annotated type, we'll check that the decl's rhs
matches the type supplied, but even if doesn't (or it's otherwise broken), we
can continue to check the rest of the program with the variable assumed to have
the user-supplied type.

I put this pattern in another class, `SeqFallibleApplicative`. I initially tried
making it part of `Fallible`, but `IO` and `Either Err` can't implement it. It
doesn't have any customers yet but we have the API and implementation if we want
to use it.

### DIFF
--- a/src/Dex/Foreign/Context.hs
+++ b/src/Dex/Foreign/Context.hs
@@ -29,6 +29,7 @@ import TopLevel
 import Parser (parseExpr, exprAsModule)
 import Env hiding (Tag)
 import PPrint
+import Err
 
 import Dex.Foreign.Util
 
@@ -49,7 +50,7 @@ dexCreateContext = do
     Right preludeEnv -> toStablePtr $ Context evalConfig preludeEnv
     Left  err        -> nullPtr <$ setError ("Failed to initialize standard library: " ++ pprint err)
   where
-    evalPrelude :: EvalConfig -> String -> IO (Either Err TopStateEx)
+    evalPrelude :: EvalConfig -> String -> IO (Either Errs TopStateEx)
     evalPrelude opts sourceText = do
       (results, env) <- runInterblockM opts initTopState $
                             map snd <$> evalSourceText sourceText
@@ -96,7 +97,8 @@ dexEvalExpr ctxPtr sourcePtr = do
       (Result [] maybeErr, newState) <- runInterblockM evalConfig env $ evalSourceBlock block
       case maybeErr of
         Right () -> do
-          let Right (AtomBinderInfo _ (LetBound _ (Atom atom))) = lookupSourceName newState v
+          let Right (AtomBinderInfo _ (LetBound _ (Atom atom))) =
+                lookupSourceName newState v :: Except AnyBinderInfo
           toStablePtr atom
         Left err -> setError (pprint err) $> nullPtr
     Left err -> setError (pprint err) $> nullPtr
@@ -105,7 +107,7 @@ dexLookup :: Ptr Context -> CString -> IO (Ptr Atom)
 dexLookup ctxPtr namePtr = do
   Context _ env <- fromStablePtr ctxPtr
   name <- peekCString namePtr
-  case lookupSourceName env $ fromString name of
+  case lookupSourceName env (fromString name) :: Except AnyBinderInfo of
     Right (AtomBinderInfo _ (LetBound _ (Atom atom))) -> toStablePtr atom
     Left  _ -> setError "Unbound name" $> nullPtr
     Right _ -> setError "Looking up an expression" $> nullPtr
--- a/src/lib/Err.hs
+++ b/src/lib/Err.hs
@@ -4,18 +4,26 @@
 -- license that can be found in the LICENSE file or at
 -- https://developers.google.com/open-source/licenses/bsd
 
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE ConstraintKinds #-}
 
-module Err (Err (..), ErrType (..), Except, SrcPos, SrcCtx,
-            throw, throwIf, modifyErr, MonadErr,
-            addContext, addSrcContext, catchIOExcept, liftEitherIO,
-            assertEq, ignoreExcept, pprint, docAsStr, asCompilerErr) where
+module Err (Err (..), Errs (..), ErrType (..), Except, ErrCtx (..),
+            SrcPosCtx, SrcTextCtx, SrcPos,
+            Fallible (..), FallibleM (..), HardFailM (..),
+            runHardFail, throw, throwIf,
+            addContext, addSrcContext, addSrcTextContext,
+            catchIOExcept, liftExcept,
+            assertEq, ignoreExcept, pprint, docAsStr, asCompilerErr,
+            SeqFallibleApplicative, traverseMergingErrs) where
 
 import Control.Exception hiding (throw)
+import Control.Applicative
 import Control.Monad
-import Control.Monad.Except hiding (Except)
+import Control.Monad.Identity
+import Control.Monad.State.Strict
+import Control.Monad.Reader
 import Data.Text (unpack)
 import Data.Text.Prettyprint.Doc.Render.Text
 import Data.Text.Prettyprint.Doc
@@ -23,7 +31,11 @@ import GHC.Stack
 import System.Environment
 import System.IO.Unsafe
 
-data Err = Err ErrType SrcCtx String  deriving (Show, Eq)
+-- === core API ===
+
+data Err = Err ErrType ErrCtx String  deriving (Show, Eq)
+newtype Errs = Errs [Err]  deriving (Show, Eq, Semigroup, Monoid)
+
 data ErrType = NoErr
              | ParseErr
              | TypeErr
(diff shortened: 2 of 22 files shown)


src/Dex/Foreign/Context.hs
```diff
@@ -105,7 +107,7 @@ dexLookup :: Ptr Context -> CString -> IO (Ptr Atom)
 dexLookup ctxPtr namePtr = do
   Context _ env <- fromStablePtr ctxPtr
   name <- peekCString namePtr
-  case lookupSourceName env $ fromString name of
+  case lookupSourceName env (fromString name) :: Except AnyBinderInfo of
```
apaszke, line 110: Why do we need those type annotations now?

src/lib/Err.hs
```diff
@@ -42,61 +54,142 @@ data ErrType = NoErr
...
+             | MonadFailErr
                deriving (Show, Eq)
 
-type Except = Either Err
+type SrcPosCtx  = Maybe SrcPos
+type SrcTextCtx = Maybe (Int, String) -- Int is the offset in the source file
+data ErrCtx = ErrCtx
+  { srcTextCtx :: SrcTextCtx
+  , srcPosCtx  :: SrcPosCtx
+  , messageCtx :: [String] }
```
apaszke, line 65: Why a list of strings?

src/lib/Err.hs
```diff
@@ -42,61 +54,142 @@ data ErrType = NoErr
...
+
+type Except = Either Errs
 
 type SrcPos = (Int, Int)
-type SrcCtx = Maybe SrcPos
 
-type MonadErr = MonadError Err
+class MonadFail m => Fallible m where
+  throwErrs :: Errs -> m a
+  addErrCtx :: ErrCtx -> m a -> m a
```
apaszke, line 74: What does adding context mean here? Error contexts don't seem to be forming a stack so it really seems like a local replacement of the context?

src/lib/Err.hs
```diff
@@ -42,61 +54,142 @@ data ErrType = NoErr
...
+  mergeErrs m1 m2 = do
+    result1 <- catchIOExcept m1
+    result2 <- catchIOExcept m2
+    liftExcept $ mergeErrs result1 result2
+
+-- === SeqFallibleApplicative ===
+
+-- Wraps a Fallible monad, presenting an applicative interface that sequences
+-- actions using the error-concatenating `mergeErrs` instead of the default
+-- abort-on-failure sequencing.
```
apaszke, line 115: I'm not sure if I fully understand the difference between `FallibleApplicative` and `SeqFaillableApplicative`. Could you please elaborate why we need both?

src/lib/Err.hs
```diff
@@ -110,15 +203,53 @@ layout :: LayoutOptions
 layout = if unbounded then LayoutOptions Unbounded else defaultLayoutOptions
   where unbounded = unsafePerformIO $ (Just "1"==) <$> lookupEnv "DEX_PPRINT_UNBOUNDED"
 
+traverseMergingErrs :: (Traversable f, FallibleApplicative m)
+                    => (a -> m b) -> f a -> m (f b)
+traverseMergingErrs f xs =
+  fromSeqFallibleApplicative $ traverse (\x -> SeqFallibleApplicative $ f x) xs
```
apaszke, line 209: nit: this thing under traverse is just `SeqFaillableApplicative . f`

apaszke: and if you want to avoid typing `fromSeqFaillableApplicative`, `coerce` would work too (but might be a little less direct)

src/lib/Err.hs
```diff
@@ -144,3 +275,80 @@ instance Pretty ErrType where
...
+instance Monoid ErrCtx where
+  mempty = ErrCtx Nothing Nothing []
+
+-- === misc util stuff ===
+
+leftmostJust :: Maybe a -> Maybe a -> Maybe a
+leftmostJust (Just x) _ = Just x
+leftmostJust Nothing y  = y
+
+rightmostJust :: Maybe a -> Maybe a -> Maybe a
```
apaszke, line 316: Isn't this `flip leftmostJust`? 😉

apaszke (approved):
Very nice! I'm very glad that we're starting to be more principled about our error handling. It's arguably one of the most important parts of a compiler.

src/lib/Err.hs
```diff
@@ -42,61 +54,142 @@ data ErrType = NoErr
...
 
 type SrcPos = (Int, Int)
-type SrcCtx = Maybe SrcPos
 
-type MonadErr = MonadError Err
+class MonadFail m => Fallible m where
+  throwErrs :: Errs -> m a
+  addErrCtx :: ErrCtx -> m a -> m a
+
+-- We have this in its own class because IO and `Either Err` can't implement it
```
oxinabox, line 76: This is not clear, i feel like it is missing a
"But XYZ *can* and should implement it."

src/lib/Err.hs
```diff
@@ -42,61 +54,142 @@ data ErrType = NoErr
...
+class Fallible m => CtxReader m where
+  getErrCtx :: m ErrCtx
+
+-- We have this in its own class because StateT can't implement it
```
oxinabox, line 80: Again, this is not clear, i feel like it is missing a
"But XYZ *can* and should implement it."

oxinabox:
I like this adding of context. Is this a Haskell pattern?
I feel like most languages are really bad at this.
you either get the original message with very little context, e.g. `ProcessExittedError(-22, `curl`, "http://example.com", "Could not resolve host: example.com")` which is bad because little context on why we were doing this;
or it is caught and and new error thrown with a totally different message that loses  information e.g. `DataSetDownloadFailed("http://example.com")` which now i don't know which of many causes caused this.

IIRC C# has a thing that all exceptions can wrap another exception which I think is for that, but I am not sure i every have seen it used.

src/Dex/Foreign/Context.hs
dougalm: Without them, GHC complains about ambiguous `e` in `Either e a`. Previously it could figure out the `e` from the constraint `MonadError Err`. It's annoying to have to write these annotations. One way around it would be to make `Except` a proper type in its own right, instead of an alias for `Either Errs`.

src/lib/Err.hs
dougalm: It's just to model what we had previously, where we added context with `addContext :: MonadErr m => String -> m a -> m a
` which caught and re-raised with the extra string appended to the end. This way we've reified the context in a data structure. For now it's just those same strings, but we can expand it to include other, more structured sorts of context.

src/lib/Err.hs
dougalm: The implementations are supposed to extend the context using the monoid instance of ErrCtx rather than replacing it. That might deserve a comment.

But your mention of a "stack" made me realize: with a suitably expanded `ErrCtx`, this `Fallible`/`CtxReader` are exactly what we need to accumulate a runtime stack in Simplify so that we can report pre-inlining stack traces at runtime!

src/lib/Err.hs
apaszke: I think it's good that we have this comment, because you can't see what's _not_ implemented from the code, but you can grep for `instance.*CtxReader XYZ` if you want. So IMO pretty good after all.

apaszke:
In Python you can also `raise ... from ...` to chain exceptions and have all of them appear in the stack trace saying that the latter has been a direct cause of the former. But yeah what we do here is pretty standard in Haskell. Monads are a convenient way of modifying how the different statements of your program get composed and this also lets you override error handling as you see fit.

src/lib/Err.hs
dougalm: We might not need `SeqFallibleApplicative` at all (we don't use it currently) but it's different from `FallibleApplicative` because it's a data type, not a class. `SeqFallibleApplicative` is a wrapper that takes a type that implements `FallibleApplicative` and uses its `mergeErrors` method to implement `Applicative`. The type being wrapped probably implements `Applicative` too, but in the usual abort-on-failure way in order to be consistent with its monad instance. Does that make sense?

apaszke: Ah yes, that makes total sense. Thanks for explaining. The name is a little confusing, but tbh I don't have better ideas so we can keep it for now.

dougalm: Agreed, it's a bad name. Maybe I'll try something that at least has `Wrapper` in the name.

src/Dex/Foreign/Context.hs
dougalm: Ok, I just made `Except` a proper type. It wasn't too bad at all and it has other benefits.

PR #655, opened 2021-09-21
Use a record environment in inference

apaszke:
This is a non-functional change that should make it easier to thread
the names of builtin interfaces through the inference process.

### DIFF
--- a/src/lib/Inference.hs
+++ b/src/lib/Inference.hs
@@ -34,7 +34,11 @@ import PPrint
 import Cat
 import Util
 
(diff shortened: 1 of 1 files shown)


dougalm:
Excellent. Good idea. We should probably just start using some form of lenses at some point as a way to standardize our patterns around record accessing/updating so we don't have to think about it.

dougalm (approved):

apaszke:
Agreed about lenses, although it might be good to avoid more heavyweight dependencies. Perhaps there's a smallish implementation that could cover most of our use cases.

PR #657, opened 2021-09-28
Add dexFromCAtom to the C API

apaszke:
This makes it possible to use `CAtom` as a common language for getting
values both into and out of Dex.

cc @oxinabox

### DIFF
--- a/python/dex/__init__.py
+++ b/python/dex/__init__.py
@@ -33,7 +33,7 @@ class Module:
     result = api.lookup(self, api.as_cstr(name))
     if not result:
       api.raise_from_dex()
-    return Atom(result, self)
+    return Atom._from_ptr(result, self)
 
 
 class Prelude(Module):
@@ -52,15 +52,35 @@ def eval(expr: str, module=prelude, _env=None):
   result = api.evalExpr(_env, api.as_cstr(expr))
   if not result:
     api.raise_from_dex()
-  return Atom(result, module)
+  return Atom._from_ptr(result, module)
 
 
 class Atom:
   __slots__ = ('__weakref__', '_as_parameter_', 'module')
 
-  def __init__(self, ptr, module):
+  def __init__(self, value):
+    catom = api.CAtom()
+    if isinstance(value, int):
+      catom.tag = 0
+      catom.value.tag = 1
+      catom.value.value = ctypes.c_int(value)
+    elif isinstance(value, float):
+      catom.tag = 0
+      catom.value.tag = 4
+      catom.value.value = ctypes.c_float(value)
+    else:
+      raise ValueError("Can't convert given value to a Dex Atom")
+    self.module = prelude
+    self._as_parameter_ = api.fromCAtom(ctypes.pointer(catom))
+    if not self._as_parameter_:
+      api.raise_from_dex()
+
+  @classmethod
+  def _from_ptr(cls, ptr, module):
+    self = super().__new__(cls)
     self._as_parameter_ = ptr
     self.module = module
+    return self
 
   def __del__(self):
     # TODO: Free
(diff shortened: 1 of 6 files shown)


python/dex/__init__.py
```diff
@@ -52,15 +52,35 @@ def eval(expr: str, module=prelude, _env=None):
...
+  def __init__(self, value):
+    catom = api.CAtom()
+    if isinstance(value, int):
+      catom.tag = 0
+      catom.value.tag = 1
+      catom.value.value = ctypes.c_int(value)
+    elif isinstance(value, float):
+      catom.tag = 0
+      catom.value.tag = 4
+      catom.value.value = ctypes.c_float(value)
```
oxinabox, line 70: Why only `Int64` and `Float32`?

I thought floats in python were `Float64`s precision.
And `Int`'s were arbitrary precision?

I have no idea how ctypes handles these though

python/dex/__init__.py
```diff
@@ -52,15 +52,35 @@ def eval(expr: str, module=prelude, _env=None):
...
+    else:
+      raise ValueError("Can't convert given value to a Dex Atom")
+    self.module = prelude
```
oxinabox, line 73: Shouldn't this be an optional argument, that is defaulted to `prelude` ?
Or is it fine because as a literal it is harmless to add it to the prelude and every other module is sure to have access to things from `prelude` ?

python/dex/api.py
```diff
@@ -17,11 +17,22 @@ def tagged_union(name: str, members: List[type]):
...
 
-CLit = tagged_union("Lit", [ctypes.c_int64, ctypes.c_int32, ctypes.c_int8, ctypes.c_double, ctypes.c_float])
+CLit = tagged_union("Lit", [
+  ctypes.c_int64,
+  ctypes.c_int32,
+  ctypes.c_uint8,
+  ctypes.c_double,
+  ctypes.c_float,
+  ctypes.c_uint32,
+  ctypes.c_uint64
```
oxinabox, line 34: A matching change to this seems like it would be needed in the Julia code.
Which would be to add `UInt32` and `UInt64` to the ends of the list of types here:
https://github.com/google-research/dex-lang/blob/07550931f59bc5b4148e47ec940a177e26518872/julia/src/api_types.jl#L43

python/tests/jax_test.py
```diff
@@ -72,12 +72,12 @@ def test_vmap_unbatched_array(self):
...
     add_two = primitive(dex.eval(r'\x:((Fin 2)=>Float). for i. x.i + 2.0'))
-    x = jnp.linspace([0, 3], [5, 8], num=4, dtype=jnp.float32)
+    x = jnp.linspace(jnp.array([0, 3]), jnp.array([5, 8]), num=4, dtype=jnp.float32)
     np.testing.assert_allclose(jax.jit(jax.vmap(add_two))(x), x + 2.0)
 
   def test_vmap_jit_nonzero_index(self):
     add_two = primitive(dex.eval(r'\x:((Fin 4)=>Float). for i. x.i + 2.0'))
-    x = jnp.linspace([0, 3], [5, 8], num=4, dtype=jnp.float32)
+    x = jnp.linspace(jnp.array([0, 3]), jnp.array([5, 8]), num=4, dtype=jnp.float32)
     np.testing.assert_allclose(
```
oxinabox, line 81: All changes in this file are unrelated right?

python/dex/__init__.py
apaszke: Just because I wrote this out quickly to unblock you and this was good enough for a simple test. I decided to downcast `int` and `float`, because `Int` and `Float` in Dex are 32-bit. If someone's serious about precision they can always use NumPy scalars to bundle the dtype they want to get on the Dex side.

python/dex/__init__.py
apaszke: Yeah the `module` should be a set of bindings that includes all definitions of free variables appearing in the atom. But if the atom is a literal then it has no free variables, so any module would do. But a prelude seems like a reasonable default.

python/dex/api.py
apaszke: Well it's not even a bug fix. The old code was correct, it just couldn't handle some of the cases.

python/tests/jax_test.py
apaszke: Yes, they're due to JAX making the NumPy API stricter than it used to be.

oxinabox (approved):

Issue #633, opened 2021-08-25
Setup ChainRulesCore.jl so Julia AD systems can work with Dex


oxinabox:
Follow up to #625

We should connect up ChainRulesCore to the `NativeFunction` callable, so that Julia AD sytems can just work with Dex functions.
This will make it work with a whole bunch of AD systems, including Zygote and Diffractor.
So will allow layers defined in Dex to be used as part of a Flux NN.

This should not be too hard. Might not be too easy eaither
There is already code that does this for Jax.

Basically we need to overload:
`ChainRulesCore.frule` to basically do `linearize`
and `ChainRulesCore.rrule` to basically do `transpose linearize`
Then there would be a little fiddling with types, e.g. `ChainRulesCore.NoTangent` is Dex's `Discrete`.

I think the harder part is going to be getting information in and out of Dex as right now the `NativeFunction` / `dexfunc` stuff probably is not OK with passing in or out functions.

PR #658, opened 2021-09-28
Add Julia API to use from_CAtom

oxinabox:
Follow up to https://github.com/google-research/dex-lang/pull/657
must be merged after that one.

Replaces the `dexize` code from #633  which will need to be rebased on top of this.

### DIFF
--- a/julia/README.md
+++ b/julia/README.md
@@ -2,7 +2,7 @@
 
 DexCall provides a mechanism for calling dex-lang code from JuliaLang.
 Three main mechanism are provided for this: `evaluate`, `DexModule` and the `dex_func` string macro.
-Two helper methods are also provided: `juliaize` and `NativeFunction`.
+Several helper methods are also provided: `dexize`, `juliaize` and `NativeFunction`.
 
 ## `evaluate`: just run a single Dex expression.
 `evaluate` takes in a Dex expression as a string and runs it, returning a `Atom` (see below).
@@ -53,20 +53,20 @@ julia> m.addTwo(m.y)
 "[44., 44., 44.]"
 ```
 
-## Atoms: `juliaize`, `NativeFunction`
+## Atoms: `dexize`, `juliaize`, `NativeFunction`
 
 `evaluate` and the contents of a `DexModule` are returned as `Atom`s.
 These can be displayed, but not much else.
 
 ```julia
 julia> typeof(m.x)
-DexCall.Atom
+Atom
 
 julia> typeof(m.addTwo)
-DexCall.Atom
+Atom
 ```
 
-To convert scalars into julia typed scalars used `juliaize`.
+To convert scalar `Atom`s into julia typed scalars used `juliaize`.
 ```julia
 julia> juliaize(m.x)
 42
@@ -75,16 +75,22 @@ julia> typeof(juliaize(m.x))
 Int32
 ```
 
-It is not presently possible to `juliaize` arrays (but you can get them as the output of functions, see below).
+It is not presently possible to `juliaize`/`dexize` arrays (but you can use them / get them as the input/output of `NativeFunction`s, see below).
 
-You can also use [`convert`](https://docs.julialang.org/en/v1/manual/conversion-and-promotion/#Conversion) to convert the atom to a Julia object.
-In which case it will make the minimal change from the Dex type to get to the type you requested
+You can also use [`convert`](https://docs.julialang.org/en/v1/manual/conversion-and-promotion/#Conversion) to convert between `Atom`sand Julia objects.
+When converting to a Julia typeit will make the minimal change from the Dex type to get to the type you requested.
 ```julia
 julia> typeof(convert(Integer, m.x))
 Int32
 
 julia> typeof(convert(Int64, m.x))
 Int64
+
+julia> convert(Atom, 1.5)
+"1.5"
+
+julia> convert(Number, convert(Atom, 1.5))
+1.5
 ```
 
 To convert function `Atom`s into something you can execute as if it was a regular julia function use `NativeFunction`.
--- a/julia/src/api_types.jl
+++ b/julia/src/api_types.jl
@@ -22,8 +22,8 @@ end
 Represents a tagged union over types `A`, `B` etc.
 Must have a first field `tag::UInt64`
 and a second field `payload` which must be some isbits type that can hold the largest 
-element of the union (Probably a `NTuple{UInt8}` or similar).
-Julia doens't directly support Unions in the mappting to-from C
+element of the union, which you can (e.g.) declare as a custom `primative`.
+This is required as Julia doens't directly support Unions in the mapping to-from C
 so we store the data as arbitary bits then force it to be reinterpretted based on the tag
 """
 abstract type TaggedUnion{T<:Tuple} end
@@ -40,12 +40,13 @@ function force_reinterpret(::Type{T}, raw) where T
     return unsafe_load(Ptr{T}(Base.pointer_from_objref(Ref(raw))))
 end
 
-
+"Holds data for CLit, big enough to hold the larges of its union members, such as a Float64"
+primitive type CLitPayload 64 end
 struct CLit <: TaggedUnion{Tuple{Int64, Int32, UInt8, Float64, Float32, UInt32, UInt64}}
     tag::UInt64
-    payload::NTuple{8, UInt8}  # actually the Union, needs to be big enough to hold largest which is an Int64
+    payload::CLitPayload
 end
-CLit(x) = CLit(CLit_tag(x), force_reinterpret(NTuple{8, UInt8}, x))
+CLit(x) = CLit(CLit_tag(x), force_reinterpret(CLitPayload, x))
 
 CLit_tag(::Int64) = 0x0000_0000_0000_0000
 CLit_tag(::Int32) = 0x0000_0000_0000_0001
@@ -55,12 +56,13 @@ CLit_tag(::Float32) = 0x0000_0000_0000_0004
 CLit_tag(::UInt32) = 0x0000_0000_0000_0005
 CLit_tag(::UInt64) = 0x0000_0000_0000_0006
 
-
+"Holds data for CAtom, big enough to hold the larges of its union members, which is the CRectArray"
+primitive type CAtomPayload 3*64 end
 struct CAtom <: TaggedUnion{Tuple{CLit, CRectArray}}
     tag::UInt64
-    payload::NTuple{3, UInt64}  # actually the Union, needs to be big enough to hold largest which is the CRectArray
+    payload::CAtomPayload
 end
-CAtom(x) = CAtom(0x0000_0000_0000_0000, force_reinterpret(NTuple{3, UInt64}, CLit(x)))
+CAtom(x) = CAtom(0x0000_0000_0000_0000, force_reinterpret(CAtomPayload, CLit(x)))
 CAtom(x::AbstractArray) = throw(DomainError(typeof(x), "Arrays not yet supported"))
 function CAtom(atm::Ptr{HsAtom})
     result = Ref{CAtom}()
--- a/julia/src/api_types.jl
+++ b/julia/src/api_types.jl
@@ -55,11 +55,3 @@ function CAtom(atm::Ptr{HsAtom})
     iszero(success) && throw_from_dex()
     return result[]
 end
-
-"""
-    juliaize(x)
-
-Get the corresponding Julia object from some output of Dex.
-"""
-juliaize(x::CAtom) = bust_union(x)
-juliaize(x::Ptr{HsAtom}) = juliaize(CAtom(x))
--- a/julia/src/api_types.jl
+++ b/julia/src/api_types.jl
@@ -33,6 +33,7 @@ function bust_union(x::TaggedUnion{U}) where U
 end
 bust_union(x) = x  # not a union, leave it alone
 
+"Forces reinterpretting `raw` as a `T`. If sizeof(raw) <: size(T) high bits will be filled with junk."
 function force_reinterpret(::Type{T}, raw) where T
     isbits(raw) || throw(ArgumentError("Can only reinterpret from a isbits type"))
     isbitstype(T) || throw(ArgumentError("Can only reinterpret into a isbits type"))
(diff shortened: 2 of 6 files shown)


julia/src/api_types.jl
```diff
@@ -33,33 +33,38 @@ function bust_union(x::TaggedUnion{U}) where U
...
 end
+CLit(x) = CLit(CLit_tag(x), force_reinterpret(NTuple{8, UInt8}, x))
+
+CLit_tag(::Int64) = 0x0000_0000_0000_0000
+CLit_tag(::Int32) = 0x0000_0000_0000_0001
+CLit_tag(::Int8) = 0x0000_0000_0000_0002
+CLit_tag(::Float64) = 0x0000_0000_0000_0003
+CLit_tag(::Float32) = 0x0000_0000_0000_0004
+CLit_tag(::UInt32) = 0x0000_0000_0000_0005
+CLit_tag(::UInt64) = 0x0000_0000_0000_0006
```
oxinabox, line 56: these could just be 0,1,2 etc.
The optimizer would do the conversions anyway.
But idk feels more honest to write them out in full as the types they are going to end up as.

julia/README.md
```diff
@@ -75,16 +75,22 @@ julia> typeof(juliaize(m.x))
 Int32
 ```
 
-It is not presently possible to `juliaize` arrays (but you can get them as the output of functions, see below).
+It is not presently possible to `juliaize`/`dexize` arrays (but you can use them / get them as the input/output of `NativeFunction`s, see below).
```
apaszke, line 78: We should implement it... What kind of array representation does Julia use? Are they strided arrays? Or contiguous Fortran-ordered arrays?

julia/src/api_types.jl
apaszke: But they will have to get reinterpreted anyway, right? The tag is 1 byte wide, so this is way too many zeros. And it's not even binary?

julia/src/api_types.jl
```diff
@@ -33,22 +33,35 @@ function bust_union(x::TaggedUnion{U}) where U
...
     return unsafe_load(Ptr{T}(Base.pointer_from_objref(Ref(raw))))
 end
 
 
-struct CLit <: TaggedUnion{Tuple{Int64, Int32, Int8, Float64, Float32, UInt64, UInt32}}
+struct CLit <: TaggedUnion{Tuple{Int64, Int32, Int8, Float64, Float32, UInt32, UInt64}}
     tag::UInt64
     payload::NTuple{8, UInt8}  # actually the Union, needs to be big enough to hold largest which is an Int64
 end
+CLit(x) = CLit(CLit_tag(x), force_reinterpret(NTuple{8, UInt8}, x))
```
apaszke, line 48: What's an `NTuple`? Do you reinterpret the bytes of `x` as an 8-byte bit stream?

julia/src/api_types.jl
```diff
@@ -33,22 +33,35 @@ function bust_union(x::TaggedUnion{U}) where U
...
+
+CLit_tag(::Int64) = 0x0000_0000_0000_0000
+CLit_tag(::Int32) = 0x0000_0000_0000_0001
+CLit_tag(::Int8) = 0x0000_0000_0000_0002
```
apaszke, line 52: This is UInt8 by the way. I think I forgot that there was this one small bugfix.

apaszke: This should also be updated in the `CLit` struct definition above.

julia/README.md
oxinabox: Not particularly important, and not blocking for this PR.
But yes, at some point.
I am not sure what a strided array is in this context.
I would have counted a contiguous Fortran ordered array as being a type of strided array.
(and indeed in Julia the default `Array` type is a subtype of the `StridedArray` union, along with many kinds of views, and transposes)

But regardless, yes the default `Array` type is a contiguous Fortran ordered array.
Probably for initial support of this can do a contiguous `C` ordered array, and can take care of the reordering in the API.
Like we do when passing `Array`s to `NativeFunction`s.

julia/src/api_types.jl
oxinabox: >the tag is 1 byte wide, so this is way too many zeros.

No? it is a `UInt64` it is 8 bytes long.
Isn't that what is going on here: https://github.com/google-research/dex-lang/blob/b80d02d7c063a9207175d5afd98f1f16fa984907/src/Dex/Foreign/Serialize.hs#L32-L33 ?

> And it's not even binary?

This is in hex as it uses the `0x` prefix (`0b` for binary).

julia/README.md
apaszke: By strided array I mean that an array is represented as:
1. a data pointer
2. an integer describing rank
3. an array of integers (of length matching rank) describing its shape (extent)
4. an array of integers (of length matching rank) describing the stride in each dimension. That is, `strides[i]` is the increment that should be applied to a data pointer to traverse to the next element along dimension `i`.

Strided array representation is a generalization of C arrays, Fortran arrays, and it can handle various views and transposes without requiring data copies. It's what NumPy and PyTorch use for example and it's extremely widespread.

Since Julia seems to be using it too, I think it makes sense to support this as the general exchange method.

julia/src/api_types.jl
apaszke: Ahhhh, right... Ok I must have mixed that up. Please ignore this comment.

julia/src/api_types.jl
oxinabox: `NTuple(8, UInt8)` is a short-hand for tuple of 8, `UInt8`s
i.e. `Tuple{UInt8,UInt8,UInt8,UInt8,UInt8,UInt8,UInt8,UInt8}`.
I just needed 64 bit of continuous memory to store the data.
It's type doesn't actually matter.
It occurs to me now, that rather than `NTuple{8, UInt8}` I could have just used a `UInt64`.
Or even declared a new primitive type (which would also work for the 192 byte case below)
`primitive type Payload64 64 end`
Calling it a  `NTuple{8, UInt8}` was helpful because when I was counting how many bytes i needed for the union.
But I think I will change to a custom primitive.

As to why: see this comment in the docstring for `TaggedUnion`
https://github.com/google-research/dex-lang/blob/b80d02d7c063a9207175d5afd98f1f16fa984907/julia/src/api_types.jl#L24-L27

apaszke: Great, thanks for the explanation. One sketchy thing in this solution is that it assumes that the byte ordering is consistent for both C and Julia, but I guess it's fine for now 🤷

julia/README.md
oxinabox: I am not sure, let me summon someone who actually understands how arrays are implemented under the hood.
(Not really part of this PR, but since we have started talking here)

oxinabox: @StefanKarpinski says:

> Adam’s definition of a strided array seems to be accurate, but Julia standard `Array`s are not that general: they have contiguous memory and strides are always in column-major order (last-index contiguous)
> We expose more complex layouts through other array types which ultimately all share memory with a contiguous `Array` object

My summary of that is: there is no type in julia that has a memory layout with that `"1. a data pointer 2. an integer describing rank ..."`
A plain `Array` in julia is a just a Fortran array.
and other layouts use other types, like the `PermutedDimsArray`.

Probably a pretty cute little package could be created that does match that layout and provides nice helpers to convert to and from various julia `AbstractArray` types using stdlib, and/or packages like @Jutho's Strided.jl [StridedView](https://github.com/Jutho/Strided.jl/blob/fe8db2cd9d0fae9b08db08eb4858b0f45ace6f23/src/stridedview.jl#L2-L8)
With the resolve being generally zero-copy.
I am not presently aware of one.

Jutho: I think my `StridedView` is exactly of that type described by @apaszke . It has a rank, a size or extent for each dimension, and a stride in each dimension. There is no particular restriction on the strides being ordered, or the smallest stride being 1, or any of that. Hence, we can compute with each entry of the array a corresponding "memory position" where that entry is to be found. The only difference is that, instead of directly getting data from memory, it assumes that all these "memory positions" correspond to the memory of an underlying (contiguous, as pointed out by Stefan) julia `Array`.

oxinabox: Does it have the right layout that we can directly use it in a `ccall` ?

Jutho: Unlikely, but I have to admit that I just commented on this specific question, without actually knowing what this discussion or even this repository is about. 
The definition is here:
https://github.com/Jutho/Strided.jl/blob/master/src/stridedview.jl#L2

It furthermore has an `offset` (so that the first entry of the `StridedView` is not necessarily the first entry of the underlying parent array), and an (involutive) operation like `conj` that can can be applied on the fly.

julia/README.md
```diff
@@ -75,16 +75,22 @@ julia> typeof(juliaize(m.x))
...
 
-You can also use [`convert`](https://docs.julialang.org/en/v1/manual/conversion-and-promotion/#Conversion) to convert the atom to a Julia object.
-In which case it will make the minimal change from the Dex type to get to the type you requested
+You can also use [`convert`](https://docs.julialang.org/en/v1/manual/conversion-and-promotion/#Conversion) to convert between `Atom`sand Julia objects.
+When converting to a Julia typeit will make the minimal change from the Dex type to get to the type you requested.
```
oxinabox, line 81:
```suggestion
When converting to a Julia type it will make the minimal change from the Dex type to get to the type you requested.
```

julia/src/api_types.jl
```diff
@@ -33,33 +33,40 @@ function bust_union(x::TaggedUnion{U}) where U
...
+
+CLit_tag(::Int64) = 0x0000_0000_0000_0000
+CLit_tag(::Int32) = 0x0000_0000_0000_0001
+CLit_tag(::UInt8) = 0x0000_0000_0000_0002
+CLit_tag(::Float64) = 0x0000_0000_0000_0003
+CLit_tag(::Float32) = 0x0000_0000_0000_0004
+CLit_tag(::UInt32) = 0x0000_0000_0000_0005
+CLit_tag(::UInt64) = 0x0000_0000_0000_0006
 
+"Holds data for CAtom, big enough to hold the larges of its union members, which is the CRectArray"
```
oxinabox, line 59:
```suggestion
"Holds data for CAtom, big enough to hold the largest of its union members, which is the CRectArray"
```

julia/src/api_types.jl
oxinabox: Isn't that promised by endianness?
So it should be fixed by architecture?
(unless you are on a rare mixed architecture machine, and choosing to used different things in your Julia and C compilers?)
Or does endianness/C not make any promises about were things endup if you give a pointer to data that is too big (I don't see how it could chose anywhere but the start)

julia/README.md
apaszke: Ok, so since strided arrays are a generalization of contiguous arrays, I'm thinking of just making those the canonical representation. You can always derive the strides of a fortran array from its sizes, so you should be able to do ccalls with both native Julia arrays and the strided extensions.

julia/src/api_types.jl
apaszke: Yeah I think it's just important that Julia and the C layout rules agree. It's both a matter of endianness (the order the bytes of a value are written to memory), but also a matter of where the padding in a union is (if you write a 4B value into an 8B union, you could write it to the first half or the second half). It'd be good to think about it carefully, but at this point I'm happy with tests being green 🤷‍♂️

julia/src/api_types.jl
```diff
@@ -22,8 +22,8 @@ end
 Represents a tagged union over types `A`, `B` etc.
 Must have a first field `tag::UInt64`
 and a second field `payload` which must be some isbits type that can hold the largest 
-element of the union (Probably a `NTuple{UInt8}` or similar).
-Julia doens't directly support Unions in the mappting to-from C
+element of the union, which you can (e.g.) declare as a custom `primative`.
```
apaszke, line 25: nit: typo in "primative"

julia/src/api_types.jl
```diff
@@ -33,33 +33,40 @@ function bust_union(x::TaggedUnion{U}) where U
 end
 bust_union(x) = x  # not a union, leave it alone
 
+"Forces reinterpretting `raw` as a `T`. If sizeof(raw) <: size(T) high bits will be filled with junk."
```
apaszke, line 36: Is this even safe if `sizeof(raw) < size(T)`? In that case wouldn't `unsafe_load` read memory out of bounds, potentially leading to a segfault?

julia/test/evaluate.jl
```diff
@@ -17,10 +17,20 @@
...
 
-        @test juliaize(evaluate("IToW8 65")) === Int8(65)
+    @testset "juliaize-dexize round-trip" begin
+        @test juliaize(dexize(Int64(3))) === Int64(3)
+        @test juliaize(dexize(Int32(3))) === Int32(3)
+        @test juliaize(dexize(UInt8(3))) === UInt8(3)
+        @test juliaize(dexize(Float64(3))) === Float64(3)
+        @test juliaize(dexize(Float32(3))) === Float32(3)
+        @test juliaize(dexize(UInt64(3))) === UInt64(3)
+        @test juliaize(dexize(UInt32(3))) === UInt32(3)
```
apaszke, line 30: no need to change anything in this PR, but I'm just curious: does Julia have a package for property based testing? This whole test ultimately says that `juliaize . dexize == id`.

apaszke (approved):
LGTM. Is there anything else you'd like to change before we merge this?

julia/src/api_types.jl
oxinabox:
```suggestion
element of the union, which you can (e.g.) declare as a custom `primitive`.
```

julia/src/api_types.jl
oxinabox: I *think* we are safe, as we know the memory that is out of bounds is still memory we have access to.
Because C unions take up the space of the largest element.
So we know our program has that memory reserved, so no segfault.
It is probably unsafe in general, but since we just use it for busting up unions, I think we are in the clear.

julia/test/evaluate.jl
oxinabox: It does not.
Would be great.

oxinabox:
yes, lets merge this

julia/src/api_types.jl
StefanKarpinski: Julia generally tries to match default C padding for structs, although, of course that's dependent on the C compiler, but in practice Clang matches GCC (intentionally), and that's what we do. Unlike C, which allows tweaking struct layout, Julia doesn't have any good tools for that (we'd probably add some low level hook and add macros that emit those hooks).

apaszke: Thanks for elaborating! It's a bit sad that C API is the standard FFI format given how underspecified it is 😕

julia/README.md
apaszke: Well it depends. We have a choice between specializing on the particular input and compiling for that, or emitting code that will be generic over all possible configurations of sizes and strides. But flexibility in that is a good thing, we can always try to figure out what makes more sense based on how people end up using this.

Also, I'm pretty sure that the indexing expression you mentioned actually is representable with strided arrays, but you do need the offset pointer that I skipped and @Jutho mentioned. But of course there are always edge cases, like indexing with another array, which will not be representable as strided arrays.

oxinabox:
Shall we merge this?

apaszke:
Ugh, sorry. I completely forgot about it!

apaszke:
Thanks Lyndon and sorry for the unnecessary delay.

PR #659, opened 2021-09-30
Make integer literals polymorphic

apaszke:
Previously, integer literals were always inferred to be of type `Int32`,
which was annoying for multiple reasons (they couldn't be used in 64-bit
computations easily, nor they could be used to represent floats). With
this change, every integer literal in the surface program is replaced by
the expression `fromInteger <literal>`, with the function being defined
in a type class:
```
interface FromInteger a where
  fromInteger : Int64 -> a
```

All this initially seemed like a simple change, but it quickly turned
out not to be the case. Below I outline some of the challenges I hit and
the solutions I arrived at.

*Protolude*

Prelude is a default, but optional part of the language (it can be even
empty when using `--prelude /dev/null`), but we need `fromInteger` to be
in scope to even be able to perform inference. So this patch also adds
_Protolude_, which is a non-optional standard library that's built into
the compiler. At the moment it's constructed in Haskell using the
builder monad and switching to a source-based specification would be
tricky, because we would need the protolude definitions to perform type
inference over protolude (there's no fully-annotated source syntax for
Dex).

*Integer literals are no longer Atoms*

This makes type inference significantly harder, because we use literals
all over the place in our types (think `Fin 5`). Because of that, I had
to add a system of "delayed unification" tasks that make it possible to
defer unification until the end of inference process.

*Integer literals cannot be fully evaluated before dictionary synthesis*

So far we had a strict phase separation between type inference and
dictionary synthesis. With this patch, some dictionary synthesis is
performed during inference (really during delayed unification).

_TODO:_ I don't think there's any good reason to keep those two separate
forever, especially now that the boundary is leaky. We should just merge
them together.

*Defaulting*

Previously expressions such as `1 == 1` were considered well typed and
it makes sense to try to keep up that behavior. The inference pass now
implements a defaulting mechanism which ensures that when the type
variables associated with integers are unconstrained, they will get
inferred as `Int32` (which really should be the same as whatever
precision the prelude-defined `Int` has).

*Type inference for table literals*

The inference rule for table literals assumed that we either have to
know the concrete type of the table (i.e. no free variables), or the
table had to have a `Fin n=>a` type for `n` matching the size of the
literal. However, with integer literals getting replaced by inference
vars (to be replaced by delayed evaluation results), we ended up only
being able to type `Fin n` tables.

This is obviously very unsatisfactory, so I worked out a completely new
inference scheme for tables. It's a bit more complicated and also uses
the delayed unification machinery. But it's also significantly stronger:
it no longer depends on the `Concrete` hack, and is able to infer rich
index sets if they're implied by the program, e.g. when they com from
function argument annotations.

_TODO:_ Remove `Concrete` from type inference. We don't use it anymore,
but I didn't want to make this patch any larger.

### DIFF
--- a/src/lib/Inference.hs
+++ b/src/lib/Inference.hs
@@ -6,6 +6,7 @@
 
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE ViewPatterns #-}
 
 module Inference (inferModule, synthModule) where
@@ -13,6 +14,7 @@ module Inference (inferModule, synthModule) where
 import Control.Applicative
 import Control.Monad
 import Control.Monad.Reader
+import Control.Monad.State.Strict
 import Data.Maybe (fromJust)
 import Data.Foldable (fold, toList)
 import Data.Functor
@@ -37,6 +39,7 @@ import Err
 data UInferEnv = UInferEnv
   { inferSubst        :: SubstEnv
   , srcCtx            :: SrcPosCtx
+  , protoludeScope    :: ProtoludeScope
   }
 type UInferM = ReaderT UInferEnv (BuilderT (SolverT Except))
 
@@ -53,9 +56,9 @@ pattern Check t <-
 
 {-# COMPLETE Infer, Check #-}
 
-inferModule :: Bindings -> UModule -> Except Module
-inferModule scope (UModule uDecl sourceMap) = do
-  (evaluated, ((bindings, synthCandidates), decls)) <- runUInferM mempty scope mempty do
+inferModule :: Bindings -> SynthCandidates -> UModule -> ProtoludeScope -> Except Module
+inferModule scope scs (UModule uDecl sourceMap) protolude = do
+  (evaluated, ((bindings, synthCandidates), decls)) <- runUInferM mempty scope scs protolude do
     substEnv <- inferUDecl uDecl
     substBuilder substEnv $ EvaluatedModule mempty mempty sourceMap
   let evaluated' = addSynthCandidates evaluated synthCandidates
@@ -87,10 +90,10 @@ addSynthCandidates (EvaluatedModule bindings scsPrev sourceMap) scs =
   EvaluatedModule bindings (scs <> scsPrev) sourceMap
 
 runUInferM :: (HasVars a, Subst a, Pretty a)
-           => SubstEnv -> Scope -> SynthCandidates
+           => SubstEnv -> Scope -> SynthCandidates -> ProtoludeScope
            -> UInferM a -> Except (a, ((Scope, SynthCandidates), Nest Decl))
-runUInferM env scope scs m = runSolverT do
-  runBuilderT scope scs $ runReaderT m $ UInferEnv env Nothing
+runUInferM env scope scs protolude m = runSolverT $ do
+  runBuilderT scope scs $ runReaderT m $ UInferEnv env Nothing protolude
 
 checkSigma :: UExpr -> (Type -> RequiredTy Type) -> SigmaType -> UInferM Atom
 checkSigma expr reqCon sTy = case sTy of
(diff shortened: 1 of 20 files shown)


apaszke:
@duvenaud Yes, adding this to floats is straightforward at this point. But, perhaps more importantly, it also puts in most of the legwork necessary to get associated types, which is what I want to add next!

src/lib/Inference.hs
```diff
@@ -642,11 +647,58 @@ checkAnn ann = case ann of
...
+typeReductionAsAtom :: SrcPosCtx -> String -> Either (Scope, Atom) Atom -> UInferM Atom
+typeReductionAsAtom ctx msg reduced =
   case reduced of
-    Just ty' -> return $ ty'
-    Nothing  -> throw TypeErr $ "Can't reduce type expression: " ++ pprint ty
+    Right ty'         -> return $ ty'
+    Left (localScope, ty') -> do
+      outerScope <- getScope
+      env <- forM (localScope `envIntersect` freeVars ty') \(AtomBinderInfo ty _) ->
+        freshType ty <&> \case Var v -> v; _ -> error "Unreachable"
```
danieldjohnson, line 696: Could you just use `freshInferenceName ty` here?

src/lib/Inference.hs
```diff
@@ -642,11 +647,58 @@ checkAnn ann = case ann of
...
+      resTy <- substBuilder (env <&> Rename . varName) ty'
```
danieldjohnson, line 697: So this is essentially renaming each of the free variables in `ty'` so that they instead refer to the new type variables created in the previous step?

danieldjohnson (commented):
This is super cool! Great work.

src/lib/Inference.hs
```diff
@@ -642,11 +647,58 @@ checkAnn ann = case ann of
   Nothing -> freshType TyKind
 
 checkUType :: UType -> UInferM Type
-checkUType ty = do
-  reduced <- typeReduceScoped $ withEffects Pure $ checkRho ty TyKind
+checkUType ty@(WithSrc ctx _) =
+  typeReduceScoped (withEffects Pure $ checkRho ty TyKind) >>=
+  typeReductionAsAtom ctx "Failed to reduce type annotation"
+
+-- Delayed unification task. When applied, tries to perform dictionary synthesis
```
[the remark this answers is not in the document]
apaszke: By "applied" I mean when you do `performDelayedSolve` on with this ADT. I'll elaborate.

src/lib/Inference.hs
apaszke: Correct. The idea here is that we evaluate as much as we can, but once we find variables in the local scope that can't be resolved to fully evaluated expressions, we replace them with inference variables and delay their full evaluation until the end of the inference process.

src/lib/Inference.hs
apaszke: Yeah I think I could. Thanks for the suggestion.

src/lib/Inference.hs
```diff
@@ -642,11 +647,58 @@ checkAnn ann = case ann of
...
+    EvalAndUnif (subst env eafOuter) (subst env eafInner) (subst env eafResult)
+                (subst env eafScs) (subst env eafExpected) eafErr
+instance IsDelayedSolve EvalAndUnif where
+  performDelayedSolve EvalAndUnif{..} = do
+    innerNoDict <- zonk eafInner <&> fmap \case
+      (AtomBinderInfo ty (LetBound ann expr)) -> AtomBinderInfo ty $ LetBound ann expr'
+        where
+          comp = traverseExpr (traverseHoles synthDictTop) expr
+          expr' = case runSubstBuilderT eafOuter eafScs comp of
+            Failure  _   -> error $ pprint ty
```
apaszke, line 682: TODO: This should not be a hard error

dougalm (commented):
Looks really great! There's a lot of subtle scope-tracking stuff that I don't understand (like substs that only work if they're scopeless). Rather than trying to understand it now, I'll wait till we apply this patch on top of my safe names version of inference where we'll be forced to make them explicit.

dougalm (approved):

src/lib/Inference.hs
```diff
@@ -324,7 +326,12 @@ checkOrInferRho (WithSrc pos expr) reqTy = do
     value' <- zonk =<< (checkRho value $ VariantTy $ Ext NoLabeledItems $ Just row)
     prev <- mapM (\() -> freshType TyKind) labels
     matchRequirement =<< emit (Op $ VariantLift prev value')
-  UIntLit  x  -> matchRequirement $ Con $ Lit  $ Int32Lit $ fromIntegral x
+  UIntLit  x  -> do
+    fromIntegerName <- protoludeFromIntegerMethod <$> asks protoludeScope
+    fromIntegerFunc <- inferRho $ WithSrc Nothing $ UVar $ UInternalVar fromIntegerName
```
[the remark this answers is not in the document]
apaszke: Yeah exactly. Nothing very smart happens here I think.

PR #660, opened 2021-10-01
Add explicit interface methods

apaszke:
It is often useful to define type -> value mappings, and the standard
way to do that is through interfaces (type classes). However, because
all methods so far had implicit type parameters, any attempt to
associate e.g. an integer with a type was difficult: without explicit
type annotations it often ended up being ambiguous.

This patch allows specifying interface parameters between each method
name and the colon that begins its type annotations, with the mentioned
parameters becoming explicit type parameters of the generated methods.
This lets us remove the awkward `TypeVehicle` abstraction from prelude
and in the future should make it possible to define associated types.
For example, a `Manifold a` interface could have a method declared as
`TangentSpace a : Type`, which would later make it possible to mention
`TangentSpace MyDataType` in types.

### DIFF
--- a/lib/prelude.dx
+++ b/lib/prelude.dx
@@ -676,38 +676,30 @@ def castPtr (ptr: Ptr a) : Ptr b =
   (MkPtr rawPtr) = ptr
   MkPtr rawPtr
 
--- Is there a better way to select the right instance for `storageSize`??
-data TypeVehicle a = MkTypeVehicle
-def typeVehicle (a:Type) : TypeVehicle a = MkTypeVehicle
-
 interface Storable a
   store : Ptr a -> a -> {IO} Unit
   load  : Ptr a ->      {IO} a
-  storageSize_ : TypeVehicle a -> Int
-
-def storageSize (a:Type) -> (d:Storable a) ?=> : Int =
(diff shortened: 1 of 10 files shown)


oxinabox:
would this unblock #343 ?

apaszke:
Not yet, but almost! This PR + #659 + ~50 LOC more are what we need to unblock it!

danieldjohnson:
I vaguely remember having a conversation where we considered spelling this as
```
interface Manifold a
  TangentSpace a : Type
```
as opposed to
```
interface Manifold a
  explicit TangentSpace : Type
```
Do you think that's worth doing?

apaszke:
Oh, that's a nice one! It also has the benefit that you can (1) list a subset of type-class binders (2) permute them. My only concern would be that
```
someMethod a : a -> Int
```
might be a little hard to parse, because the `a` on the lhs implies an argument of type `Type` and the one on rhs of `:` implies an argument of type `a`. But I guess if someone has a method that does take an instance then they won't declare it with an explicit type parameter?

apaszke:
I switched the implementation to use what @danieldjohnson has suggested. It has a temporary restriction that you either specify all or no type parameters (and without any permutations), but we can relax it in the future.

PR #664, opened 2021-10-17
Safe-names implementation of inference pass

dougalm:
Safe-names implementation of inference pass

This uses the safe names path to get from source text all the way through to the typed core IR. It hands off to the unsafe path for dictionary synthesis onwards.

It currently gets through the prelude but doesn't pass all the tests. I'll work on fixing the failures one by one, but I think this is at least ready for review now. The hardest tests to pass will be those for type errors, where we may not be able to get messages that are char-for-char identical to the previous ones. In particular the `renameForPrinting` function, where we give human-readable names to unsolved inference variables, is currently stubbed out. It's a funny function because it doesn't obey the usual alpha-equivalence properties.

Unfortunately it's about 3X slower to evaluate the prelude with this version compared to the unsafe version (30s vs 10s). There may be room to improve it at the margin but the reality is that we're doing more work now. We're carrying a renaming substition in more places (e.g. in type checking) and we're checking for escaped variables as we leave scopes. The old implementation was faster, but it was also fundamentally broken. The key to making things fast will be fine-grained caching and lazier inlining, which should become much easier now that we have a solid system for reasoning about all the scopes involved.


src/lib/SaferNames/Inference.hs
```diff
@@ -880,6 +1115,314 @@ checkAllowedUnconditionally eff = do
...
+
+-- === Solver ===
+
+newtype SolverSubst n = SolverSubst (M.Map (AtomName n) (Type n))
+
+class (CtxReader1 m, BindingsReader m) => Solver (m::MonadKind1) where
+  zonk :: (SubstE AtomSubstVal e, InjectableE e) => e n -> m n (e n)
+  -- We require the `Emits` constraint to emit solver stuff to ensure that we
+  -- don't do it at the top level. It might be better to have a solver-specific
+  -- version of `Emits`.
```
apaszke, line 1130: Nothing here requires `Emits`?

src/lib/SaferNames/Name.hs
```diff
@@ -13,44 +13,62 @@
 {-# LANGUAGE TypeFamilies #-}
```
apaszke, line 13: I gave up on reviewing this file, because 1,8k lines is a little much 😛  Let me know if there's anything you'd like me to take a look at

PR #692, opened 2021-11-22
Use multiple threads for running tests in CI

apaszke:


oxinabox:
For reference, right now Hosted Github Actions runners are always:
 - 2 core for Windows/Linux
 - 3 core for Mac
 
 https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources
 
 The code here gives some future proofing against that changing.
 And would also allow private runners, that were different.

apaszke:
Yeah I know the specs are public, but I want to avoid eyeballing them. This seemed like a more robust solution.

The downside is that we might be getting some space leaks, because the jobs seem to be running out of memory now...

PR #698, opened 2021-11-24
Make sure type reduction fails when dictionaries cannot be synthesized

apaszke:
Also start stubbing out "MTL1", a 1-scope-kinded version of MTL.


dougalm (approved):
Looks great!

PR #700, opened 2021-11-25
Implicitly abstract interfaces needed by binder types and interface params

apaszke:
This is a translation of #682 to the safer names system.

### DIFF
--- a/src/lib/SaferNames/Inference.hs
+++ b/src/lib/SaferNames/Inference.hs
@@ -21,7 +21,7 @@ import Control.Monad.State
 import Control.Monad.Writer.Strict hiding (Alt)
 import Control.Monad.Reader
 import Data.Foldable (toList, asum)
-import Data.Functor ((<&>))
+import Data.Functor ((<&>), ($>))
 import Data.List (sortOn)
 import Data.Maybe (fromJust)
 import Data.String (fromString)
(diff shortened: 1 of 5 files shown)


src/lib/SaferNames/Inference.hs
```diff
@@ -1171,22 +1268,56 @@ bindLamPat (WithSrcB pos pat) v cont = addSrcContext pos $ case pat of
...
-    (_       , Just (d:_)) ->
-      throw TypeErr $ "Couldn't synthesize a class dictionary for: " ++ pprint d
-    (_       , Nothing   ) ->
-      throw TypeErr $ "Couldn't synthesize a class dictionary required by: " ++ pprint ty
-    (Nothing , Just []   ) ->
-      throw TypeErr $ "Can't reduce type expression: " ++ pprint ty
-    (Just ty', Just []   ) -> return ty'
+checkAnnWithMissingDicts :: (EmitsInf o, Inferer m)
+                         => Maybe (UType i)
+                         -> ([Type o] -> (forall o'. (EmitsInf o', Ext o o') => m i o' (Type o')) -> m i o a)
```
dougalm, line 1278: It takes a bit of work to understand these rank-3 types. Maybe at least describe what the `forall o'. (EmitsInf o', Ext o o') => m i o' (Type o')` argument to the continuation is supposed to be. It seems to be either "generate fresh type" or "check type, reducing again"? If it was me, I'd probably just defunctionalize by making an ADT for those two cases.

An alias like `type DictType = Type` might also help understand some of these signatures.

dougalm (approved):
Looks great!

src/lib/SaferNames/Inference.hs
apaszke: Yeah it took some effort to write out too, but in the end I don't think it's too bad. The type alias is a very good idea and I'll try to add it in a few places like this one!

PR #703, opened 2021-12-03
Make sure data and type constructors are always fully applied

apaszke:
Previously we allowed them to be partially applied, in which case they
would have function types. This approach is a little surprising and
makes the implementation of interface dependencies for data definitions
more difficult, so this patch makes it so that data and type
constructors are always elaborated as nested lambda expressions that
contain fully applied `DataCon`s or `TypeCon`s in their bodies.

Note that this doesn't really change anything important. It's only a
setup for follow-up work.

### DIFF
--- a/src/lib/SaferNames/Type.hs
+++ b/src/lib/SaferNames/Type.hs
@@ -14,7 +14,7 @@
 module SaferNames.Type (
   HasType (..), CheckableE (..), CheckableB (..),
   checkModule, checkTypes, getType, litType, getBaseMonoidType,
-  instantiatePi, checkExtends, applyDataDefParams, indices,
+  instantiatePi, checkExtends, checkedApplyDataDefParams, indices,
   caseAltsBinderTys, tryGetType, projectLength,
   sourceNameType) where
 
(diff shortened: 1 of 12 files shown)


src/lib/SaferNames/Type.hs
```diff
@@ -1013,6 +975,21 @@ applyDataDefParams :: ScopeReader m => DataDef n -> [Type n] -> m n [DataConDef
 applyDataDefParams (DataDef _ bs cons) params =
   fromListE <$> applyNaryAbs (Abs bs (ListE cons)) (map SubstVal params)
 
+checkedApplyDataDefParams :: (BindingsReader m, Fallible1 m) => DataDef n -> [Type n] -> m n [DataConDef n]
+checkedApplyDataDefParams (DataDef _ bs cons) params =
+  fromListE <$> checkedApplyNaryAbs (Abs bs (ListE cons)) params
+
+-- TODO: Subst all at once, not one at a time!
```
dougalm, line 982: The type checking monad carries a substitution so we can use that.  We would then just call `runEnvReaderT`  with the empty substitution from `checkedApplyDataDefParams`

dougalm (approved):
Fantastic. That worked out really nicely.

PR #705, opened 2021-12-06
Handling more cases in safe-names version of simplification and imp translation passes

dougalm:


src/lib/Interpreter.hs
```diff
@@ -58,11 +59,38 @@ evalDecls (Nest (Let b (DeclBinding _ _ rhs)) rest) cont = do
   result <- evalExpr rhs
   extendSubst (b @> SubstVal result) $ evalDecls rest cont
 
+-- TODO: this was mostly copy-paste from `simplifyAtom`. We should either de-dup
+-- by abstracting over what distinguishes them or, even better, make the `Atom`
+-- ADT simple enough that the duplication is acceptable.
```
apaszke, line 64: Couldn't this be handled by customizing the generic traversal to the few cases we're interested in?

dougalm: Yeah, we could, but it's a fair bit of boilerplate and actually there aren't so many generic cases, because it's important that lambda and pi are `substM`, not recurse, and other cases are deliberately missing. Instead, I just made a somewhat specialized `traverseSurfaceAtomNames` combinator that both Simplify and Interpreter can use.

PR #706, opened 2021-12-07
Safe-names telescope-based defunctionalization

dougalm:
This goes a bit further than just safely reimplementing what we had before. Since we needed the telescope for `splitSimpModule` anyway, I used the same machinery in the other places we needed defunctionalization. Seems to work pretty well.

### DIFF
--- a/src/lib/Builder.hs
+++ b/src/lib/Builder.hs
@@ -288,7 +288,7 @@ buildBlock cont = liftImmut do
   Abs decls' result' <- return $ inlineLastDecl decls $ Atom result
   return $ Block (BlockAnn ty') decls' result'
 
-makeBlock :: (HasCallStack, EnvReader m) => Nest Decl n l -> Expr l -> m l (Block n)
+makeBlock :: EnvReader m => Nest Decl n l -> Expr l -> m l (Block n)
 makeBlock decls expr = do
   ty <- getType expr
   let ty' = ignoreHoistFailure $ hoist decls ty
@@ -406,7 +406,7 @@ singletonBinderNest hint ty = do
 buildNaryAbs
   :: (Builder m, SinkableE e, SubstE Name e, SubstE AtomSubstVal e, HoistableE e)
   => EmptyAbs (Nest Binder) n
-  -> (forall l. (Immut l, Ext n l) => [AtomName l] -> m l (e l))
+  -> (forall l. (Immut l, Distinct l, Ext n l) => [AtomName l] -> m l (e l))
   -> m n (Abs (Nest Binder) e n)
 buildNaryAbs (EmptyAbs Empty) body = liftImmut $ Abs Empty <$> body []
 buildNaryAbs (EmptyAbs (Nest (b:>ty) bs)) body = do
@@ -883,6 +883,8 @@ buildTelescopeVal elts telescopeTy = go elts telescopeTy
       return $ DepPair x rest ty
     go _ _ = error "zip mismatch"
 
+-- sorts name-annotation pairs so that earlier names may be occur free in later
+-- annotations but not vice versa.
 toposortAnnVars :: forall e c n. (NameColor c, HoistableE e)
                 => [(Name c n, e n)] -> [(Name c n, e n)]
 toposortAnnVars annVars =
@@ -913,6 +915,8 @@ unpackTelescope atom = do
       DepPairTy (DepPairType _ rest) -> 1 + telescopeLength rest
       _ -> error $ "not a valid telescope: " ++ pprint ty
 
+-- gives a list of atom names that are free in `e`, including names mentioned in
+-- the types of those names, recursively.
 localVarsAndTypeVars
   :: forall m b e n l.
      (EnvReader m, BindsNames b, HoistableE e)
--- a/src/lib/Builder.hs
+++ b/src/lib/Builder.hs
@@ -36,17 +36,21 @@ module Builder (
   inlineLastDecl, fabricateEmitsEvidence, fabricateEmitsEvidenceM,
   singletonBinderNest, runBuilderM, liftBuilder, makeBlock,
   indexToInt, indexSetSize, intToIndex, litValToPointerlessAtom, emitPtrLit,
-  liftMonoidEmpty
+  liftMonoidEmpty,
+  telescopicCapture, telescopicCaptureBlock, unpackTelescope,
   ) where
 
 import Control.Applicative
 import Control.Monad
 import Control.Monad.Reader
 import Control.Monad.Writer.Strict hiding (Alt)
+import qualified Data.Map.Strict as M
 import Data.Functor ((<&>))
 import Data.Foldable (toList)
 import Data.List (elemIndex)
 import Data.Maybe (fromJust)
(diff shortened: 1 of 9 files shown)


src/lib/Type.hs
```diff
@@ -1196,8 +1236,66 @@ projectLength :: (Fallible1 m, EnvReader m) => Type n -> m n Int
...
+checkDataLikeBinderNest (Abs (Nest b rest) UnitE) = do
+  checkDataLike "data con binder" $ binderType b
+  refreshBinders b \_ -> checkDataLikeBinderNest $ Abs rest UnitE
+
+-- checkDataLikeDataCon :: ( EnvReader2 m, EnvExtender2 m
+--                         , SubstReader Name m, Fallible2 m, Immut o)
+--                      => DataConDef n -> m n ()
+-- checkDataLikeDataCon (DataConDef _ bs) = do
+--   refreshBinders bs \_ -> do
+--     mapM_ (checkDataLike "data con binder" . binderAnn) bs
```
apaszke, line 1301: Some leftovers?

src/lib/Syntax.hs
```diff
@@ -1857,6 +1892,9 @@ instance SubstE AtomSubstVal Block
 deriving instance Show (Block n)
 deriving via WrapE Block n instance Generic (Block n)
 
+instance BindsOneAtomName (BinderP AtomNameC Type) where
+  boundAtomBinding (_ :> ty) = MiscBound ty
```
apaszke, line 1896: Instead of using `MiscBound`, would it be better to add another type class that would be required by `binderType`?

src/lib/Syntax.hs
```diff
@@ -137,7 +139,10 @@ data Atom (n::S) =
  | ACase (Atom n) [AltP Atom n] (Type n)
    -- single-constructor only for now
  | DataConRef (DataDefName n) [Atom n] (EmptyAbs (Nest DataConRefBinding) n)
- | BoxedRef (Atom n) (Block n) (Abs Binder Atom n)  -- ptr, size, binder/body
+ -- lhs ref, rhs ref abstracted over the eventual value of lhs ref, type
+ | DepPairRef (Atom n) (Abs Binder Atom n) (DepPairType n)
+ | BoxedRef [(Atom n, Block n)]        -- ptrptrs/sizes
```
apaszke, line 144: Can you add a note explaining when can the block be evaluated? I assume it references memory that starts uninitialized and can only be safely used after some condition is satisfied.

src/lib/Name.hs
```diff
@@ -625,7 +632,7 @@ instance BindsOneName (BinderP c ann) c where
 
 infixr 7 @@>
 class BindsNameList (b::B) (c::C) | b -> c where
-  (@@>) :: b i i' -> [v c o] -> SubstFrag v i i' o
+  (@@>) :: HasCallStack => b i i' -> [v c o] -> SubstFrag v i i' o
```
apaszke, line 635: Debugging leftovers?

src/lib/Imp.hs
```diff
@@ -614,12 +618,18 @@ makeDest allocInfo ty = do
...
   runSubstReaderT idSubst $ translateDeclNest decls $ substM result
 
-makeDestRec :: forall n l.
-               Emits n => Idxs n -> IdxNest n l -> Type l -> DestM n (Dest n)
+type IdxNest = Nest Binder
+
+makeDestRec ::
+  forall n l . Emits n
+    => Idxs n       -- indices from the enclosing table lambas we're building
+    -> IdxNest n l  -- indices from the table type we're building a dest for.
```
apaszke, line 626: I'm not sure if I understand the distinction between those two arguments

src/lib/Builder.hs
```diff
@@ -281,7 +288,7 @@ buildBlock cont = liftImmut do
   Abs decls' result' <- return $ inlineLastDecl decls $ Atom result
   return $ Block (BlockAnn ty') decls' result'
 
-makeBlock :: EnvReader m => Nest Decl n l -> Expr l -> m l (Block n)
+makeBlock :: (HasCallStack, EnvReader m) => Nest Decl n l -> Expr l -> m l (Block n)
```
apaszke, line 291: Debugging leftovers?

src/lib/Builder.hs
```diff
@@ -789,3 +813,130 @@ indexSetSize (VariantTy (NoExt types)) = do
...
+      UnitTy -> 0
+      PairTy _ rest -> 1 + telescopeLength rest
+      DepPairTy (DepPairType _ rest) -> 1 + telescopeLength rest
+      _ -> error $ "not a valid telescope: " ++ pprint ty
+
+localVarsAndTypeVars
+  :: forall m b e n l.
+     (EnvReader m, BindsNames b, HoistableE e)
+  => b n l -> e l -> m l [AtomName l]
+localVarsAndTypeVars b e =
```
apaszke, line 920: Can you please add some docs that explain what this function computes?

src/lib/Builder.hs
```diff
@@ -789,3 +813,130 @@ indexSetSize (VariantTy (NoExt types)) = do
...
+    go (x:xs) (PairTy _ xsTy) = do
+      rest <- go xs xsTy
+      return $ PairVal x rest
+    go (x:xs) (DepPairTy ty) = do
+      xsTy <- instantiateDepPairTy ty x
+      rest <- go xs xsTy
+      return $ DepPair x rest ty
+    go _ _ = error "zip mismatch"
+
+toposortAnnVars :: forall e c n. (NameColor c, HoistableE e)
```
apaszke, line 886: Sort by... what?

src/lib/Builder.hs
```diff
@@ -789,3 +813,130 @@ indexSetSize (VariantTy (NoExt types)) = do
...
+  vTys <- mapM (getType . Var) vs
+  let (vsSorted, tys) = unzip $ toposortAnnVars $ zip vs vTys
+  ty <- liftImmut $ liftEnvReaderM $
+          buildTelescopeTy (map sink vs) (map sink tys)
+  result <- buildTelescopeVal (map Var vsSorted) ty
+  let ty' = ignoreHoistFailure $ hoist bs ty
+  let ab  = ignoreHoistFailure $ hoist bs $ abstractFreeVarsNoAnn vsSorted e
+  return (result, ty', ab)
+
+-- XXX: assumes arguments are toposorted
```
apaszke, line 858: sorted according to what?

apaszke (approved):
Looks nice!

src/lib/Imp.hs
dougalm: It was confusing to me too. My latest change changes the signature a bit and adds some more explanation.

src/lib/Syntax.hs
dougalm: The block can be evaluated any time (but it usually lives under an abstraction, such at the one in `DepPairRef`. The atom is a pointer, with the usual initialization story.

src/lib/Syntax.hs
dougalm: Yes, that might be better

PR #708, opened 2021-12-13
Everything but AD [WIP]

dougalm:
Still to go
 - [x] dependent tables (need safe-names version of Algebra)
 - [ ] foreign exports
 - [ ] web notebook
 - [ ] respect `@noinline` decorator

### DIFF
--- a/src/lib/Algebra.hs
+++ b/src/lib/Algebra.hs
@@ -4,11 +4,17 @@
 -- license that can be found in the LICENSE file or at
 -- https://developers.google.com/open-source/licenses/bsd
 
-module Algebra (offsetToE, elemCountE, applyIdxs,
-                showPolyC, showSumPolyC) where
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+
+module Algebra (
+  Polynomial, ClampPolynomial, PolyName, emitCPoly,
+  emptyMonomial, poly, mono, liftC, sumC, psubst,
+  liftPoly, indexSetSizeCPoly) where
 
 import Prelude hiding (lookup, sum, pi)
 import Control.Monad
+import Data.Functor
 import qualified Data.Foldable as F
 import Data.Ratio
 import Data.Map.Strict hiding (foldl, map)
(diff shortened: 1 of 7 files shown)


src/lib/Imp.hs
```diff
@@ -1074,6 +1053,40 @@ _deviceFromCallingConvention cc = case cc of
...
+elemCountCPoly (Abs bs UnitE) = case bs of
+  Empty -> return $ A.liftPoly $ A.emptyMonomial
+  Nest b rest -> do
+    size <- A.indexSetSizeCPoly $ binderType b
+    rhsElemCounts <- refreshBinders b \(b':>_) s -> do
+      rest' <- applySubst s $ Abs rest UnitE
+      Abs b' <$> elemCountCPoly rest'
+    withFreshBinder NoHint IdxRepTy \b' -> do
+      let sumPoly = A.sumC (binderName b') (sink rhsElemCounts)
+      return $ A.psubst (Abs b' sumPoly) size
```
apaszke, line 1088: It used to be the case that we took advantage of the mutual recursion between counting elements and offsets (in that the total number of elements is obtained by substituting the upper bound for the offset). Why can't we do that anymore?

src/lib/Algebra.hs
```diff
@@ -17,297 +23,268 @@ import Data.List (intersperse)
...
-freeVarsP p = foldMap freeVarsM $ keys p
+-- TODO: figure out what to do here when we make index sets a user-definable
+-- thing
 
-freeVarsC :: Clamp -> Subst Type
-freeVarsC (Clamp p) = freeVarsP p
+indexSetSizeCPoly :: EnvReader m => Type n -> m n (ClampPolynomial n)
+indexSetSizeCPoly (TC con) = case con of
+--   ProdType tys          -> foldl mulC (liftC $ toPolynomial $ IdxRepVal 1) $
+--     indexSetSizeCPoly <$> F.toList tys
```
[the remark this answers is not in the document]
dougalm: sorry, left in a bunch of stuff from the old version, commented out. will fix

src/lib/Imp.hs
dougalm: We could still do that. I wrote it this way because it was a bit clearer to me, even though there's some slight duplication. The other way was neat, but when I first read it, it took me a while to realize the trick. I don't have strong feelings either way.

PR #719, opened 2022-01-05
Add support for interface constraints on `data` types

apaszke:
This should be the last blocker for rolling out the `Ix` constraint in
surface language. Without this syntax it would be impossible to define
a type such as:
```
data MyData n a = MkMyData (n=>a)
```
because there would be no `Ix n` constraint to satisfy the left hand
side of `=>`. Instead, this type should be defined as:
```
data MyData n a [Ix n] = MkMyData (n=>a)
```

### DIFF
--- a/src/lib/Inference.hs
+++ b/src/lib/Inference.hs
@@ -915,18 +915,18 @@ inferUDeclLocal _ _ = error "not a local decl"
 
 inferUDeclTop :: (Mut o, TopInferer m) => UDecl i i' -> m i' o a -> m i o a
 inferUDeclTop (UDataDefDecl def tc dcs) cont = do
-  def' <- liftImmut $ liftInfererM $ solveLocal $ inferDataDef def
+  PairE def' clsAbs <- liftImmut $ liftInfererM $ solveLocal $ inferDataDef def
   defName <- emitDataDef def'
-  tc' <- emitTyConName defName =<< tyConDefAsAtom defName
+  tc' <- emitTyConName defName =<< tyConDefAsAtom defName (Just clsAbs)
   dcs' <- forM [0..(nestLength dcs - 1)] \i ->
-    emitDataConName defName i =<< dataConDefAsAtom defName i
+    emitDataConName defName i =<< dataConDefAsAtom defName clsAbs i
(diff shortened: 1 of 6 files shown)


src/lib/Syntax.hs
```diff
@@ -843,9 +843,8 @@ type UConDef (n::S) (l::S) = (SourceName, Nest (UAnnBinder AtomNameC) n l)
 -- than being scoped names of the proper color of their own?
 data UDataDef (n::S) where
```
dougalm, line 844: So we have interface constraints in the surface language but not the core IR, is that right?

src/lib/Inference.hs
```diff
@@ -935,38 +935,57 @@ inferUDeclTop (UInterface paramBs superclasses methodTys className methodNames)
...
   buildPureNaryLam (EmptyAbs $ binderNestAsPiNest PlainArrow params) \params' -> do
-    defName'' <- sinkM defName'
-    return $ TypeCon sourceName defName'' $ Var <$> params'
-
-dataConDefAsAtom :: TopInferer m => DataDefName o -> Int -> m i o (Atom o)
-dataConDefAsAtom defName conIx = liftBuilder do
+    EmptyAbs clsBs' <- case maybeIfaceReq of
+      Just clsAbs -> sinkM clsAbs >>= (`applyNaryAbs` params')
+      Nothing     -> return $ EmptyAbs Empty
+    buildPureNaryLam (EmptyAbs $ binderNestAsPiNest ClassArrow clsBs') \_ -> do
```
dougalm, line 945: I see, so the interface constraints just become phantom arguments to the type constructor *function*, but they don't appear in the internal data definition itself. I can imagine we might eventually want to change that. For example, we might want the user-defined `Ix n` constraint to be available internally whenever we have something of type `n => a` so we can use it to determine storage requirements. But this is fine for now.

dougalm (approved):
Looks great!

src/lib/Inference.hs
apaszke: Correct, this makes some simplifying assumptions that only work for as long as the type expressions have phantom dependence on type-classes, which is the current situation for `Ix` on my working branch.

PR #720, opened 2022-01-05
Print cost center call stacks along errors when in debug mode

apaszke:
This should make debugging a lot easier, since we can get pretty
reliable stack traces for any `throw` that makes it to the top-level.
Of course using this requires building with `make build-prof` and then
running Dex using the following alias:
```
alias dexprof='stack exec --trace --work-dir .stack-work-prof dex --'
```


dougalm:
This is amazing. How have we lived without it?

PR #721, opened 2022-01-05
Allow local type aliases in both surface language and core

apaszke:
Turns out that at some point we might have lost (or perhaps never
acquired?) the ability to have local type aliases due to limitations
of our type equality relation and cheap reduction. We still treat alpha
equivalence as the first check, but in case it fails we do fall back to
cheap normalization before we conclude the check as unsuccessful.

### DIFF
--- a/src/lib/CheapReduction.hs
+++ b/src/lib/CheapReduction.hs
@@ -5,15 +5,20 @@
 -- https://developers.google.com/open-source/licenses/bsd
 
 {-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE InstanceSigs #-}
+{-# LANGUAGE UndecidableInstances #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 
 module CheapReduction (
-  CheaplyReducibleE (..), cheapReduce, cheapReduceWithDecls, cheapReduceToAtom) where
+  CheaplyReducibleE (..), cheapReduce, cheapReduceWithDecls, cheapNormalize) where
 
+import qualified Data.Map.Strict as M
+import Data.Maybe
 import Data.Functor.Identity
 import Control.Applicative
 import Control.Monad.Trans
 import Control.Monad.Writer.Strict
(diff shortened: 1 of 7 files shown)


src/lib/CheapReduction.hs
```diff
@@ -147,26 +151,54 @@ cheapReduceWithDeclsRec decls cont = case decls of
...
+  cheapReduceE a = liftImmut do
+    a' <- substM a
+    substMap <- fmap M.fromList $ dropSubst $
+      forM (freeVarsList AtomNameRep a') \v -> (v,) <$> cheapReduceName v
+    let subst :: Subst AtomSubstVal o o =
+          newSubst \v -> case getNameColor v of
+            AtomNameRep -> substMap M.! v
+            _ -> Rename v
+    DB bindings <- getDB
+    return $ substE (toScope bindings, subst) a'
```
apaszke, line 180: @dougalm please take a look at this awful trick and let me know if you can think of anything better. This is basically working around the limitation of the `newSubst` which doesn't let you use an effectful function. It's sensible that it has to be pure, but I'm not sure what's a good way of doing what I do here.

On the other hand... what if I just stashed all the bindings and other state and run cheapReduceName in a fresh monad underneath the subst... WDYT?

dougalm: I see. You want to be able to monadically traverse the free names, is that right? Something like:

```haskell
traverseFreeAtomNames :: (SubstE AtomSubstVal e, Monad m)
                      => (AtomName n -> m (Atom n)) -> e n -> m (e n)
```

Actually, an earlier version of `SubstE` worked exactly like this, and we used the monadic traversal to do hoisting (with the `Either Err` monad). But we ended up doing the hoisting in a different way so there were no users of the monadic traversal anymore and I removed it. Maybe we should add it back? We could add it as a method of `SubstE` alongside the pure version. The pure version is just the special case with the identity monad, but it's important enough that I'm tempted to keep it separate for performance reasons.

PR #722, opened 2022-01-05
Make an IR construct for n-ary application.

dougalm:
Previously, every application appeared on its own line. An innocuous expression
like `1.0 + 2.0` would be elaborated in type inference to this:

    v:((Add Float32) ?=> Float32 -> Float32 -> Float32) = (+) Float32
    v1:(Add Float32) = %OpExpr (SynthesizeDict ...)
    v2:(Float32 -> Float32 -> Float32) = v v1
    v3:(Float32 -> Float32) = v2 1.
    v4:Float32 = v3 2.

Now it's just this:

    v:(Add Float32) = %OpExpr (SynthesizeDict ...)
    v1:Float32 = (+) Float32 v 1. 2.

This speeds up prelude evaluation by 30% and it makes the IR more readable.
More importantly, it will let us work with non-inlined first-order n-ary
functions, coming soon.

### DIFF
--- a/src/lib/Inference.hs
+++ b/src/lib/Inference.hs
@@ -577,17 +577,27 @@ inferRho expr = checkOrInferRho expr Infer
 instantiateSigma :: (EmitsBoth o, Inferer m) => Atom o -> m i o (Atom o)
 instantiateSigma f = do
   ty <- tryGetType f
-  case ty of
-    Pi (PiType (PiBinder _ argTy ImplicitArrow) _ _) -> do
-      x <- freshType argTy
-      ans <- emit $ App f x
-      instantiateSigma $ Var ans
-    Pi (PiType (PiBinder _ argTy ClassArrow) _ _) -> do
-      ctx <- srcPosCtx <$> getErrCtx
-      dict <- emit $ Op $ SynthesizeDict ctx argTy
-      ans <- emit $ App f $ Var dict
-      instantiateSigma $ Var ans
-    _ -> return f
+  args <- getImplicitArgs ty
+  naryApp f args
+
+getImplicitArgs :: (EmitsBoth o, Inferer m) => Type o -> m i o [Atom o]
+getImplicitArgs ty = case ty of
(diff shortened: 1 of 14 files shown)


src/lib/Syntax.hs
```diff
@@ -156,13 +157,15 @@ data Atom (n::S) =
...
-   App (Atom n) (Atom n)
+   App (Atom n) [Atom n]
  | Case (Atom n) [Alt n] (Type n) (EffectRow n)
- | Atom (Atom n)
  | Op  (Op  n)
  | Hof (Hof n)
    deriving (Show, Generic)
 
+pattern Atom :: Atom n -> Expr n
+pattern Atom x = App x []
```
apaszke, line 167: I have to say this is a little sketchy 😬

src/lib/Inference.hs
```diff
@@ -766,6 +748,96 @@ checkOrInferRho (WithSrcE pos expr) reqTy = do
...
+
+-- This allows us to make the instantiated params/dicts part of an n-ary
+-- application along with the ordinary explicit args
+inferFunNoInstantiation :: (EmitsBoth o, Inferer m) => UExpr i -> m i o (Atom o)
+inferFunNoInstantiation expr@(WithSrcE pos expr') = do
+ addSrcContext pos $ case expr' of
+  UVar ~(InternalName v) -> do
+    -- XXX: deliberately no instantiation!
+    substM v >>= inferUVar
+  _ -> inferRho expr
```
apaszke, line 769: This case will do the instantiation, right?

apaszke (commented):
Wow, 30% is a lot!

src/lib/Syntax.hs
dougalm: Yeah, you're right. I was trying to save a case but we end up treating them separately in several places anyway, and the dependence on case order becomes a footgun. Let's keep `Atom` as a separate case and make `App` n-ary for n>0.

I merged this one already but I'll make a follow-up PR.

PR #724, opened 2022-01-07
Use `NonEmpty` for n-ary app, and add `NonEmptyNest`, its B-kinded co…

dougalm:
…unterpart.


apaszke (commented):
Nice! This does look much better!

PR #726, opened 2022-01-08
'/usr/bin/env bash' works on systems without /bin/bash.

athas:
Still works on systems with `/bin/bash`, too.


apaszke (approved):
Thanks!

PR #727, opened 2022-01-10
Add alpha-renaming-invariant hashing.

dougalm:
An expression in some scope `n` can contain both free names, of type `Name c n`,
and alpha-renamable bound names, whose binders are in the expression itself. We
hash the free names according to the actual name string but we use a sort of de
Bruijn scheme for the bound names. Occurrences of these names are hashed
according to the lexical position of their binder, numbered from outside to
inside (the opposite of the de Bruijn convention).

This is equivalent to the "locally nameless" approach in the taxonomy laid out
in "[Hashing Modulo Alpha-Equivalence](https://arxiv.org/abs/2105.02856)" by Krzysztof Maziarz and co (section 2.5).
As that paper describes, it has the downside that the hash of `\x. e` is not
computed in terms of the hash of `e`, so if you need both hashes you need to do
two separate passes.

I don't know if that's going to be a problem or not, but I'm planning to use the
hashes for some CSE stuff soon and we might find out then.

### DIFF
--- a/src/lib/Inference.hs
+++ b/src/lib/Inference.hs
@@ -26,6 +26,7 @@ import Data.List (sortOn, intercalate)
 import Data.Maybe (fromJust)
 import Data.String (fromString)
 import Data.Text.Prettyprint.Doc (Pretty (..))
+import qualified Data.HashMap.Strict as HM
 import qualified Data.List.NonEmpty as NE
 import qualified Data.Map.Strict as M
 import qualified Data.Set as S
@@ -1936,7 +1937,7 @@ trySynthDictBlock ty = liftImmut do
 -- TODO: we'd rather have something like this:
 --   data Givens n = Givens (M.Map (Type n) (Given n))
 -- but we need an Ord or Hashable instance on types
-data Givens n = Givens [Type n] [Given n]
+data Givens n = Givens { fromGivens :: HM.HashMap (EKey Type n) (Given n) }
 
 type Given = Block
 
@@ -1985,7 +1986,7 @@ givensFromEnv = liftImmut do
   DB bindings <- getDB
   let (SynthCandidates givens projs _) = getSynthCandidates bindings
   let givensBlocks = map AtomicBlock givens
-  return $ getSuperclassClosure bindings projs (Givens [] []) givensBlocks
+  return $ getSuperclassClosure bindings projs (Givens HM.empty) givensBlocks
 
 extendGivens :: Synther m => [Given n] -> m n a -> m n a
 extendGivens newGivens cont = do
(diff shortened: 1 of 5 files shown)


src/lib/Name.hs
```diff
@@ -1724,6 +1860,15 @@ instance (Traversable f, SubstE v e) => SubstE v (ComposeE f e) where
...
   alphaEqE (ComposeE xs) (ComposeE ys) = alphaEqTraversable xs ys
 
+instance (Traversable f, Hashable (f ()), AlphaHashableE e)
+         => AlphaHashableE (ComposeE f e) where
+  hashWithSaltE env salt (ComposeE xs) = do
+    let h = hashWithSalt salt $ void xs
+    flip execState h $
+      forM_ xs \x -> do
+        curHash <- get
+        put $ hashWithSaltE env curHash x
```
apaszke, line 1870: Wouldn't it be sufficient to have `Foldable f` and then express this with a `fold`? That would feel more natural than going for the state effect IMO

src/lib/Syntax.hs
```diff
@@ -2396,6 +2418,32 @@ instance Store (Effect n)
...
+instance Hashable UnOp
+instance Hashable BinOp
+instance Hashable CmpOp
+instance Hashable BaseType
+instance Hashable LitVal
+instance Hashable ScalarBaseType
+instance Hashable Device
+instance Hashable Arrow
+
+instance Hashable IsCUDARequired
```
apaszke, line 2435: Why do we need to hash those? 🤔 Is that for future changes?

src/lib/Inference.hs
```diff
@@ -1936,7 +1937,7 @@ trySynthDictBlock ty = liftImmut do
 -- TODO: we'd rather have something like this:
 --   data Givens n = Givens (M.Map (Type n) (Given n))
 -- but we need an Ord or Hashable instance on types
-data Givens n = Givens [Type n] [Given n]
+data Givens n = Givens { fromGivens :: HM.HashMap (EKey Type n) (Given n) }
```
apaszke, line 1940: Awesome!!

apaszke (approved):
This is super sleek, I love it. Did synthesis caching speed up the evaluation of prelude or some other examples?

src/lib/Name.hs
dougalm: It's easy to mess these instances up so I was trying to be extremely explicit. But now that it's done, I can replace it with the one-liner.

src/lib/Syntax.hs
dougalm: I just made `AlphaHashableE` instances for anything that previously had `AlphaEqE`. These instances were required for the Imp data types. How knows, maybe we'll eventually want to CSE at the Imp level too.

dougalm:
> Did synthesis caching speed up the evaluation of prelude or some other examples?

Not that I can tell. This change doesn't change the caching -- it replaces the type-to-dict list with a hash map but perhaps querying the list wasn't an important bottleneck.

PR #728, opened 2022-01-10
WIP fleshing out the implementation of Linear A.

axch:
I wanted to let you know that I started filling in some of the gaps in the Linear A implementation.  It's not all that much progress, so I don't know what a good cadence would be for merging; but on the other hand, any comments about whether I'm doing it right or not would be helpful.

### DIFF
--- a/src/LinearA.hs
+++ b/src/LinearA.hs
@@ -427,7 +427,7 @@ jvp :: JVPFuncMap -> Scope -> Subst -> TangentMap -> Expr -> Expr
 jvp funcEnv scope subst env e = case e of
   Ret vs_ [] -> ctx $ Ret ((subst !) <$> vs_) (zipWith (!) envs vs_)
     where (ctx, _, envs) = splitTangents scope env (Var <$> vs_)
-  Ret _  _  -> expectNonLinear
+  Ret _ _  -> expectNonLinear
   LetMixed vs_ [] e1 e2 ->
     ctx $ LetMixed vs vs' (rec ctxScope subst env1 e1) $
       rec ctxScope (envExt subst vs_ vs) (envExt env2 vs_ vs') e2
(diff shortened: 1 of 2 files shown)


oxinabox:
What is Linear A in this context?
I assume not https://en.wikipedia.org/wiki/Linear_A

> Linear A is a writing system that was used by the Minoans (Cretans) from 1800 to 1450 BC, to write the hypothesized Minoan language.

apaszke:
It's a prototype mini language where we try to distill some AD rules. It's just a name we've given it (there is also Linear B, but this will come later).

src/LinearA.hs
```diff
@@ -615,7 +630,21 @@ transposeExpr scope expr cts = case expr of
...
+  --   ( bCtx . LetMixed [] [v'] (Tuple $ map Var vs') . vCtx
+  --   , vScope
+  --   , vMap <> (bMap `M.withoutKeys` (S.fromList vs))
+  --   )
+  --   where
+  --     (bCtx, bScope, bMap) = transposeExpr scope body cts
+  --     vs' = (bMap !) <$> vs
+  --     v':_ = freshVars bScope
+  --     (vCtx, vScope, vMap) = transposeExpr (scopeExt bScope [v']) (Var v) [v']
+  LetUnpack vs v body -> (LetUnpack vs v . bCtx, bScope, bMap)
```
apaszke, line 645: I thought that transposition was supposed to be defined on expressions that only do linear operations, so we shouldn't need this?

apaszke (approved):
Looks nice!

src/LinearA.hs
axch: Exactly.  If the `body` is linear, the whole `LetUnpack` is linear too, and may need to be transposed.  However, the transposition is trivial, because the variable being unpacked (`v`) is non-linear by assumption.

axch:
Based on Github permitting me to do so, I assume the process is that the contributor merges after a change has been reviewed?

PR #739, opened 2022-01-20
Compile first order monomorphic functions with `@noinline` annotation to object code

dougalm:
A first step towards not inlining everything.

### DIFF
--- a/src/lib/Imp.hs
+++ b/src/lib/Imp.hs
@@ -12,7 +12,7 @@
 {-# OPTIONS_GHC -Wno-orphans #-}
 
 module Imp
-  ( toImpFunction, ImpFunctionWithRecon (..)
+  ( toImpFunction, ImpFunctionWithRecon (..), toImpStandaloneFunction
   , PtrBinder, impFunType, getIType) where
 
 import Data.Functor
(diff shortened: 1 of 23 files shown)


src/lib/TopLevel.hs
```diff
@@ -353,26 +387,35 @@ _evalMLIR block' = do
...
-    liftM LiftE $ liftIO $ impToLLVM env logger impFun
+  (_, llvmAST) <- impToLLVM mainFuncName impFun
   let IFunType _ _ resultTypes = impFunType impFun
   let llvmEvaluate = if bench then compileAndBench needsSync else compileAndEval
-  resultVals <- liftIO $ llvmEvaluate logger llvmAST mainFuncName ptrVals resultTypes
+  logger  <- getLogger
+  objFileNames <- liftImmut $ getObjFiles <$> getEnv
+  objFiles <- forM (eMapToList objFileNames) \(objFileName, UnitE) -> do
+    ObjectFileBinding objFile <- lookupEnv objFileName
+    return objFile
```
apaszke, line 411: This looks very expensive! We end up loading all of the precompiled object files for every single evaluation, even if it's just 2 + 2

src/lib/TopLevel.hs
```diff
@@ -331,9 +331,43 @@ execUDecl decl = do
...
+compileTopLevelFun :: (Mut n, MonadPasses m) => NameHint -> Atom n -> m n (Atom n)
+compileTopLevelFun hint f = do
+  ty <- getType f
+  naryPi <- asFirstOrderFunction ty >>= \case
+    Nothing -> throw TypeErr "@noinline functions must be first-order"
+    Just naryPi -> return naryPi
+  fSimp <- simplifyTopFunction naryPi f
+  fImp <- toImpStandaloneFunction fSimp
+  fSimpName <- emitBinding hint $ AtomNameBinding $ SimpLamBound naryPi fSimp
+  fImpName <- emitImpFunBinding hint fImp
```
apaszke, line 350: Do we ever query those simp and imp bindings?

src/lib/LLVMExec.hs
```diff
@@ -153,6 +156,13 @@ standardCompilationPipeline logger exports tm m = do
 
 -- === object file export ===
 
+exportObjectFileVal :: [ObjectFileName n] -> L.Module -> String -> IO (ObjectFile n)
+exportObjectFileVal deps m fname = do
+  withSystemTempFile "objfile.o" \path h -> do
+    exportObjectFile path [(m, [fname])]
+    contents <- B.hGetContents h
+    return $ ObjectFile contents [fname] deps
```
apaszke, line 164: This is very roundabout. You could just do `Mod.moduleObject` instead of `Mod.writeObjectToFile` and avoid the unnecessary filesystem indirection.

src/lib/TopLevel.hs
```diff
@@ -353,26 +387,35 @@ _evalMLIR block' = do
 _evalMLIR = error "Dex built without support for MLIR"
 #endif
 
+-- TODO: there's no guarantee this name isn't already taken. Need a better story
+-- for C/LLVM function names
+mainFuncName :: SourceName
+mainFuncName = "entryFun"
```
apaszke, line 393: This is probably fine, but it would be a good idea to have a prefix we always append to Dex generated function symbols. Something like `__dex__`

src/lib/LLVM/V9/JIT.hs
```diff
@@ -141,12 +154,21 @@ unloadNativeModule NativeModule{..} = do
...
   symbol <- OrcJIT.mangleSymbol compileLayer $ fromString funcName
   Right (OrcJIT.JITSymbol funcAddr _) <- OrcJIT.findSymbolIn compileLayer moduleKey symbol False
   return $ castPtrToFunPtr $ wordPtrToPtr funcAddr
+
+loadObjectFile :: JIT -> IO OrcJIT.ModuleKey -> ObjectFileContents -> IO ()
+loadObjectFile jit makeKey objFileContents = do
+  withSystemTempFile "objfile.o" \path h -> do
+    BS.hPut h objFileContents
+    hFlush h
+    key <- makeKey
```
apaszke, line 172: Is there any good reason why this has to make the key here, instead of accepting it as an argument?

src/lib/Imp.hs
```diff
@@ -705,18 +747,67 @@ extendIdxsTy (idxsTy, idxs) new = do
...
+-- dest for the args and the result
+-- TODO: de-dup some of the plumbing stuff here with the ordinary makeDest path
+type NaryLamDest = Abs (Nest (BinderP AtomNameC Dest)) Dest
+
+makeNaryLamDest :: (ImpBuilder m, EnvReader m)
+                => NaryPiType n -> m n (AbsPtrs NaryLamDest n)
+makeNaryLamDest piTy = liftImmut do
+  DB env <- getDB
+  cache <- get
+  let allocInfo = (LLVM, CPU, Unmanaged) -- TODO! This is just a placeholder
```
apaszke, line 765: Can you please at least add a check that we're compiling for the CPU? This should be possible to query from `ImpBuilder`. As is, people will likely get segfaults with the CUDA backend...

src/lib/Imp.hs
```diff
@@ -705,18 +747,67 @@ extendIdxsTy (idxsTy, idxs) new = do
...
 type DestIdxs n = (EmptyAbs IdxNest n, Idxs n)
+type DepVars n = [AtomName n]
 
 -- TODO: make `DestIdxs` a proper E-kinded thing
 sinkDestIdxs :: DExt n l => DestIdxs n -> DestIdxs l
 sinkDestIdxs (idxsTy, idxs) = (sink idxsTy, map sink idxs)
 
+-- dest for the args and the result
+-- TODO: de-dup some of the plumbing stuff here with the ordinary makeDest path
+type NaryLamDest = Abs (Nest (BinderP AtomNameC Dest)) Dest
```
apaszke, line 758: What is a destination for args? Why can't we just pass them directly?

apaszke (commented):
Nice! Do you have any before/after comparisons of how long does it take us to go through the tests?

I also think there's lots of TODOs that are pretty important to get to soonish. Many of them are likely to bite us quickly IMO.

PR #741, opened 2022-01-21
Define unzipping and transposition rules for App in the Linear A prototype.

axch:
Unzipping in the presence of App is a bit subtle, because we need to
typecheck the unzipped function body in order to know what type to
declare for the tape.  But the Program in which we need to typecheck
it must have all callees already unzipped!  For now, I tied the knot
with Haskell's laziness (instead of trying to explicitly compute the
call graph); seems to work on the examples present in the test suite.

I also carry around the original (pre-unzipping) program to look up
the number of results callees return, so that I know how many
variables to allocate to hold them.

Finally, added better error checking in case an unzipped program fails
to typecheck (for debugging the unzip transformation).

### DIFF
--- a/src/LinearA.hs
+++ b/src/LinearA.hs
@@ -518,22 +518,29 @@ shuffle scope es = go [] [] (freshVars scope) es
 -------------------- Unzip --------------------
 
 unzipProgram :: Program -> Program
-unzipProgram (Program funcMap) = Program $ flip M.foldMapWithKey funcMap $ \name def -> do
-  let (udef, udef') = unzipFunc def
-  M.fromList [(name ++ ".nonlin", udef), (name ++ ".lin", udef')]
+unzipProgram orig@(Program funcMap) = new where
+  new = Program $ flip M.foldMapWithKey funcMap $ \name def -> do
+    -- Tying the knot by passing `new` here does assume that laziness
+    -- will figure out a sensible order in which to actually perform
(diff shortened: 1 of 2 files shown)


src/LinearA.hs
```diff
@@ -518,73 +518,105 @@ shuffle scope es = go [] [] (freshVars scope) es
...
-  M.fromList [(name ++ ".nonlin", udef), (name ++ ".lin", udef')]
-
-unzipFunc :: FuncDef -> (FuncDef, FuncDef)
-unzipFunc (FuncDef formalsWithTys linFormalsWithTys (MixedType retTys linRetTys) body) =
+unzipProgram orig@(Program funcMap) = new where
+  new = Program $ flip M.foldMapWithKey funcMap $ \name def -> do
+    -- Tying the knot by passing `new` here does assume that laziness
+    -- will figure out a sensible order in which to actually perform
+    -- the unzipping so that the type of every unzipped callee is
+    -- available for its unzipped caller.
```
apaszke, line 526: Which means that this will succeed for as long as there is no (mutual) recursion in the program. On the other hand, given that we don't have conditionals in the language, any such program would necessarily fail to terminate anyway, but it is a bit funny that our compiler would fail to terminate as well. In any case, I think this is ok for now.

apaszke (approved):
LGTM!

PR #745, opened 2022-01-25
Add limited support for first-class record labels

apaszke:
This change extends our record system with an option to have a single
field with a statically unknown label in front of the static labels and
the row extension (which we supported previously). For example the type
```
{@v: Int & v: Float & ...rest}
```
represents to a record with a field of type `Int` and a _dynamic label name_,
specified by the variable `v` (which should be of type `Label` --- that's
another addition). The record also has a field with statically known
name called `v` (for maximum confusion) of type `Float` and a
polymorphic tail of other fields given by `rest`.

`Label`-typed values can be created using the `##name` syntax, similarly
to how we use `#name` for lenses. We might want to revisit this sugar in
the future, but this was the path of least resistance for now.

An interesting property of our type system is that it allows both
dynamic labels and label shaddowing, which to the best of my knowledge
is unique. Doing so lets us avoid dealing with lots of awkward typeclass
constraints of the form "dynamic label disjoint form record". It's
possible that there are some contradictions lurking there, but so far it
does seem to be sound. The trick was to forbid the projection of static
fields that are preceded by a dynamically named one. This means that to
unpack a record type as shown above, one _always_ has to unpack the `@v`
field before they project out the `v` field. So for example this is a
type error:
```
{v=vf, ...} = record
```
but this is allowed
```
{@v=dvf, v=vf, ...} = record
```
This makes it possible to resolve the unpacks statically without having
to worry whether the variable `v` could later get instantiaed to `##v`.

### DIFF
--- a/lib/prelude.dx
+++ b/lib/prelude.dx
@@ -22,6 +22,8 @@ Word64 = %Word64
 Byte = Word8
 Char = Byte
 
+Label = %Label
+
 RawPtr : Type = %Word8Ptr
 
 Int = Int32
--- a/src/lib/Builder.hs
+++ b/src/lib/Builder.hs
@@ -21,7 +21,7 @@ module Builder (
   Builder2, BuilderM, ScopableBuilder2,
   runBuilderT, buildBlock, app, add, mul, sub, neg, div',
   iadd, imul, isub, idiv, ilt, ieq, irem,
-  fpow, flog, fLitLike, recGetHead, buildPureNaryLam,
+  fpow, flog, fLitLike, buildPureNaryLam,
   emitMethodType, emitSuperclass,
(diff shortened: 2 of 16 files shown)


duvenaud:
Wonderful!  I'm guessing that once this lands, we'll be able to add generic instances for records such as `Eq`, `Ix`, and maybe even some form of `VSpace` instance.  I'll give it a shot in a few days probably if no one else does.

apaszke:
This might be a useful step in having typeclasses for records, but it's definitely not sufficient yet 😕 There are still some open questions that we have to resolve before we get to that I think.

dougalm (approved):
This looks great! It worked out to be surprisingly simple. Looking forward to seeing multiple dynamic labels and the `HasField` class.

PR #751, opened 2022-01-31
Update to LLVM 12

dougalm:
This required bumping the stack version number, which isn't a bad thing either.
The main change is that llvm-hs made `typeOf` a monadic operation in the
ModuleBuilder monad, which we don't use. I worked around it by invoking the monad
with an empty state, but won't work if querying a type requires access to some
of the already emitted decls.


apaszke (approved):
LGTM

PR #752, opened 2022-01-31
Track allocations in Imp and free them when they're no longer needed.

dougalm:
We've been seeing memory blowups since the safe-names refactor. It turns out
that we were just ... not freeing any memory. A few crucial lines didn't make it
through the refactor. With that fixed, tests finally pass!


apaszke (approved):
Awesome :)

PR #753, opened 2022-02-01
Safe names python bindings update

mattjj:


mattjj:
I think this was superseded by #765 and others.

PR #759, opened 2022-02-06
Complete the Linear A prototype for the linearity paper

axch:
Including the example in the paper (and fixing the name handling bug that example exposed).

### DIFF
--- a/src/LinearA.hs
+++ b/src/LinearA.hs
@@ -369,7 +369,9 @@ typecheckFunc prog@(Program funcMap) name = case typecheck prog env body of
 typecheckProgram :: Program -> Either String ()
 typecheckProgram prog@(Program progMap) = sequence_ $ typecheckFunc prog <$> M.keys progMap
 
--------------------- JVP --------------------
+-------------------- Fresh variables --------------------
+
+type Scope      = S.Set Var
 
 freshVar :: Scope -> Var
 freshVar s = case S.lookupMax s of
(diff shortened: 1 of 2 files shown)


src/LinearA.hs
```diff
@@ -418,18 +447,20 @@ splitTangents scope env es = go scope env (freeVars <$> es)
       True  -> case commonFvs of
         [] -> (subcontext, subscope, (M.restrictKeys env fvs):submaps)
           where (subcontext, subscope, submaps) = go scope (M.withoutKeys env fvs) tfvs
-        _  -> (context . subcontext, subscope, (M.fromList $ zip commonFvs dvs2'):submaps)
+        _  -> (context . subcontext, subscope, this_map:submaps)
```
apaszke, line 450: completely minor nit: so far we've used camel case, so this is slightly inconsistent

src/LinearA.hs
```diff
@@ -572,6 +600,13 @@ unzipExpr orig scope subst expr = case expr of
       e2Subst = envExt (envExt subst vs uvs) vs' uvs'
       e2Scope = scopeExt (scopeExt scope uvs) uvs'
       ((ctx2, scopeCtx2), ue2, ue2') = rec e2Scope e2Subst e2
+  LetUnpack vs v e ->
+      ((LetUnpack uvs (subst ! v) . ctx, ctxScope), ue, ue')
+    where
+      uvs = take (length vs) $ freshVars scope
+      subst' = envExt subst vs uvs
+      scope' = (scopeExt scope uvs)
```
apaszke, line 608: nit: unnecessary parens

src/LinearA.hs
```diff
@@ -724,7 +778,25 @@ transposeExpr scope expr cts = case expr of
...
+    ( dropAll (S.toList freeS) . LetMixed [] cts (ret scope' [] $ map (const LZero) freeLin)
+    , scope'
+    , M.fromList $ zip freeLin cts
+    )
+    where
+      FV freeS freeLinS = freeVars body
+      freeLin = S.toList freeLinS
+      cts = take (length freeLin) $ freshVars scope
+      scope' = scopeExt scope cts
+      dropAll vs = foldl (.) id $ map (\v -> LetMixed [] [] (Drop $ Var v)) vs
```
apaszke, line 797: Couldn't we drop a tuple?

PR #762, opened 2022-02-09
Enable building Dex with Nix

sharadmv:
This PR enables building Dex with Nix, hopefully providing a reproducible/deterministic way of getting Dex working.

Here's a quick guide to the files this PR adds:

* `default.nix` - this is the `nix` file that describes how to build Dex. It mimics the `dex.cabal` file in its flags and also builds the Dex runtime as well as compiling the Dex Haskell source.
* `flake.nix` - [flakes](https://serokell.io/blog/practical-nix-flakes) are a new feature in Nix that enable locking dependencies for reproducible builds. The `flake.nix` wraps `default.nix` but feeds in locked dependencies.
* `flake.lock` - a file that captures the locked dependencies.

`shell.nix` has also been updated to provide some more libraries and `stack.yaml` now has `nix` enabled so it can use the dependencies provided by Nix.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,0 +1,72 @@
+{ pkgs ? import <nixpkgs> {},
+  llvm-hs-src ? pkgs.fetchFromGitHub {
+    owner = "llvm-hs";
+    repo = "llvm-hs";
+    rev = "llvm-12";
+    sha256 = "IG4Mh89bY+PtBJtzlXKYsPljfHP7OSQk03pV6fSmdRY=";
+  },
+  cudaPackage ? pkgs.cudaPackages.cudatoolkit_11,
+  cuda ? false,
+  optimized ? true,
+  live ? true,
+}:
+let
+  llvm-hs-pure = pkgs.haskellPackages.callCabal2nix "llvm-hs-pure" "${llvm-hs-src}/llvm-hs-pure" {
+  };
+  llvm-hs = (pkgs.haskellPackages.callCabal2nix "llvm-hs" "${llvm-hs-src}/llvm-hs" {
+    inherit llvm-hs-pure;
+  }).overrideAttrs (oldAttrs: rec {
+    buildInputs = oldAttrs.buildInputs ++ [
+      pkgs.llvm_12
+    ];
+  });
+  buildFlags = pkgs.lib.optionals optimized [
+    "-foptimized"
+  ] ++ pkgs.lib.optionals live [
+    "-flive"
+  ] ++ pkgs.lib.optionals cuda [
+    "-fcuda"
+    "--extra-include-dirs=${cudaPackage}/include"
+    "--extra-lib-dirs=${cudaPackage}/lib64/stubs"
+  ];
+  cxxFlags = [
+    "-fPIC"
+    "-std=c++11"
+    "-fno-exceptions"
+    "-fno-rtti"
+  ] ++ pkgs.lib.optional cuda "-DDEX_CUDA"
+    ++ pkgs.lib.optional live "-DDEX_LIVE";
+  buildRuntimeCommand =  ''
+      ${pkgs.clang_9}/bin/clang++ \
+      ${builtins.concatStringsSep " " cxxFlags} \
+      -c \
+      -emit-llvm  \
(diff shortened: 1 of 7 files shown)


stack.yaml
```diff
@@ -20,3 +20,7 @@ extra-deps:
   - store-0.7.8@sha256:0b604101fd5053b6d7d56a4ef4c2addf97f4e08fe8cd06b87ef86f958afef3ae,8001
   - store-core-0.4.4.4@sha256:a19098ca8419ea4f6f387790e942a7a5d0acf62fe1beff7662f098cfb611334c,1430
   - th-utilities-0.2.4.1@sha256:b37d23c8bdabd678aee5a36dd4373049d4179e9a85f34eb437e9cd3f04f435ca,1869
+
+nix:
+  enable: false
+  packages: [ libpng llvm_12 pkg-config zlib ]
```
apaszke, line 26: Why do we have this if it has `enable: false`?

shell.nix
```diff
@@ -1,15 +1,16 @@
...
+{ pkgs ? import <nixpkgs> {} }:
+pkgs.stdenv.mkDerivation {
   name = "dex";
-  buildInputs = [
+  buildInputs = with pkgs; [
     cabal-install
-    haskell.compiler.ghc884
-    llvm_9
+    cacert
     clang_9
```
apaszke, line 7: Can we upgrade to clang_12 too?

default.nix
```diff
@@ -0,0 +1,81 @@
...
+    buildInputs = oldAttrs.buildInputs ++ [
+      pkgs.llvm_12
+    ];
+  });
+  buildFlags = (map (flag: "--ghc-option=" + flag) [
+    "-Wall"
+    "-fPIC"
+    "-optP-Wno-nonportable-include-path"
+    "-Wno-unticked-promoted-constructors"
+    "-XCPP"
```
apaszke, line 28: Does Cabal2nix really ignore the GHC flags?

apaszke (approved):
LGTM, some minor points. Might be worth adding a CI for Nix builds at some point too, just to make sure that they don't get broken if we change things up. I'm a little worried by the amount of duplication this creates among our build files, with lots of flags being embedded across Make/Cabal/Stack/Nix.

sharadmv:
> I'm a little worried by the amount of duplication this creates among our build files, with lots of flags being embedded across Make/Cabal/Stack/Nix.

Yea this is a fair point.

I reduced some of the duplication (CXX flags for the dexrt are still duplicated but GHC flags are not). The long term solution to this is to make the `makefile` work with Nix (i.e. use non-hardcoded dependencies).

stack.yaml
apaszke: Sure, that could work!

shell.nix
apaszke: Yeah it should be fixed already.

flake.lock
```diff
@@ -0,0 +1,59 @@
...
+        "rev": "3cecb5b042f7f209c56ffd8371b2711a290ec797",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "type": "github"
+      }
+    },
+    "llvm-hs-src": {
```
apaszke, line 18: Why do we have this lockfile? Can the definitions in nixpgs change? How do we update it if we do want to use some newer versions?

apaszke (approved):
LGTM, but I do have one maintenance-related question

flake.lock
sharadmv: The lockfile is for reproducible builds using Nix flakes. `default.nix` uses the current (global) version of `nixpkgs` on a given machine, which may differ from machine to machine.

On the other hand, a Nix flake will pin all the dependencies, including Nixpkgs. Upstream Nixpkgs may change -- to update dependencies in `flake.lock`, we can run [`nix flake lock`](https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-flake-lock.html). In theory, we'd have CI for the Nix flake so if we update it and it breaks, we'll be alerted by CI.

apaszke:
This is great. Thanks Sharad!

PR #765, opened 2022-02-15
Begin reviving the Python bridge

apaszke:
The Python bindings have not been updated since the safer name rewrite
so they were completely broken. This is the first change that tries to
restore them to their former glory.


oxinabox:
once this is done, I guess I will need to make a matching PR to update the julia bridge

apaszke:
@oxinabox yeah that would be awesome! I was a bit worried that I won't be able to adjust the Julia part myself. But I'll let you know once we figure out all of the details. I'm still unsure about how to organize a few things.

mattjj:
I think this has superseded the #753  that Dougal and I started, so I'm going to close that one.

PR #766, opened 2022-02-18
Add a WIP data frame exploration

apaszke:
cc @duvenaud

### DIFF
--- a/examples/data-frames.dx
+++ b/examples/data-frames.dx
@@ -1,0 +1,1201 @@
+'# Example tables
+
+-- TODO: Index those tables by the string elements or by integers?
+
+def students : (Fin 3)=>{name: String & age: Int & favColor: String} =
+  [ {name="Bob"  , age=12, favColor="blue" }
+  , {name="Alice", age=17, favColor="green"}
+  , {name="Eve"  , age=13, favColor="red"  }
+  ]
+
+def studentsMissing : (Fin 3)=>{name: String & age:(Maybe Int) & favColor:(Maybe String)} =
+  [ {name="Bob"  , age=Nothing, favColor=Just "blue" }
+  , {name="Alice", age=Just 17, favColor=Just "green"}
+  , {name="Eve"  , age=Just 13, favColor=Nothing     }
+  ]
+
+'## TODO: employees
+
+-- | Last Name    | Department ID |
+-- | ------------ | ------------- |
+-- | "Rafferty"   | 31            |
+-- | "Jones"      | 32            |
+-- | "Heisenberg" | 33            |
+-- | "Robinson"   | 34            |
+-- | "Smith"      | 34            |
+-- | "Williams"   |               |
+
+'## TODO: departments
+
+-- | Department ID | Department Name |
+-- | ------------- | --------------- |
+-- | 31            | "Sales"         |
+-- | 33            | "Engineering"   |
+-- | 34            | "Clerical"      |
+-- | 35            | "Marketing"     |
+
+'## TODO: jellyAnon
+
+-- | get acne | red   | black | white | green | yellow | brown | orange | pink  | purple |
+-- | -------- | ----- | ----- | ----- | ----- | ------ | ----- | ------ | ----- | ------ |
+-- | true     | false | false | false | true  | false  | false | true   | false | false  |
+-- | true     | false | true  | false | true  | true   | false | false  | false | false  |
+-- | false    | false | false | false | true  | false  | false | false  | true  | false  |
+-- | false    | false | false | false | false | true   | false | false  | false | false  |
+-- | false    | false | false | false | false | true   | false | false  | true  | false  |
+-- | true     | false | true  | false | false | false  | false | true   | true  | false  |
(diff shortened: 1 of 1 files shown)


apaszke:
Yeah many things have changed in Jan. Here's a quick summary (which we should really turn into a tutorial):

`&` indicates a product type and `|` indicates a sum type. So `{a:Int & b:Float}` is sort of like a `namedtuple` with fields `a` and `b` of type `Int` and `Float` respectively, though unlike a `namedtuple` it doesn't imply any field ordering. `{a:Int | b:Float}` is closer to `Either Int Float`, except that it uses labels and has no implicit left/right ordering (unlike `Either`). Both records (products) and variants (sums) additionally allow extending them with _arbitrary rows_. This was the state of records/variants as of beginning of this year.

You might have already realized that, but this actually gives you a nice way to implement named tensors. `{width: Fin n & height: Fin h}=>Float` is a matrix of `Float`s that can be indexed with expressions such as `x.{width=w, height=h}` (which is equivalent to `x.{height=h, width=w}`). And having row extensions means that you can write functions that only work over certain names and treat all other names as _private dimensions_:
```
def f (x: {width: n & height: m & ...rest}=>Float) : (n & m)=>{&...rest}=>Float =
  for {width, height} {...other}. x.{width, height, ...other}
```

What I did recently is a relaxation of the previous implementation to additionally:

1. **Allow accepting field names as arguments.** In all types above the field names such as `width` or `height` were static and it was impossible (not exactly, but it was hard) to implement a function that would e.g. select a field from a record. This is now possible:
```
def select (l: Label) (x: {@l: t & ...rest}) : (t & {...rest}) =
  {@l=v, ...other} = x
  (v & other)

select ##b {a=1, b="adsf", c=1.0}
> "asdf"
```
(`##b` is a literal that constructs the `Label`-typed value).

2. **Allow multiple extensions of records.** Previously you could only have one `...something` at the end of the record type/pattern, now you can have multiple:
```
def recordConcat (x: {...xRest}) (y: {...yRest}) : {...xRest & ...yRest} = {...x, ...y}

recordConcat {x=1} {y=2}
> {x=1, y=2}
```

3. **Add literals for `Fields`.** The `xRest` in function above is a variable of type `Fields`. Previously values of those could only have been instantiated by type inference and were inaccessible from the surface language. Now you can construct them, using `?` instead of `&` or `|` (since `Fields` work both as record and variant extensions). This lets you implement a record subsetting operation (note the `@` that indicates that we refer to set of fields defined by `f`, which is then followed by an arbitrary extension):
```
def recordSubset (f: Fields} (x: {@...f & ...other}) : ({@...f} & {...other}) =
  {@...f=front, ...back} = x
  (front & back)

recordSubset {x: _ ? y: _} {x=1, y=2, z=4} 
> {x=1, y=2}
```
Note that types of fields can be `_` and are then inferred from record type.

AriMKatz:
Cool! How are these heterogeneous "table" types laid out in memory ?  And they can be accessed without chasing around pointers too much ?

srush:
This is wild. Excited to try it out. It's challenging to wrap my head around what is values and what is types. A couple questions:

* What does & and @ to start a { mean?

* What are the punning rules? What does `for {width, height}` mean?

* What does orderlessness mean indexes? Can the user still call fromOrdinal ?

dougalm:
> Cool! How are these heterogeneous "table" types laid out in memory ? And they can be accessed without chasing around pointers too much ?

They're laid out like a columnar database: each (scalar, right-of-the-arrow) column is a separate array. This is just a consequence of how we represent arrays of structures. Internally we always convert arrays of tuples to tuples of arrays, and records behave just like tuples. So a table of prices and availability of mechanical parts, like `Part => {price : Float & num_in_stock : Int}` would be represented as an array of floats for the prices and an array of integers for the availability. A neat feature is that the key column, `Part`, is represented implicitly.

> What does orderlessness mean [for] indexes? Can the user still call fromOrdinal

Good catch. The order does get exposed through the index conversion. It's officially arbitrary, but in reality it's alphabetical in the record field names. It's a bit like iterating through keys in a supposedly unordered dictionary - you'll get them in some order, but we don't make any promises. This is a bit unsatisfying though. You can imagine a more restricted type class for indices without an ordering constraint.

apaszke:
To complement the "orderless" response: we _used to_ treat them as records in major-to-minor alphabetical order, but now records of indices are no longer indices by default. But, we do support user-defined index sets, so you have to explicitly define the order when you declare a certain record type to be an index. [Here's an example](https://github.com/google-research/dex-lang/blob/65ee2e0677bacafb4ace69123c17d90130a2bace/examples/isomorphisms.dx#L192-L199).

srush:
> What does & and @ to start a { mean?

To answer my own question: 
* {@c:a } -> c is a `Label` instance and a is its type
* {@...f} -> f is an explicit `Fields` instance
* {...f} -> in type `f` is bound to the record type 
* {...(t.i)} -> in code `t.i` is the subrecord instance
* {&...f} -> synonym for {...f}?

apaszke:
(1) and (2) seem correct. I can't fully follow (3), (4) and (5) since their meaning might depend on the context in which they're used.

PR #769, opened 2022-02-23
Start implementing unzipping

apaszke:
It's not fully featured yet, but it's starting to make sense!


axch:
Any reason not to just merge this as-is?

PR #780, opened 2022-03-07
Introduce a monad for reading the scope and env in the *input* name space

dougalm:
Previously, our two-scope-parameter monads, `m i o a`, all had distinct outputs
and shadowy inputs. Instead of maintaining distinctness on the input side we
would just carry a substitution from input names to output names. For anything
that required looking up names, like querying a type, we had to apply the
substitution first. A funny consequence of that was that we needed the input and
output IRs to use the same sort of names so that we could substitute names in
input-space terms with output-space expressions. This is why we use core's
`AtomName` in the `Imp` IR.

This commit adds a two-scope-parameter monad class, `EnvReaderI`, which has
distinctness and env-reading on its input scope parameter. I plan to use it in
the Imp pass next.


apaszke (approved):
Nice!

PR #781, opened 2022-03-08
add `HoistableE e` to the context of the `Hashable (EKey e n)` instance

tscholak:
Hi,
During compilation with ghc 8.10.7, I've been running into this error:
```
[10 of 38] Compiling Name             ( src/lib/Name.hs, dist/build/Name.o, dist/build/Name.dyn_o )

src/lib/Name.hs:1143:10: error:
    • Could not deduce (HoistableE e)
        arising from the superclasses of an instance declaration
      from the context: (AlphaEqE e, AlphaHashableE e)
        bound by the instance declaration at src/lib/Name.hs:1143:10-62
      Possible fix:
        add (HoistableE e) to the context of the instance declaration
    • In the instance declaration for ‘Hashable (EKey e n)’
     |
1143 | instance (AlphaEqE e, AlphaHashableE e) => Hashable (EKey e n) where
     |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
No idea why I'm the only one who has been seeing this, but this patch should fix it.

### DIFF
--- a/src/lib/Name.hs
+++ b/src/lib/Name.hs
@@ -1160,7 +1160,7 @@ eMapToList (EMap m) = [(k, v) | (EKey k, v) <- HM.toList m]
 eMapFromList :: (AlphaEqE k, AlphaHashableE k, HoistableE k) => [(k n, v n)] -> EMap k v n
 eMapFromList xs = EMap $ HM.fromList [(EKey k, v) | (k, v) <- xs]
 
-eSetSingleton :: (AlphaEqE k, AlphaHashableE k) => k n -> ESet k n
+eSetSingleton :: (AlphaEqE k, AlphaHashableE k, HoistableE k) => k n -> ESet k n
 eSetSingleton k = eMapSingleton k UnitE
 
 eSetToList :: ESet k n -> [k n]
--- a/src/lib/Name.hs
+++ b/src/lib/Name.hs
@@ -1151,7 +1151,7 @@ instance Store (e n) => Store (EKey e n)
 data EMap (k::E) (v::E) (n::S) = EMap (HM.HashMap (EKey k n) (v n))
                                  deriving (Show, Generic)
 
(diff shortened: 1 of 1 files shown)


apaszke (commented):
I don't think it hurts us to add this right now, but can you please leave a comment in all places you've modified that this is working around a GHC bug? Technically speaking not all E-kinded things have to satisfy `HoistableE` and at some point we might need the extra generality.

tscholak:
Hi @apaszke,
Perhaps I'm wrong on this, but I don't think that this is a ghc bug. Consider the following example:
```hs
class Foo a

data Bar a = Bar

instance Foo a => Eq (Bar a) where (==) _ _ = True

-- No instance for (Foo a)
--   arising from the superclasses of an instance declaration
-- Possible fix:
--   add (Foo a) to the context of the instance declaration
-- instance Ord (Bar a) where compare _ _ = EQ

instance Foo a => Ord (Bar a) where compare _ _ = EQ
```
Haskell forces me to add the `Foo a` constraint to the instance declaration for `Old (Bar a)` because `Eq` is in `Ord`'s context and the `Eq (Bar a)` instance has already the `Foo a` constraint. 
I believe we have a similar case with `Eq (EKey e n)` and `Hashable (EKey e n)`, where there was a `HoistableE e` constraint on `Eq (...)` but not a `HoistableE e` constraint on `Hashable (...)`.
Are you compiling with ghc 8.10.7? Perhaps ghc got more pedantic in this respect recently.
I gladly add a note, though, just felt you should know why I believe this change is necessary and more correct.

apaszke (approved):
Ok I finally understand that this error is coming from the `Eq` superclass of `Hashable`. This is a very confusing error from GHC...

But in that case LGTM!

apaszke:
Interestingly enough our `stack.yaml` also uses 8.10.7, so I have no clue why we haven't seen that

PR #785, opened 2022-03-09
Add a concatenative Nest to avoid quadratic behavior with unlucky association

apaszke:


axch:
Quoth @apaszke:
```
I benchmarked it and it didn't end up making any difference.

I expect that it's because we're always accumulating the bindings in a reader-like monad, or rather we always have at least one reader-like monad in the monad stack.  I think that this sequentializes the extensions?

Anyway, it didn't make any difference, so why merge it?
```

Therefore closing this PR.

PR #800, opened 2022-03-18
Refactor live output system (dex web, dex watch)

### LINKED ISSUE #799: Dex notebook should not wait for evaluation of cells that no longer exist
Reproduction instructions:

- Start `dex web` on some script
- Create a long-running cell with contents A and save (Dex starts executing A)
- Before that cell completes, edit it and save again, with new contents B
- Dex waits for A to finish executing, flashes the output on the screen, erases it, and begins executing the cell with contents B.

Expected behavior: Abort A and begin executing B immediately.

axch:
I do not expect any functional changes from this, but it clears space to fix #799 in a follow-on PR.


src/lib/Actor.hs
```diff
@@ -1,178 +1,67 @@
...
-  actorCfg :: m (ActorConfig msg)
-
-instance MonadActor msg (Actor msg) where
-  actorCfg = Actor ask
-
-instance MonadActor msg m => MonadActor msg (StateT s m) where
-  actorCfg = lift actorCfg
+-- We also define a send-only channel type, to help ourselves not
+-- accidentally read from channels we aren't supposed to.
+newtype PChan a = PChan { sendPChan :: a -> IO () }
```
apaszke, line 29: Why is a send only channel called a `PChan`? What's the `P` prefix for?

src/lib/Actor.hs
```diff
@@ -1,178 +1,67 @@
...
+-- In this model, an "actor" is just an IO computation (presumably
+-- running on its own Haskell thread) that receives messages on a
+-- Control.Concurrent.Chan channel.  The idea is that the actor thread
+-- only receives information (or synchronization) from other threads
+-- through messages sent on that one channel, and no other thread
+-- reads messages from that channel.
 
-sendFromIO :: PChan msg -> msg -> IO ()
-sendFromIO = sendPChan
+type Actor a = Chan a -> IO ()
```
apaszke, line 25: Can actors send messages back? If not, why do we have a type alias for send-only channels, but not for receive-only channels?

apaszke (approved):
LGTM, but since live output is a sensible grouping of functionality, that we're now splitting across multiple files, I'd suggest adding a submodule. Perhaps we can call it `Live` and then have `Live.Eval`, `Live.Terminal` and `Live.Web`? The terminal and web parts are not only about output, since they do have some degree of interactivity (scrolling, etc.).

src/lib/Actor.hs
axch: Don't know; I kept it from the previous Actor module.  Partial channel?  @dougalm ?

src/lib/Actor.hs
axch: An actor will generally not send a message on its own channel, because that would constitute sending a message to itself (though it could do that, to break up a long computation into chunks that can be interspersed with processing other messages).  However, an actor _may_ send a send-only view of its own channel to other actors, to subscribe themselves to notifications of asynchronous events.  So at creation time, an actor is given a two-way channel to allow that to happen.

Code comment added.

axch:
Good idea about that submodule.  Done.

PR #802, opened 2022-03-21
Add a worker thread to the evaluation driver

### LINKED ISSUE #799: Dex notebook should not wait for evaluation of cells that no longer exist

axch:
This way, Dex compute can be interrupted in response to external events (like the user editing the cell being computed).

Fixes #799 .

### DIFF
--- a/src/lib/Actor.hs
+++ b/src/lib/Actor.hs
@@ -8,7 +8,7 @@ module Actor (PChan, sendPChan, sendOnly, subChan,
               Actor, runActor, spawn,
               LogServerMsg (..), logServer) where
 
-import Control.Concurrent
+import Control.Concurrent (Chan, forkIO, newChan, readChan, ThreadId, writeChan)
 import Control.Monad.State.Strict
 
 import Util (onFst, onSnd)
@@ -32,7 +32,7 @@ type Actor a = Chan a -> IO ()
 newtype PChan a = PChan { sendPChan :: a -> IO () }
 
 sendOnly :: Chan a -> PChan a
-sendOnly chan = PChan $ writeChan chan
+sendOnly chan = PChan $ \ !x -> writeChan chan x
 
 subChan :: (a -> b) -> PChan b -> PChan a
 subChan f chan = PChan (sendPChan chan . f)
--- a/src/lib/Err.hs
+++ b/src/lib/Err.hs
@@ -240,9 +240,10 @@ addSrcTextContext offset text m =
 
 catchIOExcept :: MonadIO m => IO a -> m (Except a)
 catchIOExcept m = liftIO $ (liftM Success m) `catches`
-  [ Handler \(e::Errs)          -> return $ Failure e
-  , Handler \(e::IOError)       -> return $ Failure $ Errs [Err DataIOErr   mempty $ show e]
-  , Handler \(e::SomeException) -> return $ Failure $ Errs [Err CompilerErr mempty $ show e]
+  [ Handler \(e::Errs)           -> return $ Failure e
+  , Handler \(e::IOError)        -> return $ Failure $ Errs [Err DataIOErr   mempty $ show e]
+  , Handler \(e::AsyncException) -> liftIO $ throwIO e
+  , Handler \(e::SomeException)  -> return $ Failure $ Errs [Err CompilerErr mempty $ show e]
   ]
 
 liftExcept :: Fallible m => Except a -> m a
--- a/src/lib/Err.hs
+++ b/src/lib/Err.hs
@@ -242,6 +242,9 @@ catchIOExcept :: MonadIO m => IO a -> m (Except a)
 catchIOExcept m = liftIO $ (liftM Success m) `catches`
   [ Handler \(e::Errs)           -> return $ Failure e
   , Handler \(e::IOError)        -> return $ Failure $ Errs [Err DataIOErr   mempty $ show e]
+  -- Propagate asynchronous exceptions like ThreadKilled; they are
+  -- part of normal operation (of the live evaluation modes), not
+  -- compiler bugs.
   , Handler \(e::AsyncException) -> liftIO $ throwIO e
   , Handler \(e::SomeException)  -> return $ Failure $ Errs [Err CompilerErr mempty $ show e]
   ]
--- a/src/lib/Live/Eval.hs
+++ b/src/lib/Live/Eval.hs
@@ -6,13 +6,14 @@
 
 module Live.Eval (RFragment (..), SetVal(..), watchAndEvalFile) where
 
-import Control.Concurrent (forkIO, readChan, threadDelay)
+import Control.Concurrent (forkIO, killThread, readChan, threadDelay, ThreadId)
 import Control.Monad.Reader
 import Control.Monad.State.Strict
 import qualified Data.Map.Strict as M
 
 import Data.Aeson (ToJSON, toJSON, (.=))
 import qualified Data.Aeson as A
+import Data.Text.Prettyprint.Doc
 import System.Directory (getModificationTime)
 
 import Actor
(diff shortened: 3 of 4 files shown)


src/lib/Actor.hs
```diff
@@ -32,7 +32,7 @@ type Actor a = Chan a -> IO ()
 newtype PChan a = PChan { sendPChan :: a -> IO () }
 
 sendOnly :: Chan a -> PChan a
-sendOnly chan = PChan $ writeChan chan
+sendOnly chan = PChan $ \ !x -> writeChan chan x
```
apaszke, line 35: Is this shallow forcing sufficient? I imagine this is connected to our recent discussion about sending unevaluated thunks to other threads, right?

src/lib/Live/Eval.hs
```diff
@@ -35,17 +39,25 @@ watchAndEvalFile :: FilePath -> EvalConfig -> TopStateEx
...
-type DriverState = (WithId TopStateEx, CacheState)
-type DriverM = ReaderT DriverCfg (StateT DriverState IO)
+-- The evaluation-in-progress state is
+-- - The (identified) current top-level environment
+-- - The ThreadId of the currently running worker (if any)
+-- - The list of blocks that remain to be evaluated (if any)
+-- If the worker exists, it is working on evaluating the first element
+-- of the blocks list in the environment given by the TopStateEx.
+-- This is consistent at entry and exit from handling each message,
+-- but may be briefly inconsistent mid-handler.
```
apaszke, line 58: This consistency comment makes me worried. No other thread can notice this, right? The source evaluation state is entirely managed by the coordinator thread?

src/lib/Live/Eval.hs
```diff
@@ -56,46 +68,183 @@ data CacheState = CacheState
...
+  (SourceEvalState curState curWorker curPending) <- gets fst
+  case curWorker of
+    (Just _) -> return ()
+    Nothing -> do
+      case curPending of
+        [] -> return ()
+        (next:_) -> do
+          opts <- asks fst
+          self <- lift ask
+          tid <- liftIO $ forkWorker opts curState next self
```
apaszke, line 177: Do we understand the cost of `liftIO`? We always have at most one worker, so if it's expensive we might want to make it persistent

apaszke (commented):
I'm happy to review further, but I'm having trouble following the structure of this code (not your fault, I never understood it properly).

src/lib/Actor.hs
axch: I doubt it's really sufficient; I put it in there defensively.  In manual testing, it currently doesn't appear that evaluation work leaks into the driver thread, but this registers to me as far from bullet-proof.  Time to add `NFData` and use `deepSeq`?

src/lib/Live/Eval.hs
axch: Correct.  This is the actor model -- the only cross-thread information passing is by explicitly sending messages on explicit channels.  The state type has no channels in it, so no other thread can affect it.  The driver also does not send its state in messages, so no other thread can detect the transient inconsistency.

src/lib/Live/Eval.hs
axch: I assume you're concerned about the underlying `forkIO`?  AFAIK that's supposed to be cheap, because it creates a "lightweight thread" managed by the Haskell runtime (as opposed to `forkOS`, which actually forks a new OS-level thread).  The `liftIO` itself is inevitable, because we have to either fork a new worker or send a message to the persistent one, the latter of which is also an IO action.  It's also lifting `IO` into our `DriverM` stack, which shouldn't be bad?

src/lib/Err.hs
```diff
@@ -240,9 +240,10 @@ addSrcTextContext offset text m =
 
 catchIOExcept :: MonadIO m => IO a -> m (Except a)
 catchIOExcept m = liftIO $ (liftM Success m) `catches`
-  [ Handler \(e::Errs)          -> return $ Failure e
-  , Handler \(e::IOError)       -> return $ Failure $ Errs [Err DataIOErr   mempty $ show e]
-  , Handler \(e::SomeException) -> return $ Failure $ Errs [Err CompilerErr mempty $ show e]
+  [ Handler \(e::Errs)           -> return $ Failure e
+  , Handler \(e::IOError)        -> return $ Failure $ Errs [Err DataIOErr   mempty $ show e]
+  , Handler \(e::AsyncException) -> liftIO $ throwIO e
```
apaszke, line 245: Add a comment?

src/lib/Live/Eval.hs
```diff
@@ -6,46 +6,87 @@
...
+-- subscribe by sending a write-only view of its input channel.
 watchAndEvalFile :: FilePath -> EvalConfig -> TopStateEx
                  -> IO (PChan (PChan RFragment))
 watchAndEvalFile fname opts env = do
   (_, resultsChan) <- spawn logServer
   let cfg = (opts, subChan Publish resultsChan)
-  (_, sourceChan) <- spawn $ runDriver cfg env
-  forkWatchFile fname sourceChan
+  (_, driverChan) <- spawn $ runDriver cfg env
+  forkWatchFile fname $ subChan FileChanged driverChan
```
apaszke, line 58: This `subChan` business is really awkward. We should really make channels functors, because that's what it represents I think? No need to do it in this PR, but just a thought I had reading this

src/lib/Live/Eval.hs
apaszke: Ok then I am a little confused about the "mid-handler" comment. Can you clarify what the "handler" is?

src/lib/Live/Eval.hs
```diff
@@ -56,77 +97,207 @@ data CacheState = CacheState
...
+  askResultsOutput :: m (PChan RFragment)
+  askSelf :: m (PChan DriverEvent)
+  getTopState :: m (WithId TopStateEx)
+  putTopState :: WithId TopStateEx -> m ()
+  -- Resets the evaluation state to initial, from the given TopStateEx.
+  -- Returns the old top state and the old worker spec, for reuse
+  refresh :: TopStateEx -> m (WithId TopStateEx, WorkerSpec)
+  -- Get the work chunk we are waiting for, if any
+  getWorkingBlock :: m (Maybe (WithId SourceBlock))
+  -- Run the action if there is no worker
```
apaszke, line 111: And what if there is a worker?

src/lib/Live/Eval.hs
```diff
@@ -56,77 +97,207 @@ data CacheState = CacheState
...
+-- kill the worker when that happens.
+
+-- The worker communicates with the driver by sending a "work
+-- complete" message.  Note that a worker due to be killed may send a
+-- "work complete" message before the driver actually kills it.  If a
+-- "file changed" message arrived in the interim, the TopState the
+-- worker delivers remains valid to enter into the cache, but should
+-- not change the driver's then-current TopState.
+
+-- For this reason, the WorkComplete message contains the ids of the  with
```
apaszke, line 148: some misspelling?

src/lib/Live/Eval.hs
```diff
@@ -56,77 +97,207 @@ data CacheState = CacheState
...
+  -- actually kill it.  We record that result in case the user edits
+  -- back to a state where it can be shown.
+  newState <- recordTruth oldState block answer
+  curState <- getTopState
+  waitingFor <- getWorkingBlock
+  when (oldState == curState
+        && (fmap withoutId waitingFor == Just (withoutId block))) $ do
+    -- We only update our working state if this message is, in fact,
+    -- from the worker we are currently waiting for.
+    rotateWorkingState newState
```
apaszke, line 240: I feel like there is an easier way to determine whether the message comes from our current worker: just ask the worker to send back a unique integer once it's done, and remember the integer the current worker is supposed to send you back. You might still get those out-of-band messages, but the check is pretty trivial then.

I think that would be an improvement, but I don't feel strongly enough to block this PR on it. It's still an improvement. So I'll leave the decision up to you.

src/lib/Live/Eval.hs
apaszke: Also, how about renaming this to `whenAbsentWorker`? That's more consistent with `Control.Monad.when`

apaszke (approved):
Looks great!

src/lib/Err.hs
axch: Done.

src/lib/Live/Eval.hs
axch: They're not `Functor`s because they're contravariant: `subChan :: (a -> b) -> (PChan b) -> (PChan a)`.  But there is probably a class for this pattern.

src/lib/Live/Eval.hs
axch: Done.

src/lib/Live/Eval.hs
axch: Thanks!  Fixed.

src/lib/Live/Eval.hs
axch: Yes, that would probably work, but I was going for a more semantic check here.

PR #803, opened 2022-03-22
Distinguish tables and functions with separate constructors rather than arrows.

dougalm:
We end up treating them quite differently anyway so it's a reasonable thing to
do. The reason to do it now is that I want to put the `Ix` dictionary in the
table type itself rather than have it be a phantom constraint and re-synthesize
it when we need it.

It also makes the (Haskell) types a bit more precise because we syntactically
rule out table types with effects.


src/lib/Simplify.hs
```diff
@@ -228,21 +232,43 @@ simplifyApp f xs = case fromNaryLam (NE.length xs) f of
...
+    ACase e alts ty -> do
+      -- TODO: Don't rebuild the alts here! Factor out Case simplification
+      -- with lazy substitution and call it from here!
+      resultTy <- getTabAppType ty $ toList xs
+      alts' <- forM alts \(Abs bs a) -> do
+        buildAlt (EmptyAbs bs) \vs -> do
+          a' <- applySubst (bs@@>vs) a
+          naryTabApp a' (map sink $ toList xs)
+      dropSubst $ simplifyExpr $ Case e alts' resultTy Pure
+    _ -> naryTabApp f $ toList xs
```
apaszke, line 254: I've done a bunch of changes to `simplifyApp` that have unfortunately made it a bit bigger, so I'd really like to fold those two into the same function 😕

apaszke (approved):
LGTM, but I'm not loving the code duplication. It seems like in most places the only difference is that we use a slightly different pattern to unpack types (`TabPi` vs `Pi`) and slightly different constructors. I think that they all could be factors into common helpers, possibly with type class instances to get specialization to fast code.

dougalm:
I agree that the duplication isn't great. But the two paths will diverge further when we add Ix instances, which might undo any de-duplication efforts. I'll merge this as-is and then try some de-duplication after I do the Ix change.

PR #805, opened 2022-03-24
Haskell extensions cleanup

axch:


src/lib/Syntax.hs
```diff
@@ -4,19 +4,8 @@
...
 {-# LANGUAGE StrictData #-}
-{-# LANGUAGE DeriveFunctor #-}
-{-# LANGUAGE ViewPatterns #-}
-{-# LANGUAGE TypeFamilies #-}
-{-# LANGUAGE InstanceSigs #-}
 {-# LANGUAGE UndecidableInstances #-}
-{-# LANGUAGE StandaloneDeriving #-}
-{-# LANGUAGE DerivingStrategies #-}
-{-# LANGUAGE DerivingVia #-}
-{-# LANGUAGE DefaultSignatures #-}
```
apaszke, line 19: Could you please revert this change? I'm splitting Syntax into multiple smaller files and it's going to conflict pretty badly. I'll do the cleanup you did here in my patch if that's ok.

apaszke (approved):
LGTM, but please don't touch Syntax!

src/lib/Syntax.hs
axch: Done.

PR #813, opened 2022-03-30
Rename top-level names in the Dex corpus to snake_case

### LINKED ISSUE #786: Naming convention in the standard library and examples
Status quo is `camelCase`.  How about moving to `snake_case` for functions and variables (but leaving types and constructors `camelCase`)?  It seems less annoying, and more congruent with Python and Julia.

axch:
Committed one by one so that they're easy to check individually if desired.

`make tests` passes at the end, though I haven't done other testing.

This also touches a couple of places where the compiler looks for or emits a specific Dex name, but I intentionally left %-primitives and foreign functions alone, because changing those would have to be coordinated across programming languages.

Changing local names is out of scope for this change---the footprint is large enough as it is.  Local names should also be easier to update independently, because by construction any such change need not be coordinated across files.

Fixes #786.

### DIFF
--- a/benchmarks/microbench.dx
+++ b/benchmarks/microbench.dx
@@ -4,7 +4,7 @@ warmup = for i:(Fin 100). ordinal i + ordinal i
 
 def sum_bench (n:Int) (k:Int) : Fin n => Float =
   xs = rand_vec n randn (new_key k)
-  for i. snd $ withState  0.0 \ref. for j. ref := get ref + xs.i + xs.j
+  for i. snd $ with_state  0.0 \ref. for j. ref := get ref + xs.i + xs.j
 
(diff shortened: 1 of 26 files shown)


axch:
This PR is completely programmatically generated, so I can easily accommodate requests to change a name to something other than what I chose here.  Some items for consideration are mentioned on #786.

axch:
Ah, I must cover jax2dex as well.  Of course.  Will regenerate and reopen.

apaszke:
I don't think there's a lot of value in having a separate commit per rename. Perhaps let's just squish it all together? Also, are you using a script to rename those? Maybe instead of reading the eventual diff it would be easier to review the script used to generate the diff?

PR #819, opened 2022-03-31
Make InplaceT more efficient by threading out updated bindings

apaszke:
InplaceT is a reader in bindings/environment and a writer in decls. When
two monadic actions are sequenced, the environment that is a result of
the first one can be computed by applying `extendOutMap` to the input
env and the decls of the first action. However, if that first action is
a long chain of binds, it might have already materialized the extended
env inside only to throw it out! This changes InplaceT so that every
action returns both the decls and an environment obtained from extending
the input env by the emitted decls.

In terms of performance, the benchmarks are a bit noisy but it does seem
to make a difference of ~5% end-to-end in run time, but it also lets us
save on 8% of all allocations during a single run!


dougalm (approved):
Looks great!

PR #825, opened 2022-04-02
Add more non-overlapping Dex compilation modes.

axch:
We already had two development make targets: one for the default build, and one for turning on the Haskell profiler.  This change renames the "Haskell profiler" build to "dbg", and adds two more: an "opt" build for building a development version of Dex with GHC optimizations on, and a new "prof" build tuned for profiling the Dex compiler.

We keep the builds from stepping on each other's compilation caches by keeping them in different .stack-work* directories.  This should make it relatively painless to switch between them.

It may be necessary to delete your ~/.stack/snapshots/ directory once after syncing past this change (axch@ was getting weird linking errors from stack until doing that).

### DIFF
--- a/makefile
+++ b/makefile
@@ -1,3 +1,52 @@
+# Main targets of interest for development
+#
+#   make tests             Run the full test suite in a clean build
+#   make run-tests/<file>  Run one test file from tests/
+#   make build             Build Dex (fastest complete compile time)
+#   make watch             Build Dex continuously as source is edited
+#   make check-watch       Just typecheck Dex continuously (fastest feedback)
+#   make build-opt         Slower, but gives faster Dex compiles
+#   make build-dbg         Slower, but better Haskell errors
+#   make build-prof        Optimized for profiling the speed of the Dex compiler
+#
+# We suggest adding the following or equivalent aliases:
+#
+#   alias dex="stack exec dex -- --lib-path lib"
+#   alias dexopt="stack exec --work-dir .stack-work-opt dex -- --lib-path lib"
+#   alias dexdbg="stack exec --work-dir .stack-work-dbg --trace dex -- --lib-path lib"
+#   alias dexprof="stack exec --work-dir .stack-work-prof --trace dex -- --lib-path lib"
+
+# In more detail:
+#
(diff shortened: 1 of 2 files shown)


makefile
```diff
@@ -21,10 +65,17 @@ else
...
+	DBG  := --work-dir .stack-work-dbg --trace
+	PROF := --work-dir .stack-work-prof --trace \
+	  --ghc-options="-fno-prof-auto" \
+          --ghc-options="-ddump-simpl" \
+          --ghc-options="-ddump-stg" \
+	  --ghc-options="-dsuppress-all" \
+          --ghc-options="-dno-suppress-type-signatures" \
+          --ghc-options="-ddump-to-file"
+
+	dex := $(STACK) exec dex --
```
apaszke, line 78: Should we perhaps delete the DEX_DUMP_GHC-based workflow?

makefile
```diff
@@ -166,9 +223,9 @@ run-tests/%: tests/%.dx just-build
 run-examples/%: examples/%.dx just-build
 	misc/check-quine $< $(dex) script --allow-errors
 
-# Run these with profiling on while they're catching lots of crashes
+# Run these with debugging on while they're catching lots of crashes
 prop-tests: cbits/libdex.so
-	$(STACK) test $(PROF)
+	$(STACK) test $(DBG)
```
apaszke, line 228: I think this target can be safely deleted

apaszke (approved):
Cool intro! One thing that seems important to mention however, is that tests are always run using the unoptimized Dex without any debugging aid. I've been long wondering if we haven't outgrown Make as a way to metaprogram the limitations of stack...

Another interesting idea would be to have a target that runs GHC in a typecheck-only mode for really fast development cycles (`-fno-code` IIRC).

makefile
axch: Done.

makefile
axch: Done.

axch:
That's a neat idea with -fno-code, but doesn't work cleanly: There is nothing to cache from compiling dependent modules (apparently) so `stack` keeps re-compiling them.  Maybe with more investigation we could come up with something.

PR #826, opened 2022-04-04
Get superclasses from the given dict's type rather than trying all projections

dougalm:
We were doing a very silly thing in dictionary synthesis. When we first
implemented type classes we wrote superclass projections by hand. To obtain the
superclasses for a particular given, we would try to apply every superclass
projection available, almost all of which would fail to type check. Maybe that
made sense at the time, but it's completely ridiculous now. A dictionary's
superclasses are now plainly described in the dictionary's type.

This speeds up prelude evaluation by 25%.

I had to special-case the `AccumMonoid` dictionary because we still define that
one by hand, so its representation isn't like the others.

### DIFF
--- a/src/lib/Inference.hs
+++ b/src/lib/Inference.hs
@@ -86,7 +86,6 @@ inferTopUDecl (UInterface paramBs superclasses methodTys className methodNames) 
   dictDefName <- emitDataDef dictDef
   let classDef = ClassDef classSourceName methodSourceNames dictDefName
   className' <- emitClassDef classDef =<< tyConDefAsAtom dictDefName Nothing
-  mapM_ (emitSuperclass className') [0..(length superclasses - 1)]
   methodNames' <-
     forM (enumerate $ zip methodSourceNames methodTys) \(i, (prettyName, ty)) -> do
       let UMethodType (Right explicits) _ = ty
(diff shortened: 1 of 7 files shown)


src/lib/Inference.hs
```diff
@@ -2261,41 +2247,32 @@ getSuperclassClosure projs givens newGivens = do
...
-    resultTy' <- applyAbs (Abs b resultTy) param
-    params <- instantiateProjParamsM resultTy' dictTy
-    return $ Var param : params
-  Pi (PiType (PiBinder _ argTy PlainArrow) _ _) -> do
-    unify dictTy argTy
-    return []
-  _ -> error $ "unexpected projection type: " ++ pprint projTy
+      return $ getProjection [i] $ getProjection [0, 0] dictAtom
+
+  -- TypeCon sourceName dataDefName _ <- getType dict
```
apaszke, line 2258: Leftover?

src/lib/Inference.hs
```diff
@@ -2261,41 +2247,32 @@ getSuperclassClosure projs givens newGivens = do
...
+getNumSuperclasses :: EnvReader m => Given n -> m n Int
```
apaszke, line 2259: Perhaps instead of counting only to iterate later, you could return the list of superclass types that you then could iterate over.

apaszke (approved):
Amazing!

src/lib/Inference.hs
dougalm: yeah, that's better

PR #830, opened 2022-04-05
Dump GHC Core and STG even when building in PROF mode

apaszke:
This shouldn't cost much compared to actual compilation and is very
helpful in diagnosing performance problems in real optimized
executables.


axch (commented):
LGTM.

PR #834, opened 2022-04-06
Replace local instances with explicit dictionary projection/creation/application

dougalm:
We only used local instances in `run_accum`, where we're playing tricky rank-2
games with `AccumMonoid`. We can do the same thing with some privileged
primitives for working with dictionaries directly.

### DIFF
--- a/src/lib/Types/Primitives.hs
+++ b/src/lib/Types/Primitives.hs
@@ -140,6 +140,10 @@ data PrimOp e =
       -- Pointer to the stdout-like output stream
       | OutputStreamPtr
       | SynthesizeDict SrcPosCtx e  -- Only used during type inference
+      | ProjMethod e e  -- project a method, by source name, from the dict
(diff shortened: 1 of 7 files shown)


src/lib/Types/Primitives.hs
```diff
@@ -140,6 +140,10 @@ data PrimOp e =
       -- Pointer to the stdout-like output stream
       | OutputStreamPtr
       | SynthesizeDict SrcPosCtx e  -- Only used during type inference
+      | ProjMethod e e  -- project a method, by source name, from the dict
```
apaszke, line 143: There's no source name here, so the comment is a bit confusing. Perhaps clarify that you want a `Label` in the first argument?

apaszke (approved):
Looks great!

apaszke:
This is necessary for some changes I want to make next, so I figured I'll merge it myself. Hope that's ok!

PR #837, opened 2022-04-06
Optimize CheapReduce

apaszke:
With the classic bag of tricks already applied in simplification and
Imp. Not a huge deal, but still a few % end to end performance improvement
and 5% less allocations overall.


axch (approved):
I don't understand what `refreshAbs` is supposed to do, so I didn't review that change, but everything else LGTM.

PR #838, opened 2022-04-06
Use a compact encoding of strings into Names

apaszke:
Now, this makes quite a difference! A 10% end to end speedup! 10% fewer
overall allocations! _25% fewer bytes copied by GC_! It wasn't for free,
since we do loose the ability to represent longer strings in names and
the implementation requires quite a bit of bit manupulation, but seems
to be worth it!

Note that this is based on the initial implementation and ideas of
@dougalm.


tests/unit/RawNameSpec.hs
```diff
@@ -0,0 +1,62 @@
...
+        let n = 512
+        let step = \(m, ok) h ->
+                     let n = traceShowId $ R.freshRawName hint m in
+                     (R.singleton n () <> m, ok && not (n `R.member` m))
+        snd $ foldl step (initM, True) (replicate n hint)
+
+    it "string names are in bijection with short strings" do
+      property \(StringNameHint s) -> do
+        let s' = show (R.rawNameFromHint (R.getNameHint s))
+        counterexample s' $ ("s" ++ s ++ "0") == s'
```
apaszke, line 62: If you can come up with more tests, do let me know! I'd really like to make sure this implementation is correct.

dougalm (approved):
This is excellent

PR #839, opened 2022-04-07
Fix #833 

### LINKED ISSUE #833: Simplification fails on a table containing functions whose type mentions the table index
Reproduction:
```
for i:(Fin 2). \(x:Float). for j:(..i). 1
```

Expected: a value of type `((i:(Fin 2)) => (Float -> (..i) => Int32))`.

Got:
```
> Compiler bug!
>
> Leaked local variables:[v1]
```

This happens when trying to differentiate through the LU decomposition, which uses triangular matrices internally.

axch:
The root of the problem is that Dex used to capture lambda-bound
variables during defunctionalization; but that's actually too
conservative, because the argument (unlike anything from the enclosing
environment) will actually be available when the lambda is
reconstructed.

This change avoids doing that, so the return type of the lambda need
no longer be hoisted past its binder, and thereby doesn't trigger the
observed leak error.

Co-authored-by: Dougal <d.maclaurin@gmail.com>

### DIFF
--- a/src/lib/Builder.hs
+++ b/src/lib/Builder.hs
@@ -57,7 +57,7 @@ import qualified Data.Set        as S
 import qualified Data.Map.Strict as M
 import Data.Functor ((<&>))
 import Data.Graph (graphFromEdges, topSort)
-import Data.Text.Prettyprint.Doc (Pretty (..))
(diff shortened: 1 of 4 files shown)


axch:
Looks like the CLA bot got confused by the co-authored-by line.  Trying again.

PR #840, opened 2022-04-07
Fix #833 with Dougal.

### LINKED ISSUE #833: Simplification fails on a table containing functions whose type mentions the table index

axch:
The root of the problem is that Dex used to capture lambda-bound
variables during defunctionalization; but that's actually too
conservative, because the argument (unlike anything from the enclosing
environment) will actually be available when the lambda is
reconstructed.

This change avoids doing that, so the return type of the lambda need
no longer be hoisted past its binder, and thereby doesn't trigger the
observed leak error.

Co-authored-by: @dougalm <d.maclaurin@gmail.com>

### DIFF
--- a/src/lib/Simplify.hs
+++ b/src/lib/Simplify.hs
@@ -75,7 +75,7 @@ data SimplifiedBlock n = SimplifiedBlock (Block n) (ReconstructAtom n)
 -- accessible as nullary functions)
 simplifyTopBlock :: EnvReader m => Block n -> m n (SimplifiedBlock n)
 simplifyTopBlock block = liftSimplifyM do
-  (Abs UnitB block', recon) <- simplifyAbs $ Abs UnitB block
+  (Abs UnitB block', Abs UnitB recon) <- simplifyAbs $ Abs UnitB block
   return $ SimplifiedBlock block' recon
 {-# SCC simplifyTopBlock #-}
 
@@ -294,7 +294,7 @@ simplifyAtom atom = confuseGHC >>= \_ -> case atom of
     -- TODO(subst): Use EnvReaderI to getType before subst
     substM atom >>= getType >>= isData >>= \case
       True -> do
-        (Abs b' body', IdentityRecon) <- simplifyAbs $ Abs b body
+        (Abs b' body', IdentityReconAbs) <- simplifyAbs $ Abs b body
         return $ TabLam $ TabLamExpr b' body'
       False -> substM atom
   -- We don't simplify body of lam because we'll beta-reduce it soon.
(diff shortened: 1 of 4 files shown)


axch:
Fixes #833.

dougalm (approved):

src/lib/Simplify.hs
```diff
@@ -523,53 +525,61 @@ simplifyOp op = case op of
...
+    (lam', Abs b recon) <- simplifyLam lam
     ans <- liftM Var $ emit $ Hof $ For d lam'
     case recon of
       IdentityRecon -> return ans
       LamRecon reconAbs ->
         buildTabLam "i" ixTy \i' -> do
           elt <- tabApp (sink ans) $ Var i'
-          applyReconAbs (sink reconAbs) elt
+          reconAbs' <- applySubst (b @> i') reconAbs
+          applyReconAbs reconAbs' elt
```
apaszke, line 543: It would be nice if we could avoid substituting the body of `recon` twice (once in `applySubst` and once in `applyReconAbs`). Perhaps we could make `applyReconAbs` run in a `SubstReader`?

src/lib/Simplify.hs
```diff
@@ -359,26 +359,31 @@ simplifyVar v = do
...
+      -> SimplifyM i o (Atom o, Abs LamBinder ReconstructAtom o)
     doSimpLam b body = do
       (Abs b' body', recon) <- simplifyAbs $ Abs b body
       return $! (Lam $ LamExpr b' body', recon)
 
-simplifyBinaryLam :: Emits o => Atom i -> SimplifyM i o (Atom o, ReconstructAtom o)
+type BinaryLamBinder = (PairB LamBinder LamBinder)
+
+simplifyBinaryLam :: Emits o => Atom i
+  -> SimplifyM i o (Atom o, Abs (PairB LamBinder LamBinder) ReconstructAtom o)
```
apaszke, line 378: nit: use `BinaryLamBInder` here?

apaszke (approved):

src/lib/Simplify.hs
axch: Done.

src/lib/Simplify.hs
axch: Left a todo note.  I don't know how to do this; want to pair it some time?

Issue #841, opened 2022-04-07
Linearization fails on a function with a local triangular table


axch:
After PR #840 fixing Issue #833, we still have
```
func = \x:Float.
  table = for i:(Fin 2). for j:(..i). x
  tmp = (0 @ (Fin 2))
  sum table.tmp

snd (linearize func 1.0) 2.0
> Compiler bug!
> Please report this at github.com/google-research/dex-lang/issues
>
> Leaked local variables:[v1]
```

Expected behavior is to differentiate this cleanly.

axch:
This comes up when linearizing the LU decomposition in `linalg.dx`.

axch:
Fixed by PR 843 (apparently the closing comment needs to be in the PR description itself?)

PR #843, opened 2022-04-08
Fix bug 841, linearization of triangular table.

axch:
Defunctionalization is supposed to only capture local variables.  We
used to also be capturing all variables in the types of those local
variables, but we should only capture _local_ variables in the types
of local variables.

Not sure why exactly this come up during linearization of a function
with a local triangular table (such as the LU decomposition), but it
did.

Co-authored-by: @dougalm <dougalm@google.com>


axch:
Fixes #841.

dougalm (approved):

PR #844, opened 2022-04-08
Add another order of analytic sums to Algebra.hs.

### LINKED ISSUE #842: Linearization fails on a function with a pyramidal loop
The LU decomposition function has a loop of the form
```
for j.
  for i:(..j).
    foo = sum for k:(..i).
      ...
```
After patching Issue #833 and Issue #841, linearizing this hits
```
> Compiler bug!
>
> Not implemented yet!
>   error, called at src/lib/Algebra.hs:105:14 in dex-0.1.0.0
```
presumably because somebody tries to analytically figure out the size of some intermediate in that linearization.

axch:
Also improve the error message, since it is (kind of) user-facing.

Fixes #842.

Co-authored-by: @dougalm <dougalm@google.com>

### DIFF
--- a/src/lib/Algebra.hs
+++ b/src/lib/Algebra.hs
@@ -99,10 +99,13 @@ sumMonoC lim (Abs b (ClampMonomial clamps m)) =
 sumMono :: PolyName n -> Abs PolyBinder Monomial n -> Polynomial n
 sumMono lim (Abs b (Monomial m)) = case lookup (binderName b) m of
   -- TODO: Implement the formula for arbitrary order polynomials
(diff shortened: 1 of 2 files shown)


apaszke (approved):

apaszke:
I'm kind of surprised by this! I didn't expect linearization to increase the order of triangular tables

axch:
I don't think it does increase their order -- but the LU decomposition has a three-deep triangular loop, which I assume causes AD to materialize an intermediate of that shape (that isn't necessary in the primal program).  Needless to say, I did not delve carefully into what exactly happens, but just observed a crash that this change averts.

PR #845, opened 2022-04-08
Add a reverse nest (snoc nest) for linear-time decl accumulation

apaszke:
Back when we used the classic implementation of Writer for InplaceT this
didn't make a big difference, but now that the extensions are generally
always on the right, it does! With this change GC copies 12% fewer bytes
over a single run of the kernel regression example, leading to a ~5%
end-to-end speedup!

### DIFF
--- a/src/lib/Core.hs
+++ b/src/lib/Core.hs
@@ -17,6 +17,7 @@
 {-# LANGUAGE DerivingStrategies #-}
 {-# LANGUAGE DerivingVia #-}
 {-# LANGUAGE DefaultSignatures #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
 
 module Core where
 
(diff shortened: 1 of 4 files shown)


src/lib/Name.hs
```diff
@@ -323,6 +330,18 @@ data Nest (binder::B) (n::S) (l::S) where
   Nest  :: binder n h -> Nest binder h l -> Nest binder n l
   Empty ::                                  Nest binder n n
 
+data RNest (binder::B) (n::S) (l::S) where
+  RNest  :: RNest binder n h -> binder h l -> RNest binder n l
+  REmpty ::                                   RNest binder n n
+
+unrevNest :: RNest b n l -> Nest b n l
```
axch, line 337: I find the name `unrevNest` unintuitive.  Maybe call this `reverseNest` (though that's ambiguous with going from `Nest` to `RNest`), or `unRNest` to evoke the type more precisely?  Or `rNestToNest`?

src/lib/Name.hs
```diff
@@ -806,6 +825,18 @@ doJoinNest l r = case l of
   Empty     -> r
   Nest b lt -> Nest b $ doJoinNest lt r
 
+--joinRNest l r = trace (show (nestLength $ unrevNest l) ++ " " ++ show (nestLength $ unrevNest r)) $ joinRNest' l r
```
axch, line 828: Stale debug print?

src/lib/Core.hs
```diff
@@ -324,6 +325,10 @@ nestToEnvFrag = nestToEnvFragRec emptyOutFrag
 instance BindsEnv b => (BindsEnv (NonEmptyNest b)) where
   toEnvFrag (NonEmptyNest b rest) = toEnvFrag $ Nest b rest
 
+instance BindsEnv b => (BindsEnv (RNest b)) where
```
axch, line 328: Why is this instance so much simpler than the corresponding instance for Nest?

axch (approved):

src/lib/Name.hs
apaszke: Renamed to `unRNest`.

src/lib/Name.hs
apaszke: Good catch, thanks!

src/lib/Core.hs
apaszke: I rebased an older commit that I wrote before I optimized some Nest functions. But some of those Nest optimizations are not necessary anymore, thanks to making InplaceT be a state monad in decls, not a writer (implementation-wise). I ported the ones that might still be useful to the RNest case.

PR #847, opened 2022-04-08
Make EitherE6 faster

apaszke:
We use it all over the place in GenericE instances, but instead of being
a flat ADT it's actually a nest of binary sum types! To inject the last
case into the generic representation one has to allocate 6 extra
constructors, and that obviously adds up. Especially given that GHC is
currently unwilling to optimize out toE/fromE. And even when I try hard
to get it to simplify those away, GHC becomes very shy about inlining
things into very deep case statements.

So, overall it's a win all accross the board. If we simplify away
GenericE our code will be faster. And if we don't... it will still be
faster! With this change, we allocate On the kernel regression example we
allocate 15% less data overall, letting us save on 15% of gen-0 garbage
collections. It doesn't seem to translate to huge performance
improvements, but it does seem to let us get a percent or two
end-to-end.

### DIFF
--- a/src/lib/Name.hs
+++ b/src/lib/Name.hs
@@ -52,10 +52,7 @@ module Name (
   substM, ScopedSubstReader, runScopedSubstReader,
   HasNameHint (..), NameHint (..), Color (..),
   GenericE (..), GenericB (..),
-  EitherE1, EitherE2, EitherE3, EitherE4, EitherE5, EitherE6,
-    pattern Case0, pattern Case1, pattern Case2, pattern Case3, pattern Case4, pattern Case5,
-  EitherB1, EitherB2, EitherB3, EitherB4, EitherB5,
-    pattern CaseB0, pattern CaseB1, pattern CaseB2, pattern CaseB3, pattern CaseB4,
+  EitherE2, EitherE3, EitherE4, EitherE5, EitherE6 (..),
   splitNestAt, joinNest, nestLength, nestToList, binderAnn,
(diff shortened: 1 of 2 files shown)


src/lib/Name.hs
```diff
@@ -2121,55 +2118,81 @@ instance Store (Scope n)
...
-
-pattern CaseB1 :: e1 n l -> EB e0 (EB e1 rest) n l
-pattern CaseB1 e = RightB (LeftB e)
-
-pattern CaseB2 :: e2 n l -> EB e0 (EB e1 (EB e2 rest)) n l
-pattern CaseB2 e = RightB (RightB (LeftB e))
-
-pattern CaseB3 :: e3 n l -> EB e0 (EB e1 (EB e2 (EB e3 rest))) n l
-pattern CaseB3 e = RightB (RightB (RightB (LeftB e)))
+data EitherE6 (e0::E) (e1::E) (e2::E) (e3::E) (e4::E) (e5::E) (n::S)
```
axch, line 2121: Some (all?) of the explanation you gave for this change in your PR message would go well here as a code comment, explaining why this is 6 and not 5 or 7, and which it's a flat structure rather than a chain of pairs.

axch (approved):
Nice!  Just one request for an explanatory comment :)

src/lib/Name.hs
apaszke: Added a comment. Some of the details from PR description are more before vs after, and I skipped those. One can always find them through git history.

PR #868, opened 2022-04-15
Make a dedicated expression language for synthesized dictionaries

dougalm:


src/lib/Type.hs
```diff
@@ -98,8 +97,8 @@ sourceNameType v = do
       UAtomVar    v' -> getType $ Var v'
       UTyConVar   v' -> lookupEnv v' >>= \case TyConBinding _     e -> getType e
       UDataConVar v' -> lookupEnv v' >>= \case DataConBinding _ _ e -> getType e
-      UClassVar   v' -> lookupEnv v' >>= \case ClassBinding  _    e -> getType e
-      UMethodVar  v' -> lookupEnv v' >>= \case MethodBinding _ _  e -> getType e
+      UClassVar   v' -> lookupEnv v' >>= \case ClassBinding  _   -> undefined
```
apaszke, line 100: We should probably implement this?

src/lib/Type.hs
```diff
@@ -261,10 +260,7 @@ instance CheckableE SourceMap where
   checkE sourceMap = substM sourceMap
 
 instance CheckableE SynthCandidates where
-  checkE (SynthCandidates xs ys) =
-    SynthCandidates <$> mapM checkE xs
-                    <*> (M.fromList <$> forM (M.toList ys) \(k, vs) ->
-                          (,) <$> substM k <*> mapM checkE vs)
+  checkE scs = substM scs
```
apaszke, line 263: This looks like a regression

src/lib/Type.hs
```diff
@@ -368,6 +365,8 @@ instance HasType Atom where
       params' <- traverse checkE params
       void $ checkedApplyNaryAbs (Abs tyParamNest' (ListE absCons')) params'
       return TyKind
+    DictCon dictExpr -> getTypeE dictExpr
+    DictTy _ -> return TyKind  -- TODO: check params
```
apaszke, line 369: Can you please implement this?

src/lib/Type.hs
```diff
@@ -607,6 +628,10 @@ typeCheckPrimCon con = case con of
   ParIndexCon t v -> t|:TyKind >> v|:IdxRepTy >> substM t
   RecordRef xs -> (RawRefTy . StaticRecordTy) <$> traverse typeCheckRef xs
   LabelCon _   -> return $ TC $ LabelType
+  ExplicitDict dictTy _  -> do
+    -- TODO: check method
```
apaszke, line 632: Can you please implement that? CI is already picking up errors in the way we used explicit app, so those checks are important

src/lib/Simplify.hs
```diff
@@ -525,21 +527,28 @@ simplifyOp op = case op of
...
+projectDictMethod :: Emits o => Atom o -> Int -> SimplifyM i o (Atom o)
+projectDictMethod d i = do
+  cheapNormalize d >>= \case
+    DictCon (InstanceDict instanceName args) -> dropSubst do
+      args' <- mapM simplifyAtom args
+      InstanceDef _ bs _ body <- lookupInstanceDef instanceName
+      let InstanceBody _ methods = body
+      let method = methods !! i
+      extendSubst (bs@@>(SubstVal <$> args')) $
+        simplifyBlock method
```
apaszke, line 545: Can you please explain how is it that we end up with `InstanceDict`s in the end? Why couldn't it be a superclass projection?

apaszke (commented):
A partial review for everything, except inference. Mostly LGTM, but I don't understand the new convention and it could use a few code comments.

src/lib/Inference.hs
```diff
@@ -1982,8 +2020,14 @@ instance Unifiable Atom where
          unless (c == c') empty >> unifyFoldable xs xs'
        (TC con, TC con') -> unifyFoldable con con'
        (Eff eff, Eff eff') -> unify eff eff'
+       (DictTy d, DictTy d') -> unify d d'
        _ -> empty
 
+instance Unifiable DictType where
+  unifyZonked (DictType _ c params) (DictType _ c' params')
+    | c == c'   = unifyFoldable params params'
+    | otherwise = empty
```
apaszke, line 2029: 🏌️‍♂️ This can also be spelled as `guard (c == c') *> unifyFoldable params params'`

src/lib/Inference.hs
```diff
@@ -2156,30 +2200,42 @@ synthTopBlock block = do
...
-        (d, _):_ -> return d
-{-# SCC trySynthDictBlock #-}
+        (d, _):_ -> cheapNormalize d
+{-# SCC trySynthTerm #-}
 
-data Givens n = Givens { fromGivens :: HM.HashMap (EKey Type n) (Given n) }
+-- The type of a `SynthExpr` atom must be `DictTy _` or `Pi _`. If it's `Pi _`,
+-- then the atom itself is either a variable or a trivial lambda whose body
+-- is also a `SynthExpr`.
+type SynthExpr = Atom
```
apaszke, line 2230: Should `trySynthTerm` return a `SynthExpr`?

apaszke: Also, to avoid confusion, perhaps rename this to `SynthAtom`?

src/lib/Inference.hs
```diff
@@ -2156,30 +2200,42 @@ synthTopBlock block = do
...
+      synthTy <- typeAsSynthType ty
+      solutions <- liftSyntherM $ synthTerm synthTy
       let solutionsInstance = filter ((== UsedInstance) . snd) solutions
       unless (length solutionsInstance <= 1) $
         throw TypeErr $ "Multiple candidate class dictionaries for: " ++ pprint ty
       case solutions of
         [] -> throw TypeErr $ "Couldn't synthesize a class dictionary for: " ++ pprint ty
-        (d, _):_ -> return d
-{-# SCC trySynthDictBlock #-}
+        (d, _):_ -> cheapNormalize d
```
apaszke, line 2224: Why do we need to cheap normalize here?

src/lib/Inference.hs
```diff
@@ -2156,30 +2200,42 @@ synthTopBlock block = do
...
+-- The type of a `SynthExpr` atom must be `DictTy _` or `Pi _`. If it's `Pi _`,
+-- then the atom itself is either a variable or a trivial lambda whose body
+-- is also a `SynthExpr`.
+type SynthExpr = Atom
 
-type Given = Block
+data SynthType n =
+   SynthDictType (DictType n)
+ | SynthPiType (Abs PiBinder SynthType n)
+   deriving (Show, Generic)
```
apaszke, line 2235: What's the benefit of maintaining `SynthType`? It seems like we have to do conversions in quite a few places and we have to write whole functions that e.g. interpret an instance as a `SynthType`, even though we already wrote a function that creates a `Type` for an instance!

src/lib/Inference.hs
```diff
@@ -2248,102 +2313,109 @@ getSuperclassClosure givens newGivens = do
...
     markAsVisited :: EnvReader m => Given n -> StateT1 Givens m n ()
     markAsVisited x = do
-      ty <- getType x
+      ty <- getSynthType x
       modify \(Givens m) -> Givens $ HM.insert (EKey ty) x m
 
 getDirectSuperclasses :: EnvReader m => Given n -> m n [Given n]
-getDirectSuperclasses dict = do
-  numSuperclasses <- getNumSuperclasses dict
+getDirectSuperclasses synthExpr = do
```
apaszke, line 2325: You've defined a `SynthExpr` type above, but `synthExpr` here as a type of `Given`? I'm not sure how to interpret that

src/lib/Types/Core.hs
```diff
@@ -925,7 +945,8 @@ instance GenericE Atom where
         DataCon printName defName params con args
       Case3 (LiftE sourceName `PairE` defName `PairE` ListE params) ->
         TypeCon sourceName defName params
-      _ -> error "impossible"
+      Case4 d -> DictCon d
+      Case5 d -> DictTy  d
```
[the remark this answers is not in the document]
dougalm: Yes, definitely. But I'll need your help figuring out how to check whether it's still being inlined.

src/lib/Types/Primitives.hs
```diff
@@ -139,10 +141,7 @@ data PrimOp e =
       | SumToVariant e
       -- Pointer to the stdout-like output stream
       | OutputStreamPtr
-      | SynthesizeDict SrcPosCtx e  -- Only used during type inference
-      | ProjMethod e e  -- project a method, by source name, from the dict
-      -- Used in prelude for `run_accum`. Only works for single-method classes.
-      | ExplicitDict e e  -- dict type, dict method
+      | ProjMethod e Int  -- project a method from the dict
```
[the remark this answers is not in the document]
dougalm: Because method projection is now where work can happen! Previously, synthesizing a dictionary for `Add (n=>a)` would create an array of zeros for the `zero` method. Now, that only happens at method projection, when you actually projection out the `zero` method. This means we don't have to play tricks with thunking either at the user level or inside the compiler.

src/lib/Types/Core.hs
```diff
@@ -217,6 +217,46 @@ instance GenericB EffectBinder where
...
+
+data InstanceBody (n::S) =
+  InstanceBody
+    [Atom n]   -- superclasses dicts
+    [Block n]  -- method definitions
+  deriving (Show, Generic)
+
+data MethodType (n::S) =
+  MethodType
+    [Bool]    -- indicates explicit args
```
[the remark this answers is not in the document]
dougalm: That's true. But at the top level, source names for methods point to this piece of data. It's similar to how we keep the implicit-vs-explicit arrow types in the core IR, even though they aren't relevant after type inference. At least in this case the data only lives at the top level, so it's not like we're paying the cost of carrying the `[Bool]` in the expressions we're compiling.

src/lib/Types/Core.hs
```diff
@@ -217,6 +217,46 @@ instance GenericB EffectBinder where
...
+    (Type n)  -- actual method type
+ deriving (Show, Generic)
+
+data DictType (n::S) = DictType SourceName (ClassName n) [Type n]
+     deriving (Show, Generic)
+
+data DictExpr (n::S) =
+   InstanceDict (InstanceName n) [Atom n]
+ | InstantiatedGiven (Atom n) (NonEmpty (Atom n))
```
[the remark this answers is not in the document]
dougalm: > I feel like this needs more docs.
Yes, agreed

> Why can givens only be referred to when there's at least one argument to apply? It doesn't sound like a true property to me.
This was an attempt to normalize things a bit. You can have a given without arguments indeed, that's the common case! But you have to express it as the `Var` case of `Atom` rather than using a `DictExpr`. If we allowed givens without arguments in `DictExpr` then you could express the same atom as either `Var v` or `DictCon (Given (Var v) [])`. Now, maybe that's actually ok, because we don't make any promises that `DictExpr`s are normalized. wdyt?

src/lib/Types/Core.hs
```diff
@@ -217,6 +217,46 @@ instance GenericB EffectBinder where
...
+ | SuperclassProj (Atom n) Int  -- (could instantiate here too, but we don't need it for now)
+   deriving (Show, Generic)
```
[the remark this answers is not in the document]
dougalm: Yeah, the recursion goes through the atom. The `Atom` cases that get used in practice are `Var` and `Lam`. In my first attempt I had `DictExpr` be directly recursive as you suggest. But reimplementing the `Lam`-like case ended up duplicating things more than I liked. Another option is to make `DictExpr` syntactically normalized by disallowing the `Lam` case. But that means that substitution has to do beta reduction, and that seemed like a big change.

src/lib/Name.hs
```diff
@@ -738,7 +738,7 @@ applyAbs (Abs b body) x = applySubst (b@>x) body
 
 applyNaryAbs :: ( SinkableV v, FromName v, ScopeReader m, BindsNameList b c, SubstE v e
                 , SubstB v b, SinkableE e)
-             => Abs b e n -> [v c n] -> m n (e n)
+             => HasCallStack => Abs b e n -> [v c n] -> m n (e n)
```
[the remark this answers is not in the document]
dougalm: Yep, debugging leftovers. Thanks.

src/lib/CheapReduction.hs
```diff
@@ -170,13 +169,25 @@ cheapReduceName v =
 class CheaplyReducibleE (e::E) (e'::E) where
   cheapReduceE :: e i -> CheapReducerM i o (e' o)
 
+-- Used to avoid ambiguous vars
+cheapReduceToAtomE :: CheaplyReducibleE e Atom => e i -> CheapReducerM i o (Atom o)
+cheapReduceToAtomE = cheapReduceE
+{-# INLINE cheapReduceToAtomE #-}
```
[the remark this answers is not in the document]
dougalm: Nice. I'll do that.

src/lib/Types/Primitives.hs
```diff
@@ -79,6 +78,9 @@ data PrimCon e =
       | RecordRef (LabeledItems e)
       | ParIndexCon e e        -- Type, value
       | LabelCon String
+      -- Used in prelude for `run_accum`. Only works for single-method classes.
+      | ExplicitDict e e  -- dict type, dict method
+      | SynthesizeDict SrcPosCtx e  -- Only used during type inference
```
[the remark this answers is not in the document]
dougalm: Sure. Or how about `DictHole`, which we used to use?

src/lib/Inference.hs
dougalm: Good point. I added it because I needed help figuring out which types were valid arguments to `instantiateSynthArgs`. But it would probably be fine to just make `SynthType` an alias with a comment explaining that it has to be either `DictTy` or `Pi`.

src/lib/Inference.hs
dougalm: Good point. Maybe we should remove the `Given` alias and use `SynthExpr` exclusively. "Given" refers to the role that a particular `SynthExpr` plays rather than the set of values it can take.

src/lib/Simplify.hs
dougalm: If it were a superclass projection, then either you're projecting out of an instance, in which case you can immediately resolve that projection, or else you're projecting out of a var, but that's impossible because vars can't have `DictTy` types post-simplification.

src/lib/CheapReduction.hs
```diff
@@ -170,13 +169,25 @@ cheapReduceName v =
...
     -- we have a chance to apply tham. Also, recursive traversal of those bodies
     -- means that we will follow the full call chain, so it's really expensive!
     Lam _   -> substM a
+    Con (SynthesizeDict ctx ty') -> do
+      ty <- cheapReduceE ty'
+      runFallibleT1 (trySynthTerm ty) >>= \case
+        Success d -> return d
+        Failure _ -> do
+          reportSynthesisFail ty
+          return $ Con $ SynthesizeDict ctx ty
```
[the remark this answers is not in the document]
dougalm: Good idea. I can't see a reason that wouldn't work. I'll try it.

src/lib/CheapReduction.hs
```diff
@@ -186,11 +197,27 @@ instance CheaplyReducibleE Atom Atom where
     Variant ty l c p -> do
       ExtLabeledItemsE ty' <- substM $ ExtLabeledItemsE ty
       Variant ty' l c <$> cheapReduceE p
+    DictCon d -> cheapReduceE d
     -- Do recursive reduction via substitution
     _ -> do
       a' <- substM a
       dropSubst $ traverseNames cheapReduceName a'
 
+instance CheaplyReducibleE DictExpr Atom where
```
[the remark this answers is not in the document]
dougalm: One reason would be to potentially make the expression smaller. Assuming it's already cheap-reduced, then the pass shouldn't be any more expensive than the substitution you would have to do anyway. But you've got much better instincts for performance than me. I'll try it out.

src/lib/Inference.hs
```diff
@@ -1290,67 +1311,78 @@ checkULam (ULamExpr _ (UPatAnn p ann) body) piTy@(PiType (PiBinder _ _ arr) _ _)
...
-  -> m i o (Atom o)
-checkInstanceArgs Empty cont = cont
+  -> (forall o'. (EmitsInf o', Ext o o') =>  m i' o' (e o'))
+  -> m i o (Abs (Nest PiBinder) e o)
+checkInstanceArgs Empty cont = Abs Empty <$> cont
 checkInstanceArgs (Nest (UPatAnnArrow (UPatAnn p ann) arrow) rest) cont = do
   case arrow of
     ImplicitArrow -> return ()
     ClassArrow    -> return ()
+    PlainArrow    -> return ()
```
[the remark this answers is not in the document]
dougalm: I think it was for `AddMonoid` and `MulMonoid`, which used to be defined as functions wrapping local instances. Now that we don't have local instance anymore, we need to allow top-level instances to take explicit args.

src/lib/Inference.hs
```diff
@@ -1361,16 +1393,18 @@ introDictTys (h:t) m = buildPiInf (getNameHint @String "_autoq") ClassArrow h \_
...
   MethodBinding className' i _ <- substM v >>= lookupEnv
   when (className /= className') do
-    ClassBinding (ClassDef classSourceName _ _) _ <- lookupEnv className
+    ClassBinding (ClassDef classSourceName _ _ _ _) <- lookupEnv className
     throw TypeErr $ pprint sourceName ++ " is not a method of "
                  ++ pprint classSourceName
-  let methodTy = methodTys !! i
-  rhs' <- checkSigma rhs methodTy
+  let MethodType _ methodTy = methodTys !! i
+  rhs' <- buildBlockInf do
```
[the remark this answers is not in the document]
dougalm: It's a real change! Notice that `checkMethodDef` now just uses `EmitsInf o` instead of `EmitsBoth o`. Methods are now blocks, and the decls get evaluated at projection time instead of at instance definition time and dictionary synthesis time. (Similar to the "automatic thunking" we had discussed.) So defining an instance doesn't actually emit any decls anymore.

src/lib/Inference.hs
```diff
@@ -1290,67 +1311,78 @@ checkULam (ULamExpr _ (UPatAnn p ann) body) piTy@(PiType (PiBinder _ _ arr) _ _)
...
+    PlainArrow    -> return ()
     _ -> throw TypeErr $ "Not a valid arrow for an instance: " ++ pprint arrow
-  checkAnnWithMissingDicts ann \ds getArgTy -> do
+  ab <- checkAnnWithMissingDicts ann \ds getArgTy -> do
     introDicts (eSetToList ds) $ do
       argTy <- getArgTy
-      buildLamInf (getNameHint p) arrow argTy (const $ return Pure) \v -> do
-        bindLamPat p v $
+      buildPiAbsInf (getNameHint p) arrow argTy \v -> do
+        WithSrcB pos (UPatBinder b) <- return p  -- TODO: enforce this syntactically
```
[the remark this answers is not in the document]
dougalm: > Why? I don't see a reason why it has to be a plain binder. Can't we use bindLamPat like we used to?

The reason it has to be a plain binder is that other patterns would produce decls, and instances don't emit decls anymore.

> Oh wait. Why are we building up a nest of Pis and not lambdas?

The binders we're creating here are eventually going to end up in an `InstanceDef`, which we didn't have before. I used `PiBinder`s instead of `LamBinder`s for the arg binders in `InstanceDef` because they don't have effects, but it's true that they're more like lambda binders in spirit.

src/lib/Inference.hs
dougalm: Just to reduce code size, if we happen to have e.g. a superclass projection from a concrete instance.

src/lib/Types/Core.hs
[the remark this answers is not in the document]
apaszke: Do they have to point to this, instead of to a fully formed atom that is a lambda with a single `ProjectMethod` op inside? That would be closer to the old representation and it was nice in that we only had to elaborate the atom once, as opposed to recreating it multiple times every time it's used. WDYT?

src/lib/Types/Core.hs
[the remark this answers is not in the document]
apaszke: I think that would be ok, but perhaps this is fine as well, given that our `App`s are non-empty as well.

src/lib/Inference.hs
apaszke: I guess I was really just confused by the `synthExpr` name

src/lib/Name.hs
```diff
@@ -691,11 +691,11 @@ instance Color c => BindsOneName (BinderP c ann) c where
   {-# INLINE binderName #-}
 
 infixr 7 @@>
-(@@>) :: (Foldable f, BindsNameList b c) => b i i' -> f (v c o) -> SubstFrag v i i' o
+(@@>) :: HasCallStack => (Foldable f, BindsNameList b c) => b i i' -> f (v c o) -> SubstFrag v i i' o
 (@@>) bs xs = bindNameList bs (toList xs)
 
 class BindsNameList (b::B) (c::C) | b -> c where
-  bindNameList :: b i i' -> [v c o] -> SubstFrag v i i' o
+  bindNameList :: HasCallStack => b i i' -> [v c o] -> SubstFrag v i i' o
```
apaszke, line 698: More leftovers

src/lib/Name.hs
```diff
@@ -691,11 +691,11 @@ instance Color c => BindsOneName (BinderP c ann) c where
   {-# INLINE binderName #-}
 
 infixr 7 @@>
-(@@>) :: (Foldable f, BindsNameList b c) => b i i' -> f (v c o) -> SubstFrag v i i' o
+(@@>) :: HasCallStack => (Foldable f, BindsNameList b c) => b i i' -> f (v c o) -> SubstFrag v i i' o
```
apaszke, line 694: More leftovers

src/lib/Types/Primitives.hs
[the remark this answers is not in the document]
apaszke: `DictHole` sounds good as well. I'm ok with anything for as long as it doesn't include a verb

src/lib/Inference.hs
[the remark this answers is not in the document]
apaszke: Ohhh right, because you've made them named instances! Nice.

src/lib/Inference.hs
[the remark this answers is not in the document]
apaszke: > other patterns would produce decls

That's not necessarily true, is it? The main reason why `ProjectElt` exists is to allow us to refer to pieces of patterns without emitting decls.

src/lib/Inference.hs
apaszke: Would that really reduce code size? Wouldn't that inline the concrete `Atom` of the superclass?

src/lib/Inference.hs
apaszke: Yeah I think that would be a bit nicer and more consistent with what we do in other places.

src/lib/Types/Core.hs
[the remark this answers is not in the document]
dougalm: Method source names actually point to both. The `Binding` case looks like: `MethodBinding :: Name ClassNameC n -> Int -> Atom n -> Binding MethodNameC  n`. The atom there is the projection, so we don't have to elaborate it every time.

src/lib/Inference.hs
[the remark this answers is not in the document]
dougalm: Right, but some patterns produce decls, like table patterns.

src/lib/Inference.hs
dougalm: True, there's no guarantee that it'll be smaller, but I think in the common case it will be. The `DictExpr` for the superclass instance will be similar in size to the original instance, but you won't have the projection.

src/lib/Inference.hs
dougalm: On reflection, I think I prefer it the way it is currently. It's more precise and not much extra code. There's only one conversion, `typeAsSynthType`, and that's actually a meaningful function, not just boilerplate, because it checks whether the type is the sort of thing we know how to synthesize.

PR #870, opened 2022-04-18
Optimize SubstFrag values

apaszke:
Previously every element in a substitution or a scope was a list of
`WithColor` objects that existentially hid the color of the substitution
element. While it worked fine, it meant that putting a new value into scope
required us to allocate at least two extra constructors (`[]`+`WithColor`)
and every lookup required us to perform the extra pointer chasing to unpack
them all.

This change adds a `SubstItem` that we always use as a value in
`RawNameMap`s. It packs together an unboxed integer that encodes the
color of the `v` and whether the name for which this entry is present
has been shadowed in the scope fragment it represents.

Overall the effect on total allocations is not very dramatic (given how
unoptimized many paths still are), but it does result in a noisy, but
noticeable performance improvement.

### DIFF
--- a/src/lib/Name.hs
+++ b/src/lib/Name.hs
@@ -5,6 +5,8 @@
 -- https://developers.google.com/open-source/licenses/bsd
 
 {-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE AllowAmbiguousTypes #-}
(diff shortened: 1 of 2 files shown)


dougalm (approved):
Looks great!

PR #875, opened 2022-04-19
Don't inline the error message in from_ordinal

apaszke:
from_ordinal is used all over the place, but because it's polymorphic we
have no choice but to inline it into all use sites. (Well, we'd do that
even if it was monomorphic, but in principle that will change in the
future.) However, its body is large-ish, because it embeds sizable
string literals and inlines their concatenation code, which obviously
blows up the eventual code size.

This tiny change gives us a 15% end-to-end speedup and removes 27% of all
allocations on the kernel regression example! Seems like the changes
from the last month have shaken out most low hanging fruit from Haskell,
but we likely have plenty of those on the Dex side!


dougalm (approved):
Good trick! I didn't think of doing it that way.

PR #876, opened 2022-04-19
Added table index sets, and used them to make fft types enforce power…

duvenaud:
…s-of-2 sizes.

This was very satisfying.  Now we can programmatically iterate over tables of arbitrary dimension, which let me clean up the CTC and FFT examples.

The FFT inner loop also now avoids some unnecessary discrete math.  The FFT code now typechecks and passes tests again, thanks to [#877](https://github.com/google-research/dex-lang/pull/877)!

There are further enhancements that can be made to the FFT example once we can handle incrementing and decrementing `Nat`s at the type level.

### DIFF
--- a/examples/ctc.dx
+++ b/examples/ctc.dx
@@ -131,15 +131,13 @@ labels = for i:position. randIdxNoZero Vocab (new_key (ordinal i))
 -- semantics of the marginal likelihood being computed, and whether
 -- e.g. the summed-over labels should include blanks.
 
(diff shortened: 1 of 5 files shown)


duvenaud:
Okay, great, your PR unblocked me, and the FFT was already passing tests.  I added a few comments and added FFT back to the tests.  I think it's ready to merge now.

apaszke:
Thanks, David!

PR #880, opened 2022-04-27
Refactor Blocks to terminate in Atoms rather than Exprs

axch:
It turns out that this was already the case, except for the consequences of `inlineLastDecl`.  Moved that out of Builder and into
the two consumers that seem to depend upon it, namely PPrint.hs (of a chain of for expressions?) and Serialize.hs (which has a fast path guarded by pattern-matching for trivial string tables).

### DIFF
--- a/src/lib/Serialize.hs
+++ b/src/lib/Serialize.hs
@@ -63,13 +63,17 @@ getDexString (DataCon _ _ _ 0 [_, xs]) = case tryParseStringContent xs of
     tryParseStringContent tabAtom  = do
       TabLam (TabLamExpr i body) <- return tabAtom
       Fin (IdxRepVal n) <- return $ binderType i
-      Block _ (Nest castDecl (Nest offDecl Empty)) loadExpr <- return body
+      Block _ (Nest castDecl (Nest offDecl (Nest loadDecl Empty))) (Var result) <- return body
       Let v (DeclBinding _ _ (Op (CastOp IdxRepTy (Var i')))) <- return castDecl
       guard $ binderName i == i'
       Let v1 (DeclBinding _ _ (Op (PtrOffset (Var ptrName) (Var v')))) <- return offDecl
       guard $ binderName v == v'
+      Let r (DeclBinding _ _ loadExpr) <- return loadDecl
+      guard $ binderName r == result
       Hof (RunIO (Lam (LamExpr iob iobody))) <- return loadExpr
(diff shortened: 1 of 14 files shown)


src/lib/Serialize.hs
```diff
@@ -63,13 +63,15 @@ getDexString (DataCon _ _ _ 0 [_, xs]) = case tryParseStringContent xs of
     tryParseStringContent tabAtom  = do
       TabLam (TabLamExpr i body) <- return tabAtom
       Fin (IdxRepVal n) <- return $ binderType i
-      Block _ (Nest castDecl (Nest offDecl Empty)) loadExpr <- return body
+      Abs (Nest castDecl (Nest offDecl Empty)) loadExpr <-
+        return $ inlineLastDeclBlock body
```
apaszke, line 67: There's an easy edit to this code that won't require us to go through `inlineLastDeclBlock` can you please try this:
```hs
Block _ (Nest castDecl (Nest offDecl (Nest loadDecl))) resultAtom <- return body
```

Then, just add a check that `resultAtom` is a `Var` that's bound by `loadDecl`

src/lib/Serialize.hs
```diff
@@ -63,13 +63,15 @@ getDexString (DataCon _ _ _ 0 [_, xs]) = case tryParseStringContent xs of
...
       Let v (DeclBinding _ _ (Op (CastOp IdxRepTy (Var i')))) <- return castDecl
       guard $ binderName i == i'
       Let v1 (DeclBinding _ _ (Op (PtrOffset (Var ptrName) (Var v')))) <- return offDecl
       guard $ binderName v == v'
       Hof (RunIO (Lam (LamExpr iob iobody))) <- return loadExpr
-      HoistSuccess (Block _ Empty (Op (PtrLoad (Var v1')))) <- return $ hoist iob iobody
+      HoistSuccess (Abs Empty (Op (PtrLoad (Var v1')))) <-
+        return $ hoist iob $ inlineLastDeclBlock iobody
```
apaszke, line 74: Same here, you could just modify the pattern slightly.

apaszke (approved):
Looks good! I quite like ending `Block`s with `Atom`s. If you look at the (currently dead) `Parallelize` module, you'll see that I implemented that conversion there, because it was more convenient to have all effectful/expensive operations appear in decls, for uniformity.

src/lib/Serialize.hs
axch: Done.

src/lib/Serialize.hs
axch: Done.

PR #882, opened 2022-04-28
Track the effects actually used by a Block

axch:
Specifically:
- Add a field to BlockAnn to record the effects the Block uses
- Add `computeAbsEffects` that computes effects in the decls by traversing them
- Add code to Block-building operations to hoist those effects and stash the result in the new BlockAnn field

Knock-on changes:
- Cleaned up the implementation of `exprEffects` to remove the `MonadFail1` constraint.
- Inlined the monadic `makeBlock` into its one use site and rename `makeBlockOfType` to `makeBlock`.

I expect no functional changes from this, because the new effects annotation is not meaningfully used (that's for the next PR).

The code comments contain several questions for the reviewer, but at this point I expect the effects of the answers to be fairly local.

### DIFF
--- a/src/lib/Builder.hs
+++ b/src/lib/Builder.hs
@@ -10,7 +10,7 @@ module Builder (
   emit, emitOp, emitUnOp,
   buildPureLam, BuilderT (..), Builder (..), ScopableBuilder (..),
   Builder2, BuilderM, ScopableBuilder2,
-  liftBuilderT, buildBlock, app, add, mul, sub, neg, div',
+  liftBuilderT, buildBlock, withType, absToBlock, app, add, mul, sub, neg, div',
   iadd, imul, isub, idiv, ilt, ieq, irem,
   fpow, flog, fLitLike, buildPureNaryLam, emitMethod,
   select, getUnpacked, emitUnpacked,
@@ -35,12 +35,12 @@ module Builder (
   TopEnvFrag (..), emitPartialTopEnvFrag, emitLocalModuleEnv,
   fabricateEmitsEvidence, fabricateEmitsEvidenceM,
   singletonBinderNest, varsAsBinderNest, typesAsBinderNest,
-  liftBuilder, liftEmitBuilder, makeBlock, makeBlockOfType,
+  liftBuilder, liftEmitBuilder, makeBlock,
   indexToInt, indexSetSize, intToIndex,
   getIxImpl, IxImpl (..),
   litValToPointerlessAtom, emitPtrLit,
   liftMonoidEmpty, liftMonoidCombine,
-  telescopicCapture, telescopicCaptureBlock, unpackTelescope,
+  telescopicCapture, unpackTelescope,
   applyRecon, applyReconAbs, clampPositive,
   emitRunWriter, mCombine, emitRunState, emitRunReader, buildFor, unzipTab, buildForAnn,
   zeroAt, zeroLike, maybeTangentType, tangentType,
@@ -58,7 +58,7 @@ import qualified Data.Set        as S
 import qualified Data.Map.Strict as M
 import Data.Functor ((<&>))
 import Data.Graph (graphFromEdges, topSort)
-import Data.Text.Prettyprint.Doc (Pretty (..), line, (<+>))
+import Data.Text.Prettyprint.Doc (Pretty (..), group, line, nest)
 import GHC.Stack
 
 import qualified Unsafe.Coerce as TrulyUnsafe
@@ -67,7 +67,7 @@ import qualified RawName as R
 import Name
 import Syntax
 import Type
-import PPrint ()
+import PPrint (prettyBlock)
 import CheapReduction
 import {-# SOURCE #-} Inference
 import MTL1
@@ -425,27 +425,29 @@ buildBlock
   :: ScopableBuilder m
   => (forall l. (Emits l, DExt n l) => m l (Atom l))
   -> m n (Block n)
-buildBlock cont = do
-  Abs decls results <- buildScoped do
-    result <- cont
-    ty <- {-# SCC blockTypeNormalization #-} cheapNormalize =<< getType result
-    return $ result `PairE` ty
-  let (result `PairE` ty) = results
-  let msg = "Decls:" <+> line <> pretty decls <> line
-            <> "Result:" <+> line <> pretty result <> line
-            <> "Of type:" <+> line <> pretty ty
-  ty' <- liftHoistExcept' (docAsStr msg) $ hoist decls ty
-  return $ Block (BlockAnn ty') decls result
+buildBlock cont = buildScoped (cont >>= withType) >>= computeAbsEffects >>= absToBlock
 
-makeBlock :: EnvReader m => Nest Decl n l -> Atom l -> m l (Block n)
-makeBlock decls atom = do
-  ty <- {-# SCC blockTypeNormalization #-} cheapNormalize =<< getType atom
-  return $ makeBlockOfType decls atom ty
+withType :: (EnvReader m, HasType e) => e l -> m l ((e `PairE` Type) l)
+withType e = do
+  ty <- {-# SCC blockTypeNormalization #-} cheapNormalize =<< getType e
+  return $ e `PairE` ty
+{-# INLINE withType #-}
+
+makeBlock :: Nest Decl n l -> EffectRow l -> Atom l -> Type l -> Block n
+makeBlock decls effs atom ty = Block (BlockAnn ty' effs') decls atom where
+  ty' = ignoreHoistFailure $ hoist decls ty
+  effs' = ignoreHoistFailure $ hoist decls effs
 {-# INLINE makeBlock #-}
 
-makeBlockOfType :: Nest Decl n l -> Atom l -> Type l -> Block n
-makeBlockOfType decls atom ty = Block (BlockAnn ty') decls atom
-  where ty' = ignoreHoistFailure $ hoist decls ty
+absToBlock :: (Fallible m) => Abs (Nest Decl) (EffectRow `PairE` (Atom `PairE` Type)) n -> m (Block n)
+absToBlock (Abs decls (effs `PairE` (result `PairE` ty))) = do
+  let msg = "Block:" <> nest 1 (prettyBlock decls result) <> line
+            <> group ("Of type:" <> nest 2 (line <> pretty ty)) <> line
+            <> group ("With effects:" <> nest 2 (line <> pretty effs))
+  ty' <- liftHoistExcept' (docAsStr msg) $ hoist decls ty
+  effs' <- liftHoistExcept' (docAsStr msg) $ hoist decls effs
+  return $ Block (BlockAnn ty' effs') decls result
+{-# INLINE absToBlock #-}
 
 buildPureLam :: ScopableBuilder m
              => NameHint -> Arrow -> Type n
@@ -625,7 +627,7 @@ buildNaryLam binderNest body = do
     block -> emitBlock block
   where
     naryAbsToNaryLam :: Abs (Nest Binder) Block n -> Block n
-    naryAbsToNaryLam (Abs nest block) = case nest of
+    naryAbsToNaryLam (Abs binders block) = case binders of
       Empty -> block
       Nest (b:>ty) bs ->
         AtomicBlock $ Lam $ LamExpr (LamBinder b ty PlainArrow Pure) $
@@ -1320,14 +1322,6 @@ applyReconAbs
 applyReconAbs ab x = do
   xs <- unpackTelescope x
   applyNaryAbs ab $ map SubstVal xs
-
-telescopicCaptureBlock
-  :: (EnvReader m, HoistableE e)
-  => Nest Decl n l -> e l -> m l (Block n, ReconAbs e n)
-telescopicCaptureBlock decls result = do
-  (result', ty, ab) <- telescopicCapture decls result
-  let block = Block (BlockAnn ty) decls result'
-  return (block, ab)
 
 telescopicCapture
   :: (EnvReader m, HoistableE e, HoistableB b)
--- a/src/lib/GenericTraversal.hs
+++ b/src/lib/GenericTraversal.hs
@@ -14,7 +14,6 @@ import Control.Monad
 import Name
 import Builder
 import Syntax
-import Type
 import PPrint
 
 import LabeledItems
@@ -106,13 +105,7 @@ instance GenericallyTraversableE Atom where
 
 instance GenericallyTraversableE Block where
   traverseGenericE (Block _ decls result) = do
-    Abs decls' (PairE ty result') <-
-      buildScoped $ traverseDeclNest decls do
-        result' <- traverseAtom result
-        resultTy <- getType result'
-        return $ PairE resultTy result'
-    ty' <- liftHoistExcept $ hoist decls' ty
-    return $ Block (BlockAnn ty') decls' result'
+    buildBlock $ traverseDeclNest decls $ traverseAtom result
 
 instance GenericallyTraversableE FieldRowElems where
   traverseGenericE elems = do
--- a/src/lib/Inference.hs
+++ b/src/lib/Inference.hs
@@ -35,7 +35,7 @@ import Name
 import Builder
 import Syntax hiding (State)
 import Type
-import PPrint (pprintCanonicalized)
+import PPrint (pprintCanonicalized, prettyBlock)
 import CheapReduction
 import GenericTraversal
 import MTL1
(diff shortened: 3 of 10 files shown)


axch:
Also, which functions should I `INLINE`?  How does one know that?  (Do we have any performance regression tests?)

axch:
Asked Dougal all the TODO(reviewer) questions and updated based on the answers.

axch:
Update: Ran the kernel regression example dexopt mode, and found a ~5% increase in allocations from this patch.  Before:
```
Compiling prelude
Compiling linalg
Compiling png
Compiling diagram
Compiling plot
   6,137,825,880 bytes allocated in the heap
     242,132,664 bytes copied during GC
      31,377,784 bytes maximum residency (12 sample(s))
         300,680 bytes maximum slop
              94 MiB total memory in use (0 MB lost due to fragmentation)

                                     Tot time (elapsed)  Avg pause  Max pause
  Gen  0       358 colls,     0 par    0.282s   0.289s     0.0008s    0.0118s
  Gen  1        12 colls,     0 par    0.171s   0.171s     0.0142s    0.0502s

  TASKS: 5 (1 bound, 4 peak workers (4 total), using -N1)

  SPARKS: 0 (0 converted, 0 overflowed, 0 dud, 0 GC'd, 0 fizzled)

  INIT    time    0.001s  (  0.001s elapsed)
  MUT     time    6.914s  (  6.924s elapsed)
  GC      time    0.453s  (  0.460s elapsed)
  EXIT    time    0.001s  (  0.006s elapsed)
  Total   time    7.368s  (  7.390s elapsed)

  Alloc rate    887,768,712 bytes per MUT second

  Productivity  93.8% of total user, 93.7% of total elapsed
```

After:
```
Compiling prelude
Compiling linalg
Compiling png
Compiling diagram
Compiling plot
   6,452,447,592 bytes allocated in the heap
     242,569,944 bytes copied during GC
      28,255,064 bytes maximum residency (12 sample(s))
         253,096 bytes maximum slop
              84 MiB total memory in use (0 MB lost due to fragmentation)

                                     Tot time (elapsed)  Avg pause  Max pause
  Gen  0       377 colls,     0 par    0.278s   0.279s     0.0007s    0.0050s
  Gen  1        12 colls,     0 par    0.159s   0.159s     0.0133s    0.0434s

  TASKS: 5 (1 bound, 4 peak workers (4 total), using -N1)

  SPARKS: 0 (0 converted, 0 overflowed, 0 dud, 0 GC'd, 0 fizzled)

  INIT    time    0.001s  (  0.001s elapsed)
  MUT     time    7.179s  (  7.314s elapsed)
  GC      time    0.437s  (  0.438s elapsed)
  EXIT    time    0.001s  (  0.008s elapsed)
  Total   time    7.618s  (  7.760s elapsed)

  Alloc rate    898,739,457 bytes per MUT second

  Productivity  94.2% of total user, 94.2% of total elapsed
```

Not sure whether and how to optimize it to reduce that.  Some increased work is expected, because we are computing an effect annotation for every `Block` we construct.

apaszke:
I would be expecting some amount of increase in allocations, because `Block`s are now larger and we use those all over the place. So that's not really that big of an issue to me. What's more concerning is the 5% end-to-end slowdown that is unlikely to be caused by the extra allocations.

src/lib/Builder.hs
```diff
@@ -1321,14 +1325,6 @@ applyReconAbs ab x = do
   xs <- unpackTelescope x
   applyNaryAbs ab $ map SubstVal xs
 
-telescopicCaptureBlock
-  :: (EnvReader m, HoistableE e)
-  => Nest Decl n l -> e l -> m l (Block n, ReconAbs e n)
-telescopicCaptureBlock decls result = do
```
apaszke, line 1327: Was this dead?

apaszke (commented):
Mostly LGTM, but I'd like to get rid of the significant slowdown before this is merged. I hope that inlining the entry points to effect computation will help! Otherwise we can play a bit more with representation of blocks, since they're just all over the place.

src/lib/Builder.hs
```diff
@@ -1152,6 +1154,8 @@ instance GenericE IxImpl where
 
 instance SinkableE IxImpl
 
+instance SubstE Name IxImpl
```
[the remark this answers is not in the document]
axch: Removed.

src/lib/Builder.hs
axch: Yes, it was.

src/lib/GenericTraversal.hs
```diff
@@ -106,13 +106,10 @@ instance GenericallyTraversableE Atom where
...
-      buildScoped $ traverseDeclNest decls do
-        result' <- traverseAtom result
-        resultTy <- getType result'
-        return $ PairE resultTy result'
-    ty' <- liftHoistExcept $ hoist decls' ty
-    return $ Block (BlockAnn ty') decls' result'
+    -- Recompute the effects for the result Block, because the
+    -- traversal may change them
+    buildScoped (traverseDeclNest decls (traverseAtom result >>= withType))
+      >>= computeAbsEffects >>= absToBlock
```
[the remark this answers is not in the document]
axch: Done.

src/lib/Inference.hs
```diff
@@ -736,16 +736,17 @@ checkOrInferRho (WithSrcE pos expr) reqTy = do
...
                 effs' <- checkUEffRow effs
                 ty'   <- checkUType   ty
                 return $ PairE effs' ty'
-            cheapReduceWithDecls decls piResult >>= \case
+            cheapReduceWithDecls decls (effs'' `PairE` result) >>= \case
               (Just (PairE effs' ty'), DictTypeHoistSuccess, []) -> return $ (effs', ty')
               _ -> throw TypeErr $ "Can't reduce type expression: " ++
-                     pprint (Block (BlockAnn TyKind) decls $ snd $ fromPairE piResult)
+                     -- Coercing for printing only
+                     pprint (Block (BlockAnn TyKind $ unsafeCoerceE effs'') decls result)
```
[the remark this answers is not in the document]
axch: Well, status quo is that the Pretty instance of Block does not, in fact, read the annotation.  Encoded that in a separate `prettyBlock` function which can be used in a place like this to avoid constructing a bogus annotation.

src/lib/Inference.hs
```diff
@@ -758,8 +759,9 @@ checkOrInferRho (WithSrcE pos expr) reqTy = do
           bindLamPat (WithSrcB pos' pat) v' $ checkUType ty
         cheapReduceWithDecls decls piResult >>= \case
           (Just ty', DictTypeHoistSuccess, []) -> return ty'
+          -- Pure effects for printing
           _ -> throw TypeErr $ "Can't reduce type expression: " ++
-                 pprint (Block (BlockAnn TyKind) decls piResult)
+                 pprint (Block (BlockAnn TyKind Pure) decls piResult)
```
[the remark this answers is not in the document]
axch: Same.

src/lib/Types/Core.hs
```diff
@@ -153,7 +153,7 @@ data Block n where
   Block :: BlockAnn n l -> Nest Decl n l -> Atom l -> Block n
 
 data BlockAnn n l where
-  BlockAnn :: Type n -> BlockAnn n l
+  BlockAnn :: Type n -> EffectRow n -> BlockAnn n l
```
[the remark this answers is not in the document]
axch: Tried it.  No effect on allocations, though the maximum residency seems to increase by ~3%.  WDYT, should we keep it lazy or revert?

src/lib/Type.hs
```diff
@@ -115,9 +115,29 @@ getClassTy (ClassDef _ _ bs _ _) = go bs
...
 
-exprEffects :: (MonadFail1 m, EnvReader m) => Expr n -> m n (EffectRow n)
+computeAbsEffects :: (EnvExtender m, SubstE Name e)
+  => Abs (Nest Decl) e n -> m n (Abs (Nest Decl) (EffectRow `PairE` e) n)
+computeAbsEffects it = refreshAbs it \decls result -> do
+  effs <- declNestEffects decls
+  return $ Abs decls (effs `PairE` result)
+
+declNestEffects :: (EnvReader m) => Nest Decl n l -> m l (EffectRow l)
+declNestEffects decls = liftEnvReaderM $ declNestEffectsRec decls
```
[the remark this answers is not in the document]
axch: Done.  Also inlined `withType` in Builder.hs for the same reason, and `makeBlock` because it was already inlined before (though it's pure, so maybe it doesn't matter as much).

src/lib/Simplify.hs
```diff
@@ -613,7 +619,8 @@ simplifyBlock :: Emits o => Block i -> SimplifyM i o (Atom o)
 simplifyBlock (Block _ decls result) = simplifyDecls decls $ simplifyAtom result
 
 exceptToMaybeBlock :: Emits o => Block i -> SimplifyM i o (Atom o)
-exceptToMaybeBlock (Block (BlockAnn ty) decls result) = do
+exceptToMaybeBlock (Block (BlockAnn ty _) decls result) = do
+  -- Ignore the effects here because they will implicitly be recomputed.
```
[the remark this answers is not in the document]
axch: Removed.

src/lib/Type.hs
```diff
@@ -115,9 +115,29 @@ getClassTy (ClassDef _ _ bs _ _) = go bs
...
+
+declNestEffectsRec :: Nest Decl n l -> EnvReaderM l (EffectRow l)
+declNestEffectsRec Empty = return mempty
+declNestEffectsRec n@(Nest decl rest) = withExtEvidence n $ do
+  expr <- sinkM $ declExpr decl
+  deff <- exprEffects expr
+  reff <- declNestEffects rest
```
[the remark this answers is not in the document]
axch: The Rec part requires an `EnvExtender` constraint, but when Dougal and I were pair-coding this part, we thought it would be better for the entry point to only require `EnvReader`, since that is semantically more accurate.

But it should recursively call itself (not the non-recursive variant), and I also made the tail-recursion change.  Allocations are down ~1% (though that may be confounded with the effects of inlining).

axch:
Responded to each specific point; PTAL.

As for timing, I'm not sure there is a significant slowdown.  At least on my laptop, timing measurements are noisy enough that I don't trust them to make performance decisions, at least not with such small effects as these.  Should I try a Cloudtop (where I am not concurrently running Chrome, at least)?

Here are my inconclusive timing experiments.  I compared 4 code states: status quo ante, this PR as reviewed, this PR as uploaded post-review, and the same but with keeping the EffectRow field strict.
```
Timings observed for status quo ante
MUT 6.914, Total 7.368
MUT 6.509, Total 6.934
MUT 6.139, Total 6.540
MUT 6.493, Total 6.918
MUT 6.420, Total 6.846

Timings observed for the patch as reviewed
MUT 7.179, Total 7.618
MUT 6.613, Total 7.046
MUT 6.217, Total 6.625
MUT 6.092, Total 6.495
MUT 6.140, Total 6.541

Timings observed for the patch as reviewed (some time later)
MUT 6.732, Total 7.161
MUT 6.904, Total 7.343
MUT 6.803, Total 7.240
MUT 6.940, Total 7.379
MUT 6.831, Total 7.274

Timings observed for the patch as uploaded now
MUT 7.697, Total 8.175
MUT 7.664, Total 8.168
MUT 6.848, Total 7.302
MUT 7.062, Total 7.519

Timings observed for the patch as uploaded now (strict EffectRow)
MUT 6.859, Total 7.301
MUT 6.776, Total 7.212
MUT 6.959, Total 7.401
MUT 6.899, Total 7.344
MUT 7.010, Total 7.457

MUT 7.038, Total 7.476
MUT 7.178, Total 7.627
MUT 7.020, Total 7.471
MUT 6.867, Total 7.307
MUT 6.990, Total 7.440

Timings observed for the patch as uploaded now (some time later)
MUT 6.771, Total 7.216
MUT 6.654, Total 7.088
MUT 6.856, Total 7.296
MUT 6.696, Total 7.131
MUT 6.873, Total 7.313
```

To me, fluctuations in when I run the tests (and perhaps how busy other processes are) seem more significant than any code changes here contemplated.  In your opinion, does this merit further investigation, and if so how?

src/lib/Builder.hs
```diff
@@ -425,27 +425,28 @@ buildBlock
...
-makeBlockOfType decls atom ty = Block (BlockAnn ty') decls atom
-  where ty' = ignoreHoistFailure $ hoist decls ty
+absToBlock :: (Fallible m) => Abs (Nest Decl) (EffectRow `PairE` (Atom `PairE` Type)) n -> m (Block n)
+absToBlock (Abs decls (effs `PairE` (result `PairE` ty))) = do
+  let msg = "Block:" <> nest 1 (prettyBlock decls result) <> line
+            <> group ("Of type:" <> nest 2 (line <> pretty ty)) <> line
+            <> group ("With effects:" <> nest 2 (line <> pretty effs))
+  ty' <- liftHoistExcept' (docAsStr msg) $ hoist decls ty
+  effs' <- liftHoistExcept' (docAsStr msg) $ hoist decls effs
+  return $ Block (BlockAnn ty' effs') decls result
```
apaszke, line 449: Can you please at least mark this as `INLINABLE`? It still runs in some arbitrary monad and we should at least open up the opportunities to specialization, though I imagine this will be inhibited by the huge `msg` expression in its body

src/lib/GenericTraversal.hs
```diff
@@ -106,13 +105,9 @@ instance GenericallyTraversableE Atom where
...
   traverseGenericE (Block _ decls result) = do
-    Abs decls' (PairE ty result') <-
-      buildScoped $ traverseDeclNest decls do
-        result' <- traverseAtom result
-        resultTy <- getType result'
-        return $ PairE resultTy result'
-    ty' <- liftHoistExcept $ hoist decls' ty
-    return $ Block (BlockAnn ty') decls' result'
+    -- Recompute the effects for the result Block, because the
+    -- traversal may change them
```
apaszke, line 109: There's no effect recomputation visible here now. It's all in `buildBlock`, so I find this a bit confusing

src/lib/Simplify.hs
```diff
@@ -438,14 +440,18 @@ simplifyAbs (Abs bs body) = fromPairE <$> do
   substBinders bs \bs' -> do
     ab <- buildScoped $ simplifyBlock body
     refreshAbs ab \decls result -> do
-      getType result >>= isData >>= \case
+      effs <- declNestEffects decls
```
apaszke, line 443: Actually, we could probably make this faster: `simplifyBlock` never modifies the set of effects (or effect for that matter) in a block, so we could reuse the annotations of `body`. Now that I look at it, even the `getType result` part is unnecessary, since we could `substM` the type from `body`'s annotation

src/lib/Type.hs
```diff
@@ -115,9 +115,31 @@ getClassTy (ClassDef _ _ bs _ _) = go bs
...
+  return $ Abs decls (effs `PairE` result)
+{-# INLINE computeAbsEffects #-}
+
+declNestEffects :: (EnvReader m) => Nest Decl n l -> m l (EffectRow l)
+declNestEffects decls = liftEnvReaderM $ declNestEffectsRec decls mempty
+{-# INLINE declNestEffects #-}
+
+declNestEffectsRec :: Nest Decl n l -> EffectRow l -> EnvReaderM l (EffectRow l)
+declNestEffectsRec Empty !acc = return acc
+declNestEffectsRec n@(Nest decl rest) !acc = withExtEvidence n $ do
```
apaszke, line 131: This is only a suggestion. This code is ok, but we have `BlockArguments` enabled, so `$ do ...` is equivalent to just `do ...`

src/lib/Type.hs
```diff
@@ -115,9 +115,31 @@ getClassTy (ClassDef _ _ bs _ _) = go bs
...
+  expr <- sinkM $ declExpr decl
+  deff <- exprEffects expr
```
apaszke, line 133: Ohhh I wonder how much of the slowdown is actually due to `exprEffects` not getting specialized to `EnvReaderM`!

src/lib/Types/Core.hs
```diff
@@ -153,7 +153,11 @@ data Block n where
   Block :: BlockAnn n l -> Nest Decl n l -> Atom l -> Block n
 
 data BlockAnn n l where
-  BlockAnn :: Type n -> BlockAnn n l
+  -- Make the EffectRow lazy because it requires a traversal to
+  -- compute, and there are many places in the code that ignore it.
+  -- (Also, it's a S.Set, so isn't obviously much smaller than the
+  -- thunk.)
+  BlockAnn :: Type n -> ~(EffectRow n) -> BlockAnn n l
```
apaszke, line 160: Since there doesn't seem to be any significant difference in your benchmarks, we could just make it eager for simplicity.

apaszke (approved):
LGTM. I think there are still two potential optimization opportunities (`simplifyAbs` and specialization of `exprEffects`), but this is looking great.

src/lib/Types/Core.hs
axch: Done.

src/lib/Type.hs
axch: Done.

src/lib/GenericTraversal.hs
axch: Comment removed.

src/lib/Builder.hs
axch: Done.

src/lib/Simplify.hs
axch: The annotation doesn't always exist, so the `getType result` code path must still be present.  I'm assuming `getType` (of an `Atom`) isn't so expensive that the extra `case` is worth it, but I am removing the `declNestEffects` call.

src/lib/Type.hs
axch: `SPECIALIZE` pragma added.

Between the two changes, we now have
```
MUT 6.551, Total 6.974
MUT 6.614, Total 7.040
MUT 6.402, Total 6.817
MUT 6.489, Total 6.907
MUT 6.453, Total 6.876
```

which is either a significant improvement or some other process on my laptop having given up in the interim.

PR #886, opened 2022-04-29
Precise effects

### LINKED ISSUE #848: AD + state effect + case = crash
Reproduction:

```
def min' (a:Float) (b:Float) : Float =
  yield_state a \s.
    best = get s
    new_best = select (best < b) best b
    s := new_best

grad (\x . (min' x (x+1))) 1.0
> Compiler bug!
>
> Missing zero case for a tangent type: (|Unit | Unit|)
>   error, called at src/lib/Builder.hs:820:19 in dex-0.1.0.0-8hDfthyGTXmzhkTo2ydOn:Builder
>   zeroAt, called at src/lib/Builder.hs:812:34 in dex-0.1.0.0-8hDfthyGTXmzhkTo2ydOn:Builder
>   zeroAt, called at src/lib/Builder.hs:857:11 in dex-0.1.0.0-8hDfthyGTXmzhkTo2ydOn:Builder
```

Expected behavior: gradient at x should be 1.

axch:
Compute (more) precise effects for case, for, and "effect lambda".  Doing this for case fixes #848, and Dougal and I picked up the others since they were pretty easy.

### DIFF
--- a/src/lib/Builder.hs
+++ b/src/lib/Builder.hs
@@ -708,15 +708,17 @@ buildCase :: (Emits n, ScopableBuilder m)
           -> (forall l. (Emits l, DExt n l) => Int -> [AtomName l] -> m l (Atom l))
           -> m n (Atom n)
 buildCase scrut resultTy indexedAltBody = do
-  eff <- getAllowedEffects
   scrutTy <- getType scrut
   altsBinderTys <- caseAltsBinderTys scrutTy
-  alts <- forM (enumerate altsBinderTys) \(i, bs) -> do
-    buildNaryAbs bs \xs -> do
-      buildBlock do
+  (alts, effs) <- unzip <$> forM (enumerate altsBinderTys) \(i, bs) -> do
+    (Abs bs' (body `PairE` eff')) <- buildNaryAbs bs \xs -> do
+      blk <- buildBlock do
         ListE xs' <- sinkM $ ListE xs
         indexedAltBody i xs'
-  liftM Var $ emit $ Case scrut alts resultTy eff
+      eff <- effectsE blk
+      return $ blk `PairE` eff
+    return (Abs bs' body, ignoreHoistFailure $ hoist bs' eff')
+  liftM Var $ emit $ Case scrut alts resultTy $ mconcat effs
 
 buildSplitCase :: (Emits n, ScopableBuilder m)
                => LabeledItems (Type n) -> Atom n -> Type n
@@ -1189,7 +1191,7 @@ indexSetSize ty = do
 -- XXX: the first argument is the true case, following the
 -- surface-level `if ... then ... else ...`, but the order
 -- is flipped in the actual `Case`, because False acts like 0.
--- TODO: consider a version that figures out the resul type itself.
+-- TODO: consider a version that figures out the result type itself.
 emitIf :: (Emits n, ScopableBuilder m)
        => Atom n
        -> Type n
--- a/src/lib/Builder.hs
+++ b/src/lib/Builder.hs
@@ -741,9 +741,17 @@ buildEffLam
 buildEffLam rws hint ty body = do
   eff <- getAllowedEffects
   buildLam noHint PlainArrow TyKind Pure \h -> do
-    let eff' = extendEffect (RWSEffect rws (Just h)) (sink eff)
-    buildLam hint PlainArrow (RefTy (Var h) (sink ty)) eff' \ref ->
-      body (sink h) ref
+    let ty' = RefTy (Var h) (sink ty)
+    withFreshBinder hint (LamBinding PlainArrow ty') \b -> do
+      let ref = binderName b
+      h' <- sinkM h
+      let eff' = extendEffect (RWSEffect rws (Just h')) (sink eff)
+      body' <- withAllowedEffects eff' $ buildBlock $ body (sink h) $ sink ref
+      -- Contract the type of the produced function to only mention
+      -- the effects actually demanded by the body.  This is safe because
+      -- it's immediately consumed by an effect discharge primitive.
+      effs <- effectsE body'
+      return $ Lam $ LamExpr (LamBinder b ty' PlainArrow effs) body'
 
 buildForAnn
   :: (Emits n, ScopableBuilder m)
--- a/src/lib/Builder.hs
+++ b/src/lib/Builder.hs
@@ -751,9 +751,11 @@ buildForAnn
   -> (forall l. (Emits l, DExt n l) => AtomName l -> m l (Atom l))
   -> m n (Atom n)
 buildForAnn hint ann ty body = do
-  -- TODO: consider only tracking the effects that are actually needed.
-  eff <- getAllowedEffects
-  lam <- buildLam hint PlainArrow ty eff body
+  lam <- withFreshBinder hint (LamBinding PlainArrow ty) \b -> do
+    let v = binderName b
+    body' <- buildBlock $ body $ sink v
+    effs <- effectsE body'
+    return $ Lam $ LamExpr (LamBinder b ty PlainArrow effs) body'
   liftM Var $ emit $ Hof $ For ann lam
 
 buildFor :: (Emits n, ScopableBuilder m)
--- a/src/lib/PPrint.hs
+++ b/src/lib/PPrint.hs
@@ -39,21 +39,48 @@ import Name
 import Syntax
 import Parser (showPrimName)
 
--- Specifies what kinds of operations are allowed to be printed at this point.
--- Printing at AppPrec level means that applications can be printed
--- with no parentheses, but binops must be parenthesized.
-data PrecedenceLevel  = LowestPrec
-                      | AppPrec
-                      | ArgPrec  -- Only single symbols and parens allowed
-                      deriving (Eq, Ord)
-
+-- A DocPrec is a slightly context-aware Doc, specifically one that
+-- knows the precedence level of the immediately enclosing operation,
+-- and can decide to parenthesize itself accordingly.
+-- For example, when printing `x = f (g 1)`, we know that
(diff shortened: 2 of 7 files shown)


axch:
Performance-wise, this change reduces allocations in kernelregression.dx by ~23%, and also seems (insofar as this is trustworthy) to be good for runtime
```
MUT 6.379, Total 6.809
MUT 6.478, Total 6.908
MUT 6.605, Total 7.031
MUT 6.174, Total 6.584
MUT 6.014, Total 6.424
```
(though that may just be that I walked away from the computer while running these?)

The CI failure is specific to the debug build (some giant intermediate expression doesn't typecheck?).  Not sure how to minimize and debug the failure.

axch:
Found the bug that was causing CI to fail.

src/lib/PPrint.hs
```diff
@@ -39,21 +39,48 @@ import Name
...
   prettyPrec :: a -> DocPrec ann
 
 -- `atPrec prec doc` will ensure that the precedence level is at most
 -- `prec` when running `doc`, wrapping with parentheses if needed.
+-- To wit,
+-- - `atPrec LowestPrec` means "wrap unless the context permits all
+--   subexpressions unwrapped"
+-- - `atPrec AppPrec` means "wrap iff the context requires ArgPrec"
+-- - `atPrec ArgPrec` means "always wrap" (unless the
+--   `PrecedenceLevel` ADT is extended later).
```
apaszke, line 83: I'm not sure if I follow this. If you `atPrec ArgPrec $ Con (Lit (Int32Lit 2))` then it should be printed without parens, so "always wrap" seems incorrect?

src/lib/Type.hs
```diff
@@ -115,27 +116,12 @@ getClassTy (ClassDef _ _ bs _ _) = go bs
...
 -- === querying effects ===
 
-computeAbsEffects :: (EnvExtender m, SubstE Name e)
-  => Abs (Nest Decl) e n -> m n (Abs (Nest Decl) (EffectRow `PairE` e) n)
-computeAbsEffects it = refreshAbs it \decls result -> do
-  effs <- declNestEffects decls
-  return $ Abs decls (effs `PairE` result)
-{-# INLINE computeAbsEffects #-}
+class HasEffectsE (e::E) where
+  effectsE :: (EnvReader m) => e n -> m n (EffectRow n)
```
apaszke, line 120: Couldn't we specialize this method to `EnvReaderM`? It's easy enough to `liftEnvReaderM`, or even to define
```
class HasEffectsE e where
  effectsEImpl :: e n -> EnvReaderM n (EffectRow n)

effectsE = liftEnvReaderM . effectsEImpl
{-# INLINE effectsE #-}
```
That would have the same ergonomics, but would let us guarantee that the relevant specializations are happening. At this point I tend to consider monad-polymorphic typeclass methods a smell (unless we're _very_ careful to inline them).

src/lib/Simplify.hs
```diff
@@ -151,6 +149,13 @@ simplifyExpr expr = confuseGHC >>= \_ -> case expr of
             liftM Var $ emit $ Case e' alts' resultTy' eff'
           False -> defuncCase e' alts resultTy'
 
+caseComputingEffs
+  :: forall m n. (MonadFail1 m, EnvReader m)
+  => Atom n -> [Alt n] -> Type n -> m n (Expr n)
+caseComputingEffs scrut alts resultTy = do
+  effs <- forM alts effectsE
```
apaszke, line 156: you might be able to `foldMapM` here. I think we define that in Utils. Then this whole function becomes
```
Case scrut alts resultTy <$> foldMapM effectE alts
```

src/lib/Builder.hs
```diff
@@ -708,15 +708,16 @@ buildCase :: (Emits n, ScopableBuilder m)
...
   altsBinderTys <- caseAltsBinderTys scrutTy
-  alts <- forM (enumerate altsBinderTys) \(i, bs) -> do
-    buildNaryAbs bs \xs -> do
+  (alts, effs) <- unzip <$> forM (enumerate altsBinderTys) \(i, bs) -> do
+    blk <- buildNaryAbs bs \xs -> do
       buildBlock do
         ListE xs' <- sinkM $ ListE xs
         indexedAltBody i xs'
-  liftM Var $ emit $ Case scrut alts resultTy eff
+    eff <- effectsE blk
```
apaszke, line 718: The usage of `effectsE` in this patch follows a slightly annoying pattern where you often build up a bunch of `Alt`s and then have to open them up (through `refreshAbs`) in the implementation of `effectsE`. Now, since there are generally no emissions between their building and their analysis, this opening up should amount to checking that the names bound in the `Alt` are fresh, but it's still a check we have to do. It would be nice to consider if we could perhaps just call `effectsE` between `buildNaryAbs` and `buildBlock`.

src/lib/Type.hs
```diff
@@ -17,7 +17,7 @@ module Type (
   oneEffect, lamExprTy, isData, asFirstOrderFunction, asFFIFunType,
   isSingletonType, singletonTypeVal, asNaryPiType,
   numNaryPiArgs, naryLamExprType, getMethodIndex,
-  extendEffect, effectsE, exprEffects, declNestEffects,
```
apaszke: This is newly exported, but I can't see any uses

src/lib/Builder.hs
```diff
@@ -740,9 +740,14 @@ buildEffLam
...
-    buildLam hint PlainArrow (RefTy (Var h) (sink ty)) eff' \ref ->
-      body (sink h) ref
+    let ty' = RefTy (Var h) (sink ty)
+    withFreshBinder hint (LamBinding PlainArrow ty') \b -> do
+      let ref = binderName b
+      h' <- sinkM h
+      let eff' = extendEffect (RWSEffect rws (Just h')) (sink eff)
+      body' <- withAllowedEffects eff' $ buildBlock $ body (sink h) $ sink ref
+      effs <- effectsE body'
+      return $ Lam $ LamExpr (LamBinder b ty' PlainArrow effs) body'
```
apaszke, line 750: This change is a little unclear to me. Why did we even have to do it? Is there some subtle interplay between your patches and `getAllowedEffects`?

apaszke: Would this be fixed if only we adjusted `buildLam`?

apaszke (approved):
Very nice! I left a bunch of comments inline. Do you have any sense for where might the performance improvements be coming from? It is quite puzzling to me.

src/lib/PPrint.hs
axch: Good catch.  That was a typo for "never wrap".  Fixed.

src/lib/Type.hs
axch: Done.

src/lib/Simplify.hs
axch: Done.

src/lib/Type.hs
axch: Good catch.  Unexported.

src/lib/Builder.hs
axch: Why do this at all?  This way, the effects associated with an effect lambda are more precise than before.  The difference is that this implementation extracts the effects from the block and uses them in the `LamBinder`, whereas `buildLam` puts the input effects `eff'` in the `LamBinder`.

Why do this here and not in `buildLam`?  Dougal and I went back and forth on this a bit.  The issue, as I currently understand it, is that an `EffectRow` can mean one of three different things: it can mean the set of effects permitted in some context (which is what `getAllowedEffects` produces); it can mean the set of effects actually used by some code (which is what the annotations on `Block` and `Case` now mean); and it can be part of the type of a function, where it serves both purposes (limiting what's allowed in the body, and indicating what's demanded by calls).

So, at the level of `buildLam`, we have to treat the input effect row as defining the type of the function be are building, which in general we are not free to modify.  However, at the more specific `buildEffLam`, we know the function is about to be consumed by an effect discharge primitive (I think?), so we actually are free to produce a function with a more specific type.

Historical note: I actually tried it the other way first, always tagging every `LamExpr` with the effects consumed by the `Block`, but hit a test failure I didn't understand.  Dougal gave me the above explanation about function types, and we decided to back off to this more limited change.

Added a brief comment to the code to allude to this explanation.

src/lib/Builder.hs
axch: Thanks for the suggestion!  Moved extracting the block effect under `buildNaryAbs`.  No visible effect on allocations (perhaps none was expected), but seems to improve runtime (again, with the usual caveats about noise).

Not sure what you mean by "pattern", though.  This seems to be the only place in this patch where this happens?

axch:
I don't have a super-clear idea of why performance improves so much from this, but generically I expect having less stuff to always be good.  Completely speculating, maybe this change detects a material number of actually-pure subexpressions, which in turn allow more inlining or a simpler representation?  Or some code gets proven dead and eliminated that wasn't before?

src/lib/Simplify.hs
```diff
@@ -151,6 +149,12 @@ simplifyExpr expr = confuseGHC >>= \_ -> case expr of
             liftM Var $ emit $ Case e' alts' resultTy' eff'
           False -> defuncCase e' alts resultTy'
 
+caseComputingEffs
+  :: forall m n. (MonadFail1 m, EnvReader m)
+  => Atom n -> [Alt n] -> Type n -> m n (Expr n)
+caseComputingEffs scrut alts resultTy = do
+  Case scrut alts resultTy <$> foldMapM effectsE alts
```
apaszke, line 156: We should make sure that this either gets inline or is specialized to `SimplifyM`!

axch: Done (commit 289e8ceb79ae06547dcf6616176b229b1a0bf901).

PR #890, opened 2022-05-03
Moved FFT to lib and split off its tests.

duvenaud:
Now that FFT works again, it seems like a good time to move it into `/lib`, and split off its tests into their own file.

Now this doesn't need to be done anymore in @srush 's [MFCC demo](#603) PR.


apaszke (approved):
Nice!

PR #892, opened 2022-05-04
Teach transposition not to emit runWriters with trivial accumulators

axch:
The theory is that this micro-optimization will improve compile times by reducing the size of the IR that needs to be processed after transposition.

Indeed, on a test example (not committed) of differentiating through the log-density of a 1-D multivariate Gaussian, this
- reduces allocations by 2.5%
- reduces maximum memory residency by 10%
- seems to improve end-to-end execution time by 10% (subject to the usual measurement noise)

Co-authored-by: Dougal Maclaurin <dougalm@google.com>

### DIFF
--- a/src/lib/Transpose.hs
+++ b/src/lib/Transpose.hs
@@ -28,8 +28,8 @@ transpose lam = liftBuilder do
   let resultTy' = ignoreHoistFailure $ hoist piBinder resultTy
   runReaderT1 (ListE []) $ runSubstReaderT idSubst $
     buildLam noHint LinArrow resultTy' Pure \ct ->
-      withAccumulator (sink argTy) \ref ->
-        extendSubst (b @> LinRef ref) $
+      withAccumulator (sink argTy) \refSubstVal ->
+        extendSubst (b @> refSubstVal) $
           transposeBlock body (sink $ Var ct)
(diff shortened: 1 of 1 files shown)


src/lib/Transpose.hs
```diff
@@ -70,12 +70,21 @@ isLin e = do
...
-  -> (forall o'. (Emits o', DExt o o') => Atom o' -> TransposeM i o' ())
+  -> (forall o'. (Emits o', DExt o o') => TransposeSubstVal AtomNameC o' -> TransposeM i o' ())
   -> TransposeM i o (Atom o)
 withAccumulator ty cont = do
-  baseMonoid <- getBaseMonoidType ty >>= tangentBaseMonoidFor
-  getSnd =<< emitRunWriter noHint ty baseMonoid \_ ref ->
-               cont (Var ref) >> return UnitVal
+  singletonTypeVal ty >>= \case
+    Nothing -> do
+      baseMonoid <- getBaseMonoidType ty >>= tangentBaseMonoidFor
```
apaszke, line 78: nit: it would read more like English if you did `tangentBaseMonoidFor =<< getBaseMonoidType ty`

apaszke (approved):

PR #893, opened 2022-05-06
Convert transposeLinear to transpose_linear, yieldAccum to yield_accum

niklasschmitz:
This PR converts the two prelude-exported names `transposeLinear` and `yieldAccum` to snake_case (to match e.g. `yield_state`) in line with #786, as follow-up to PR #814.


apaszke (approved):
Looks great! Thanks for fixing this.

PR #894, opened 2022-05-10
Type querying

axch:
Write separate code for just querying types, as opposed to the check-and-query operation implemented in Type.hs.

Why?  Because checking types is restricted to rename-only substitutions (which the checker itself generates), but we want to be able to query types through a more meaningful substitution, such as the one carried by SimplifyM.  Also, specializing the implementation to just querying lets us worry less about relying on laziness (in the HardFail monad) for performance.

Classify type API endpoints (as best I could) by their querying or checking nature and move the querying ones to the new QueryType module.  After the split, rename Type.hs to CheckType.hs, since that's now what it's for.

The first commit in this PR reduces allocations on the kernel regression example by 6%, and improves runtime by about 2.5%.  I haven't carefully checked the whole PR, but the rest should be performance-neutral (or incidentally helpful, e.g., due to opportunistic inline annotations and monad concretizations).


axch:
Question: Should the "Data class" subsystem move to QueryType?  Pro: it seems semantically pretty query-oriented, but Con: the component thereof `checkFFIFunTypeM` is reused in CheckType in a TyperT monad.  The exported interface I'm thinking of is `isData`, `asFFIFunType`, and `asFirstOrderFunction`.

src/lib/Imp.hs
```diff
@@ -1589,9 +1590,10 @@ impInstrTypes instr = case instr of
 -- TODO: reuse type rules in Type.hs
 impOpType :: IPrimOp n -> IType
 impOpType pop = case pop of
-  ScalarBinOp op x y -> runHardFail $ checkBinOp op (getIType x) (getIType y)
-  ScalarUnOp  op x   -> runHardFail $ checkUnOp  op (getIType x)
-  VectorBinOp op x y -> runHardFail $ checkBinOp op (getIType x) (getIType y)
+  ScalarBinOp op x _ -> typeBinOp op (getIType x)
+  -- All unary ops preserve the type of their input
```
apaszke, line 1594: While this is true right now, I don't think this is a good idea. Can't we use some type querying function from `QueryType` here? With this we end up with the typing rules for `UnOp`s scattered across 3 files and I'm afraid those will easily get out of sync. You won't get a compiler warning here if you add more cases!

src/lib/Linearize.hs
```diff
@@ -143,10 +143,8 @@ liftTangentM args m = liftSubstReaderT $ lift11 $ runReaderT1 args m
 
 isTrivialForAD :: Expr o -> PrimalM i o Bool
 isTrivialForAD expr = do
-  trivialTy  <- (maybeTangentType <$> getType expr) >>= \case
-    Nothing -> return False
-    Just tTy -> isSingletonType tTy
-  hasActiveEffs <- exprEffects expr >>= \case
+  trivialTy  <- any isSingletonType . maybeTangentType <$> getType expr
```
apaszke, line 146: FWIW I find this use of `any` a bit confusing. The previous explicit statement was clearer, because one didn't have to think what is the neutral element of the "any" monoid.

src/lib/QueryType.hs
```diff
@@ -0,0 +1,849 @@
...
+getAppType f xs = liftTypeQueryM idSubst $ typeApp f xs
+{-# INLINE getAppType #-}
+
+getTabAppType :: EnvReader m => Type n -> [Atom n] -> m n (Type n)
+getTabAppType f xs = case nonEmpty xs of
+  Nothing -> getType f
+  Just xs' -> liftTypeQueryM idSubst $ typeTabApp f xs'
+{-# INLINE getTabAppType #-}
+
+getMethodType :: EnvReader m => ClassName n -> Int -> m n (Type n)
```
apaszke, line 90: This is exported, but I can't find any uses. Perhaps remove it form the export list to allow GHC to specialize/inline this internally to this file?

apaszke (approved):
LGTM! Haven't reviewed QueryType very carefully, since I assume it's mostly code moved from the old Type.

src/lib/Imp.hs
axch: Done.

src/lib/Linearize.hs
axch: Clarified.

src/lib/QueryType.hs
axch: Unexported.

axch: Never mind, it's actually used in CheckType.

PR #900, opened 2022-05-14
Print strings like `"abc"` instead of `AsList 3 "abc"`, matching the way we parse them.

dougalm:


axch (approved):
LGTM, but why did the Github actions not run the Haskell test suite (only Julia)?

dougalm:
> LGTM, but why did the Github actions not run the Haskell test suite (only Julia)?

I think it only runs the Haskell tests when you're proposing to merge into main, but this is proposing to merge into "fenceposts" since it's a follow-up on that.

src/lib/Serialize.hs
```diff
@@ -90,24 +90,26 @@ getTableElements tab = do
 -- This doesn't handle parentheses well. TODO: treat it more like PrettyPrec
 prettyVal :: (MonadIO1 m, EnvReader m, Fallible1 m) => Val n -> m n (Doc ann)
 prettyVal val = case val of
+  -- Pretty-print strings
+  DataCon "AsList" _ [Word8Ty] _ [_, tab@(TabVal _ _)] -> do
+    atoms <- getTableElements tab
```
apaszke, line 95: Can you please use `getDexString` here? It's _a lot_ faster

apaszke (approved):
Nice!

src/lib/Serialize.hs
dougalm: Good idea

PR #901, opened 2022-05-17
Reimplement filter in terms of cat_maybes, so it's not asymptotically slow

axch:


dougalm (approved):
Good idea. For context, I think David wrote the original `filter` in terms of monoidal accumulation with the hope that we'd be able to parallelize it. But to actually do that we'll need efficient parallel list accumulation. Your current version is better for now, and at least it allows the evaluation of `f` on each element to be done in parallel, but we should consider going back to the accumulation version one day if we can make it efficient.

apaszke (approved):

PR #902, opened 2022-05-17
Open the log file exactly once, if it exists.

### LINKED ISSUE #815: Dex --logto can't write to a file?
```
$ dex script foo.dx --logto log.txt
dex: log.txt: openFile: resource busy (file is locked)
```

Also happens when I try to fool it with an output redirection
```
$ dex script foo.dx --logto /dev/stdout > log.txt
dex: /dev/stdout: openFile: resource busy (file is locked)
```

I can work around by piping through `cat`, like so:
```
$ dex script foo.dx --logto /dev/stdout | cat > log.txt
```
but still.

I'm guessing this falls into the broad class of "lazy IO problems", but I don't know where to look for it beyond that.

axch:
Fixes #815.


dougalm (approved):
Looks great

src/lib/TopLevel.hs
```diff
@@ -59,7 +59,8 @@ data EvalConfig = EvalConfig
   { backendName   :: Backend
   , libPath       :: Maybe FilePath
   , preludeFile   :: Maybe FilePath
-  , logFile       :: Maybe FilePath }
+  , logFileName   :: Maybe FilePath
```
apaszke, line 62: Do we even need the file name? Perhaps we could construct the handle before we create `EvalConfig`?

axch: Well, we need some ADT for commandline flags, no?

Plus, in case Dex ever wants to know what the name of the log file was (such as to report where the logs were printed to), it's not crazy to keep it.

PR #906, opened 2022-05-18
Be more careful about skipping zonks of unsolved env

apaszke:
Previous implementation tried to be smart and avoid filtering the
unsolved part of the environment when the solver substitution was empty.
Seems sensible, right? After all, if the solution hasn't been extended,
then nothing should change, right? Right??

Well, no. We were often very conservative when constructing the unsolved
sets and whenever the solver subst was empty we would fail to filter out
the unnecessary entries from them. Those added up, and made us filter
huge unsolved envs down the road. This wouldn't be a big deal with a
flat block, but if the filtering happens under a nested lambda, then
those changes might not propagate back (due to how `refreshAbs` is
implemented)!

Now, here's the good part. I made a silly benchmark: a lambda that contained
```
ve = for i0:(%IntRange (%monoLit 0) (%monoLit (53))). max ((vc).(i0)) (f_to_f64 (0.000000))
```
repeated 2000 times in its body, and nothing else. Without this change,
inference took over 30s to elaborate this expression. Now, the whole
program (including parsing) terminates in a little under a second! 30x
speedup in that case, and even more in other, since this effectively
makes our inference linear instead of being quadratic.

### DIFF
--- a/src/lib/Inference.hs
+++ b/src/lib/Inference.hs
@@ -284,16 +284,6 @@ instance SinkableE UnsolvedEnv where
 getAtomNames :: Distinct l => EnvFrag n l -> S.Set (AtomName l)
 getAtomNames frag = S.fromList $ nameSetToList $ toNameSet $ toScopeFrag frag
 
-extendInfOutMapSolver :: Distinct n => InfOutMap n -> SolverSubst n -> InfOutMap n
-extendInfOutMapSolver (InfOutMap bindings ss dd un) ss' = do
-  -- We won't solve any names if the solver subst extension is empty.
-  let (un', bindings') = if substIsEmpty ss' then (un, bindings) else zonkUnsolvedEnv ss' un bindings
-  let ssFinal = catSolverSubsts (toScope bindings) ss ss'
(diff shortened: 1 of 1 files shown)


src/lib/Inference.hs
```diff
@@ -269,8 +269,10 @@ instance ExtOutMap InfOutMap EnvFrag where
   extendOutMap (InfOutMap bindings ss dd oldUn) frag =
     withExtEvidence frag do
       let newUn = UnsolvedEnv $ getAtomNames frag
-      -- as an optimization, only do the zonking for the new stuff
       let newEnv = bindings `extendOutMap` frag
+      -- As an optimization, only do the zonking for the new stuff.
+      -- Also, we zonk unconditionally, even when the solver substitution
```
dougalm, line 274: Maybe this second comment ("Also, ...")  should go with the new definition of `zonkUnsolvedEnv`, since nothing has changed here.

dougalm (approved):
Nice. Very subtle!

src/lib/Inference.hs
apaszke: Good idea. I moved it there!

PR #911, opened 2022-05-19
Allow dex web to be used without a network connection

### LINKED ISSUE #610: Configure `dex web` to work offline
Very lo-pri. 

Right now there are some mandatory JS files that need an internet connection. Would be nice if things were setup locally so you can Dex on a Plane.

axch:
The issue was that the dynamic content rendering loop was only started
after KaTeX was loaded, in the `onload` property.  But naturally
enough that trigger doesn't run if the load fails, as when there is no
network.  Added starting the same loop `onerror` as well, and tweaked
the Dex-specific Javascript to survive KaTeX being unavailable (by
just not rendering any LaTeX with it).

Fixes #610.

### DIFF
--- a/static/index.js
+++ b/static/index.js
@@ -44,11 +44,13 @@ function lookup_address(cell, address) {
 }
 
 function renderLaTeX() {
-    // Render LaTeX equations in prose blocks via KaTeX.
-    var proseBlocks = document.querySelectorAll(".prose-block");
-    Array.from(proseBlocks).map((proseBlock) =>
-        renderMathInElement(proseBlock, katexOptions)
-    );
+    // Render LaTeX equations in prose blocks via KaTeX, if available.
+    if (typeof renderMathInElement != 'undefined') {
+        var proseBlocks = document.querySelectorAll(".prose-block");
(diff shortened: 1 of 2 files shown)


static/index.js
```diff
@@ -44,11 +44,21 @@ function lookup_address(cell, address) {
...
+    }
     // Render LaTeX equations in prose blocks via KaTeX.
     var proseBlocks = document.querySelectorAll(".prose-block");
     Array.from(proseBlocks).map((proseBlock) =>
         renderMathInElement(proseBlock, katexOptions)
     );
+        var proseBlocks = document.querySelectorAll(".prose-block");
+        Array.from(proseBlocks).map((proseBlock) =>
+            renderMathInElement(proseBlock, katexOptions)
+        );
```
apaszke, line 60: What happened here? This code seems to be duplicated?

dan-zheng: Yes – fixed later by @axch in https://github.com/google-research/dex-lang/commit/02ac25c2ce77e71cb61c67b2a7a889a6c70eeace.

PR #918, opened 2022-05-21
Fix type annotation for table literals

axch:


src/lib/Inference.hs
```diff
@@ -1812,33 +1813,57 @@ checkExtLabeledRow (Ext types (Just ext)) = do
...
+      staticallyKnownIdx (Abs decls res) = unsafeLiftInterpMCatch (evalDecls decls $ evalExpr $ Atom res) >>= \case
+        (Success (IdxRepVal ix)) -> return $ Just ix
+        _ -> return Nothing
+
+      context = "If attempting to construct a fixed-size table not\n" <>
+                "indexed by 'Fin n' for some static n, this error may\n" <>
+                "indicate there was not enough information to infer\n" <>
+                "a concrete index set; try adding an explicit\n" <>
+                "annotation."
+      {-# INLINE staticallyKnownIdx #-}
```
apaszke, line 1873: Please move this next to the function definition

apaszke (approved):
Looks good!

src/lib/Interpreter.hs
```diff
@@ -321,6 +334,18 @@ interpIndexSetSize ixTy = liftBuilderInterp $ indexSetSize $ sink ixTy
...
 interpIntToIndex ixTy i = liftBuilderInterp $ intToIndex (sink ixTy) (sink i)
 
+-- A variant of `indices` that accepts an expected number of them and
+-- only tries to construct the set if the expected number matches the
+-- given index set's size.
+indicesLimit :: EnvReader m => Int -> IxType n -> m n (Either Int32 [Atom n])
+indicesLimit sizeReq ixTy = do
+  ~(IdxRepVal size) <- interpIndexSetSize ixTy
+  if size == fromIntegral sizeReq then
+    Right <$> forM [0..size-1] \i -> interpIntToIndex ixTy $ IdxRepVal i
```
apaszke, line 344: This `forM` will likely run in a virtualized monad that lifts to interpreter at every iteration. Perhaps we could just run it all in the interpreter monad?

src/lib/Inference.hs
axch: Done.

src/lib/Interpreter.hs
axch: Done.

PR #920, opened 2022-05-23
Separate `stack test` work space from `stack build` work space

axch:
because they evict each other's caches.


apaszke (approved):
LGTM, but please add this new directory to the [CI cache](https://github.com/google-research/dex-lang/blob/83555d8836d1a35c0000b7f54b86524a1c048ca4/.github/workflows/haskell-ci.yaml#L47).

PR #927, opened 2022-05-26
Add Levenshtein Distance to the set of docs auto-built for Github Pages

axch:
(oops)


dougalm (approved):
Looks good. We could also just comment out the timing parts, leaving the timings we saw as comments.

PR #928, opened 2022-05-27
Add a prototype of an incremental lowering path

apaszke:
The jump from Core to Imp (implemented in `Imp.hs`) is huge! We have to
do the AoS/SoA conversion, bufferize our code, do destination passing to
elide copies (which apparently we no longer do since safer names!),
assign buffers to memory spaces, etc etc. This is a lot, and once you do
so much in a single pass it is easy to make mistakes, or suboptimal
decisions.

To solve that problem I want us to have an intermediate "abstract
machine" IR for Dex, which is where most of those operations would be
performed. And it would also be the level where many optimizations would
be performed. This includes loop splitting, when an index set is a sum
type, which can be generalized to tiling! And, since we can talk about
mutable buffers, we won't need _any_ custom special primitives ops for that!

This change is a tiny first step in that direction. The code is
effectively dead, but it's good enough to explore some simple ideas. As
a follow-up, I will want to add a flag that conditionally enables this
progressive lowering feature, and then we can start fleshing out the
support in codegen etc.


makefile
```diff
@@ -236,6 +236,15 @@ run-tests/%: tests/%.dx just-build
 run-examples/%: examples/%.dx just-build
 	misc/check-quine $< $(dex) script
 
+lower-tests: export DEX_LOWER=1
+lower-tests: export DEX_ALLOW_CONTRACTIONS=0
+lower-tests: export DEX_TEST_MODE=t
+lower-tests: just-build
+	misc/check-quine tests/lower.dx $(dex) script
+update-lower-tests: just-build
```
axch, line 244: Should we set `DEX_LOWER=1` etc for the `update` target as well?

src/lib/Lower.hs
```diff
@@ -0,0 +1,130 @@
...
+type LowerM = GenericTraverserM LFS
+instance SinkableE LFS where
+  sinkingProofE _ LFS = LFS
+instance HoistableState LFS where
+  hoistState _ _ LFS = LFS
+  {-# INLINE hoistState #-}
+
+instance GenericTraverser LFS where
+  traverseExpr expr = case expr of
+    Hof (For dir ixTy (Lam body)) -> traverseFor Nothing dir ixTy body
```
axch, line 36: Can there be any `For`s whose body is not inlined as a `Lam`?  Do we want to skip those, or assert they are impossible?

src/lib/Lower.hs
```diff
@@ -0,0 +1,130 @@
...
+
+type AtomNameMap (i'::S) (o::S) = M.Map (AtomName i') (Dest o)
+type DestAssignment (i'::S) (o::S) = AtomNameMap i' o
+
+lookupDest :: DestAssignment i' o -> AtomName i' -> Maybe (Dest o)
+lookupDest = flip M.lookup
+
+decomposeDest :: Emits o => Dest o -> Atom i' -> LowerM i o (Maybe (DestAssignment i' o))
+decomposeDest dest = \case
+  Var v -> return $ Just $ M.singleton v dest
```
axch, line 70: I think a TODO comment here about other intended decompositions would help make the rest of the code clearer.

src/lib/Lower.hs
```diff
@@ -0,0 +1,130 @@
...
+traverseExprWithDest dest expr = case expr of
+  Hof (For dir ixTy (Lam body)) -> traverseFor dest dir ixTy body
+  _ -> do
+    expr' <- traverseExprDefault expr
+    case dest of
+      Nothing -> return expr'
+      Just d  -> do
+        ans <- Var <$> emit expr'
+        void $ emitOp $ Place d ans
+        return $ Atom ans
```
axch, line 108: I am confused about the intended relationship between writing a value to a dest and being able to reference that value through its atom as well.  Shouldn't the dest be the source of truth, such that we emit a reconstruction from it here?  Or am I missing something?

axch (approved):
LGTM, but there's one part of the concept I don't understand (which may be for the doc?), about the relationship between writing a value to a dest and returning that value as an atom.  So, feel free to merge, but I would like to understand that.

src/lib/Lower.hs
apaszke: All `For`s always embed a `Lam`. It's the only way to elaborate them from surface syntax, and we never outline functions, so it always holds.

makefile
apaszke: I guess it doesn't matter for the update target, since they should both produce the same results. So it might be an even better test to update using the golden implementation without this pass, but check that the results don't change once we enable it!

But it was an accident, good catch.

src/lib/Lower.hs
apaszke: Good point, I'll add it tomorrow!

src/lib/Lower.hs
apaszke: The dest is where the result of the whole block will be written, but it doesn't have to be the only use site of that value. Consider
```
v0 = x + 2
v1 = v0 * 2
(v0, v1)
```
if we split the destination for a pair into `v0` and `v1` we can't replace `v0 = x + 2` with a `place`, because it would make it unavailable for the computation of `v1`. What you see here effectively lifts the `place` ops from the end of the block to where the equations are. That is, instead of
```
v0 = ...
v1 = ...
place d0 v0
place d1 v1
```
we'll end up with
```
v0 = ...
place d0 v0
v1 = ...
place d1 v1
```

PR #938, opened 2022-06-03
Proposal to change the syntax for "type-annotate expression" to use `::`, not `:`

### LINKED ISSUE #933: The syntax of ":" is super-confusing
What might a line like
```
foo : (Fin 3) => Float
```
mean, in Dex syntax?

To the uneducated eye, it might seem that this is a straightforward case of referencing the variable `foo`, with type annotation `(Fin 3) => Float`.  Indeed, that was the effect I was trying to achieve, for I had defined `foo` as `ones`, a function of type `a:Type ?-> Ix a ?=> a => Float` with all arguments implicit; I was trying to supply them with the type annotation.

Imagine my confused frustration when the above refused to parse.

Trying to understand what was happening, I source-dove the expression parser, and discovered an alternative: Perhaps this is being interpreted as a pi type, namely the array type `(Fin 3) => Float`, where I named the elements of the `(Fin 3)` as `foo` but neglected to reference them?  I tried to forestall that possibility by writing
```
foo : ((Fin 3) => Float)
```
but that did not have the desired effect either.  Nor, in retrospect, could it have had the desired effect, because the parser for pi types commits when it sees a complete annotated binder, without waiting for an arrow.  (Why?)

Further flailing at length brought me to the parser for _decls_, which it took me a long time to think of because I was very strongly primed that I was authoring an expression, whereupon I discovered that _it_ commits to a type-annotated let binding when it sees the `:`, without waiting for the `=` that one would think would be dispositive.

In retrospect, it is possible to forestall that interpretation by wrapping my putative expression in parens, thus:
```
(foo : (Fin 3) => Float)
```

whereupon it does successfully parse --- as a pi type.  In the event, I temporarily modified the decl parser to not commit until the `=`.  But that's still not the parse I want!  What can I do to prevent the pi type interpretation?

After a great deal of additional flailing, I finally thought to check the parser for the thing I was trying to create, namely a variable occurrence.  Surely a string of letters outside a pattern-match context that does not form a keyword is a variable occurrence?  No!  A string of letters outside a pattern match context that does not form a keyword is a variable occurrence _unless followed by a `:`_, because that's how pi type binders are disambiguated!  So in the end, I am able to achieve the desired effect by hiding from the colon on both sides:
```
((foo) : (Fin 3) => Float)
```

Because `(foo)` is definitely an expression (which can be type-annotated), but just `foo` stops being an expression and becomes a pi type binder as soon as you try to type-annotate it.  With any type, even one that cannot be a pi because it doesn't expose any arrows.

Perhaps we should remove the special case that lets `:` retroactively make a name not be a variable, and accept a parser performance regression by making decl commit on `=` and pi commit on arrow?  Or maybe change the syntax of either "type-annotate a binder" or "type-annotate an expression" so they are not the same?  Otherwise we will have to teach this to users over and over: "`:` is the type annotation operator, except if the LHS is a name, in which case the whole thing is a type-annotated binder, whether there is any context that can consume that binder or not."

axch:
Why?  The `:` already means "type-annotate binder", and in particular
serves as the marker that distinguishers a binder from a variable
reference.  For instance,

  foo : (Fin 3) => Float

parses as a pi type for an array whose indices are named `foo`, but
whose elements happen not to depend on the index, even though it is in
scope.  In the situation where the user wants to write a reference
to the variable `foo` and annotate it with the type `(Fin 3) =>
Float`, this change allows them to write

  foo :: (Fin 3) => Float

instead of the unintuitive (to me, at least)

  (foo) : (Fin 3) => Float

that is required by the status quo.

This PR also tweaks the parse rule for `decl` to commit only when it
sees the `=` sign, rather than committing eagerly on `:`.  That way,
we can keep the behavior that

  foo : (Fin 3) => Float = stuff

parses as binding `foo` to `stuff` while annotating `foo` with the
type `(Fin 3) => Float` while simultaneously allowing a line beginning
with

  foo : (Fin 3) => Float

to parse as a pi type instead of trying to parse as a binding of `foo`
and failing.

Committing later is presumably something of a performance regression
for the parser, but I think it only matters in this case, namely
trying to parse

  foo : (Fin 3) => Float

as a binding, then giving up for lack of an `=` sign and re-parsing it
as a pi type.

Fixes #933.

### DIFF
--- a/src/lib/Parser.hs
+++ b/src/lib/Parser.hs
@@ -344,7 +344,7 @@ dataConDefBinder = annBinder <|> (UAnnBinder UIgnore <$> containedExpr)
 
 decl :: Parser (UDecl VoidS VoidS)
 decl = do
-  lhs <- simpleLet <|> funDefLet
+  lhs <- funDefLet <|> (try $ simpleLet <* lookAhead (sym "="))
   rhs <- sym "=" >> blockOrExpr
   return $ lhs rhs
 
@@ -1090,7 +1090,7 @@ limFromMaybe (Just x) = InclusiveLim x
 
 annotatedExpr :: Operator Parser (UExpr VoidS)
 annotatedExpr = InfixL $ opWithSrc $
-  sym ":" $> (\pos v ty -> WithSrcE (Just pos) $ UTypeAnn v ty)
+  sym "::" $> (\pos v ty -> WithSrcE (Just pos) $ UTypeAnn v ty)
 
 inpostfix :: Parser (UExpr VoidS -> Maybe (UExpr VoidS) -> UExpr VoidS)
           -> Operator Parser (UExpr VoidS)
@@ -1208,7 +1208,7 @@ doubleLit = lexeme $
 
 knownSymStrs :: HS.HashSet String
 knownSymStrs = HS.fromList
-  [".", ":", "!", "=", "-", "+", "||", "&&", "$", "&", "|", ",", "+=", ":="
+  [".", ":", "::", "!", "=", "-", "+", "||", "&&", "$", "&", "|", ",", "+=", ":="
   , "->", "=>", "?->", "?=>", "--o", "--", "<<<", ">>>", "<<&", "&>>"
   , "..", "<..", "..<", "..<", "<..<", "?"]
 
--- a/examples/isomorphisms.dx
+++ b/examples/isomorphisms.dx
@@ -44,13 +44,13 @@ that produce isos. We will start with the first two:
 - `#?x` produces a "prism-like" variant matcher `Iso {x:a | ...r} (a | {|...r})`
 
 %passes parse
-:t #b : Iso {a:Int & b:Float & c:Unit} _
+:t #b :: Iso {a:Int & b:Float & c:Unit} _
 > (Iso {a: Int32 & b: Float32 & c: Unit} (Float32 & {a: Int32 & c: Unit}))
 > (MkIso {fwd=(\{b=x, ...r}. (,) x r), bwd=(\(x, r). {b=x, ...r})}
 >  : Iso {a: Int & b: Float & c: Unit} _)
 
 %passes parse
-:t #?b : Iso {a:Int | b:Float | c:Unit} _
+:t #?b :: Iso {a:Int | b:Float | c:Unit} _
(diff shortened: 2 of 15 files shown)


src/lib/Parser.hs
```diff
@@ -344,7 +344,7 @@ dataConDefBinder = annBinder <|> (UAnnBinder UIgnore <$> containedExpr)
 
 decl :: Parser (UDecl VoidS VoidS)
 decl = do
-  lhs <- simpleLet <|> funDefLet
+  lhs <- try $ (simpleLet <|> funDefLet) <* lookAhead (sym "=")
```
dougalm, line 347: I'm worried that this `try` will lead to bad parse error messages, which we don't have good tests for. If you write `def (x:Int) (y: Int) : <some syntax error> =` then the parser will fail without consuming input and we'll fall back to trying to parse it as an expression instead of a decl. I think that at least `def` should commit early. Perhaps simple let bindings can have the long backtrack, because the common case there is just a single binder, `x = ...` so there's not much that can go wrong before you hit `=`.

dougalm (changes requested):
Looks good. My only worry is that the new `try` will produce bad parse errors. Can we at least commit to parsing a decl as soon as we see `def`?

src/lib/Parser.hs
axch: Changed it to try `funDefLet` first, which should commit if it sees the `def`.  On the suggested example of a bad definition, this does preserve the status quo error message, though said error message wasn't so great to begin with.

dougalm (approved):

apaszke:
I'm a bit afraid that this will end up confusing people, because we now use both `::` and `:` to mean "has type". Perhaps we should just drop this type annotation syntax and introduce a function that does it? I think Idris does something along those lines? Here's an example:
```
Float64 `isTypeOf` 2.0 
```

axch:
Hm.  Turning it into a function didn't even occur to me -- an interesting idea.  I think I will go ahead an merge this change as-is, because I think it's an improvement over the status quo, but we can keep that function idea in mind for the future.

dougalm:
I like Adam's idea of removing `UTypeAnn` completely and just using the prelude one-liner `def isTypeOf (a:Type) (x:a) : a = x` instead. Type annotation of expressions (rather than binders) isn't very common even in Haskell, and it's less common in Dex.

PR #939, opened 2022-06-03
Hack checking that the type of the object passed to :html is a String

### LINKED ISSUE #934: Compiler Bug! :html x
The final output of the code asked me to report this as a bug (Although, I wouldn't be surprised if I was making a mistake somewhere).

running `dex web script_name.dx` where script_name contains the following code

```
Height = Fin 3 
Width = Fin 8

x = for i:Height. for j:Width. 1.0
x
```
produces the following error 
```
Parse error:5:2:
  |
5 | x
  |  ^
unexpected end of input
expecting "..", "..<", "<..", "<..<", "=>", '$', ''', '.', ':', '@', '_', alphanumeric character, arrow, backquoted name, end of line, expression, infix operator, or let binding
```
taking a guess as to how to correct this, I tried  the following
```
Height = Fin 3 
Width = Fin 8

x = for i:Height. for j:Width. 1.0
:html x 
```
which produces the following: 
```
Compiler bug!
Please report this at github.com/google-research/dex-lang/issues

Not a string: (view i:(Fin 3) . 
 view i1:(Fin 8) . 
  v:Int32 = %cast Int32 i
  v1:Int32 = %cast Int32 i1
  v2:Int32 = %imul v 8
  v3:Int32 = %iadd v1 v2
  v4:(Ptr (CPU, Float32)) = ptr3 +> v3
  v5:Float32 =
   %runIO
     (\_ign:Unit.
         v5:Float32 = load v4
        v5)
 v5)
CallStack (from HasCallStack):
  error, called at src/lib/Serialize.hs:76:18 in dex-0.1.0.0-6YSqRVo9AYNCATbpRfLD0m:Serialize
```

axch:
'cause having the compiler crash when somebody tries to :html an array
or something is not cool.

Fixes #934.

### DIFF
--- a/src/lib/TopLevel.hs
+++ b/src/lib/TopLevel.hs
@@ -193,13 +193,15 @@ evalSourceBlock' mname block = case sbContents block of
   EvalUDecl decl -> execUDecl mname decl
   Command cmd expr -> case cmd of
     EvalExpr fmt -> do
-      val <- evalUExpr expr
+      annExpr <- case fmt of
+        Printed -> return expr
+        RenderHtml -> return $ addTypeAnn expr $ referTo "String"
+      val <- evalUExpr annExpr
       case fmt of
         Printed -> do
           s <- pprintVal val
           logTop $ TextOut s
         RenderHtml -> do
-          -- TODO: check types before we get here
           s <- getDexString val
           logTop $ HtmlOut s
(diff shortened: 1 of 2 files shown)


dougalm (approved):
Looks good. Not much of a hack at all imo.

apaszke (approved):

PR #944, opened 2022-06-08
Add tutorial data download commands to Makefile

alexreinking:
The data is downloaded precisely as a dependency to the tutorial test
case. Naturallly, this requires an internet connection to succeed, but
one may also place the data files in the expected location manually
(e.g. as GitHub Actions currently does).

Also adds a phony `tutorial-data` target for easy downloading (e.g. if
we want to update the GHA workflows to use this) and adjusts the clean
target appropriately.

### DIFF
--- a/makefile
+++ b/makefile
@@ -219,11 +219,17 @@ doc-example-names = $(example-names:%=doc/examples/%.html)
 
 doc-lib-names = $(lib-names:%=doc/lib/%.html)
 
+t10k-images-idx3-ubyte-sha256 = 346e55b948d973a97e58d2351dde16a484bd415d4595297633bb08f03db6a073
+t10k-labels-idx1-ubyte-sha256 = 67da17c76eaffca5446c3361aaab5c3cd6d1c2608764d35dfb1850b086bf8dd5
+
 tutorial-data = t10k-images-idx3-ubyte t10k-labels-idx1-ubyte
 tutorial-data := $(tutorial-data:%=examples/%)
 
 $(tutorial-data):
 	wget -qO $@.gz http://fashion-mnist.s3-website.eu-central-1.amazonaws.com/$(@F).gz
+	@echo $($(@F)-sha256) $@.gz > $@.sha256
+	sha256sum -c $@.sha256
+	$(RM) $@.sha256
 	gunzip $@.gz
 
 .PHONY: tutorial-data
--- a/makefile
+++ b/makefile
@@ -219,6 +219,18 @@ doc-example-names = $(example-names:%=doc/examples/%.html)
 
 doc-lib-names = $(lib-names:%=doc/lib/%.html)
 
+tutorial-data = t10k-images-idx3-ubyte t10k-labels-idx1-ubyte
+tutorial-data := $(tutorial-data:%=examples/%)
+
+$(tutorial-data):
+	wget -qO $@.gz http://fashion-mnist.s3-website.eu-central-1.amazonaws.com/$(@F).gz
+	gunzip $@.gz
+
+.PHONY: tutorial-data
+tutorial-data: $(tutorial-data)
+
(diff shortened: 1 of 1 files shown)


makefile
```diff
@@ -188,6 +188,17 @@ dexrt-llvm: src/lib/dexrt.bc
 %.bc: %.cpp
 	$(CLANG) $(CXXFLAGS) -DDEX_LIVE -c -emit-llvm $^ -o $@
 
+# --- downloading tutorial data ---
+
+examples/:
+	mkdir -p $@
+
+examples/%.gz: | examples/
+	wget -qO $@ http://fashion-mnist.s3-website.eu-central-1.amazonaws.com/$(@F)
```
apaszke, line 197: That feels like a bit too much automation, don't you think? It's not like every gz file can be obtained from that URL. Is there a nice way to specify this as the recipe for the two files we're using at the moment?

makefile
```diff
@@ -219,6 +219,18 @@ doc-example-names = $(example-names:%=doc/examples/%.html)
 
 doc-lib-names = $(lib-names:%=doc/lib/%.html)
 
+tutorial-data = t10k-images-idx3-ubyte t10k-labels-idx1-ubyte
+tutorial-data := $(tutorial-data:%=examples/%)
+
+$(tutorial-data):
```
alexreinking, line 225: Normally, putting multiple outputs on the LHS of a `:` in make is a bit of a foot-gun, but it's exactly what we want here. When you have "multiple" outputs and it's not a pattern (`%`) rule, Make creates one fresh rule per output, with the same dependencies and commands as in the original definition. When it _is_ a pattern rule, then Make believes that the single set of commands will generate _all_ of the listed files (i.e. true multi-output).

My instinct is normally to use pattern rules since their behavior is less surprising.

makefile
```diff
@@ -219,6 +219,18 @@ doc-example-names = $(example-names:%=doc/examples/%.html)
...
+	wget -qO $@.gz http://fashion-mnist.s3-website.eu-central-1.amazonaws.com/$(@F).gz
```
axch, line 226: Since we're fetching these data over http, would you mind adding a `sha256sum` integrity check?  I know it's kind of annoying, but using an unsigned file off the internet without at least checking that it's the same as it was when the makefile was written gives me the willies.

axch (changes requested):

makefile
alexreinking: @axch - done!

makefile
```diff
@@ -219,6 +219,24 @@ doc-example-names = $(example-names:%=doc/examples/%.html)
...
+t10k-images-idx3-ubyte-sha256 = 346e55b948d973a97e58d2351dde16a484bd415d4595297633bb08f03db6a073
+t10k-labels-idx1-ubyte-sha256 = 67da17c76eaffca5446c3361aaab5c3cd6d1c2608764d35dfb1850b086bf8dd5
+
+tutorial-data = t10k-images-idx3-ubyte t10k-labels-idx1-ubyte
+tutorial-data := $(tutorial-data:%=examples/%)
+
+$(tutorial-data):
+	wget -qO $@.gz http://fashion-mnist.s3-website.eu-central-1.amazonaws.com/$(@F).gz
+	@echo $($(@F)-sha256) $@.gz > $@.sha256
+	sha256sum -c $@.sha256
```
apaszke, line 231: You could probably `sha256sum -c <(echo $($(@F)-sha256) $@.gz) ` to avoid creating the intermediate file

apaszke (approved):

makefile
alexreinking: That's a bash-only thing right? I don't think sh supports that. Do we care?

apaszke: I don't think we care that much, but it's a minor thing. This is ok too

PR #951, opened 2022-06-09
Make buildCase simplify eagerly

alexreinking:
There are many callers to buildCase, not all of which examine the
scrutinee to see if the branch is determined. Yet, not doing so can lead
to code bloat, especially when doing automatic transformations like the
exception-to-maybe pass.

We modify buildCase to handle calling trySelectBranch and specializing
eagerly.

Pair-programmed with @dougalm

### DIFF
--- a/src/lib/Transpose.hs
+++ b/src/lib/Transpose.hs
@@ -186,8 +186,9 @@ transposeExpr expr ct = case expr of
       False -> do
         e' <- substNonlin e
         void $ buildCase e' UnitTy \i vs -> do
+          vs' <- mapM emitAtomToName vs
           Abs bs body <- return $ alts !! i
-          extendSubst (bs @@> map RenameNonlin vs) do
+          extendSubst (bs @@> map RenameNonlin vs') do
             transposeBlock body (sink ct)
           return UnitVal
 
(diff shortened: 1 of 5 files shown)


src/lib/Transpose.hs
```diff
@@ -186,8 +186,9 @@ transposeExpr expr ct = case expr of
       False -> do
         e' <- substNonlin e
         void $ buildCase e' UnitTy \i vs -> do
+          vs' <- mapM emitAtomToName vs
```
apaszke, line 189: Given that there are places where we do depend on the variables being plain names like this one, wouldn't it be better to just keep the old behavior? It's a change that seems independent from the simplification you want to add anyway, right?

apaszke (approved):
LGTM, but the change from AtomName to Atom does regress at least one place, so I'm not sure if it's worth it.

src/lib/Transpose.hs
dougalm: `emitAtomToName` checks if it's already a name, so I think there's no loss in this case. The simplifying version of `buildCase` does need to be able to pass atoms rather than names to the continuation, because that's what you get when you successfully unpack the scrutinee (unless you then emit those atoms to obtain names, which *would* be inefficient). So the change isn't independent.

PR #958, opened 2022-06-13
Fix bug in Algebra conflating indices and ordinals

dougalm:


src/lib/Algebra.hs
```diff
@@ -20,81 +18,101 @@ import Data.Map.Strict hiding (foldl, map, empty, (!))
...
+  => b n l -> Block l -> m n (Abs b Block n)
+hoistDecls b block = do
+  Abs hoistedDecls rest <- liftEnvReaderM $
+    refreshAbs (Abs b block) \b' (Block _ decls result) ->
+      hoistDeclsRec b' Empty decls result
+  ab <- emitDecls hoistedDecls rest
+  refreshAbs ab \b'' blockAbs' ->
+    Abs b'' <$> absToBlockInferringTypes blockAbs'
+
+type Decls = Nest Decl -- TODO: Let's put this in Core. `Nest Decl` is common.
```
apaszke, line 75: Sure, why not do it now?

src/lib/Algebra.hs
```diff
@@ -20,81 +18,101 @@ import Data.Map.Strict hiding (foldl, map, empty, (!))
...
-add :: OrdE mono => PolynomialP mono n -> PolynomialP mono n -> PolynomialP mono n
+-- This is the main entrypoint. Doing polynomial math sometimes lets
+-- us compute sums in closed form. This tries to compute
+-- `\sum_{i=0}^(lim-1) body`. `i`, `lim`, and `body` should all have type `Nat`.
+sumUsingPolys :: (Builder m, Fallible1 m, Emits n)
+              => Atom n -> Abs Binder Block n -> m n (Atom n)
+sumUsingPolys lim (Abs i body) = do
+  Abs i' body' <- refreshAbs (Abs i body) \i' body' ->
+    Abs i' <$> simplifyBlockToBlock body'
+  Abs i'' body'' <- hoistDecls i' body'
```
apaszke, line 52: I feel like both simplification and this hoisting should not be the responsibility of algebra

apaszke (commented):
Mostly LGTM, but I don't understand how we now use algebra and why does it need to do so many things it used to not have to do.

PR #960, opened 2022-06-14
Minor bugfixes and improvements for lowering

apaszke:
With all of those, we're actually able to run all tests and all examples except for `schrodinger`!

#### Fix a bug in destination passing implementation

Destination passing tries to lift the writes to the destination for the
block result up to the places where the written values are defined. But,
not all block results have to be defined in the same block! The old
implementation would fail to insert the necessary writes, leaving us
with uninitialized dests.
 
#### Run type-checks and logs after lowering pass

In particular, enable `%passes lower`.

### DIFF
--- a/src/lib/CheckType.hs
+++ b/src/lib/CheckType.hs
@@ -576,15 +576,15 @@ typeCheckPrimOp op = case op of
       MExtend _ x -> x|:s >> declareEff (RWSEffect Writer $ Just h') $> UnitTy
   IndexRef ref i -> do
     getTypeE ref >>= \case
-      RefTy h (TabTy (b:>IxType iTy _) eltTy) -> do
+      TC (RefType h (TabTy (b:>IxType iTy _) eltTy)) -> do
         i' <- checkTypeE iTy i
         eltTy' <- applyAbs (Abs b eltTy) (SubstVal i')
-        return $ RefTy h eltTy'
+        return $ TC $ RefType h eltTy'
       ty -> error $ "Not a reference to a table: " ++
                        pprint (Op op) ++ " : " ++ pprint ty
   ProjRef i ref -> do
     getTypeE ref >>= \case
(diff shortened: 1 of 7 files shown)


src/lib/Lower.hs
```diff
@@ -89,6 +92,10 @@ traverseBlockWithDest dest (Block _ decls ans) = do
       ans' <- traverseDeclNest decls $ traverseAtom ans
       void $ emitOp $ Place dest ans'
     Just destMap -> do
+      let declsScope = toScopeFrag decls
+      void $ flip M.traverseWithKey destMap \v d -> case decideScope declsScope v of
+        Left vh -> void $ emitOp . Place d =<< substM (Var vh)
```
axch, line 97:
```suggestion
        -- If the variable `vh` was defined outside the block,
        -- place it here so we don't forget to.
        Left vh -> void $ emitOp . Place d =<< substM (Var vh)
```

src/lib/CheckType.hs
```diff
@@ -576,15 +576,15 @@ typeCheckPrimOp op = case op of
...
     getTypeE ref >>= \case
-      RefTy h (TabTy (b:>IxType iTy _) eltTy) -> do
+      TC (RefType h (TabTy (b:>IxType iTy _) eltTy)) -> do
         i' <- checkTypeE iTy i
         eltTy' <- applyAbs (Abs b eltTy) (SubstVal i')
-        return $ RefTy h eltTy'
+        return $ TC $ RefType h eltTy'
       ty -> error $ "Not a reference to a table: " ++
                        pprint (Op op) ++ " : " ++ pprint ty
   ProjRef i ref -> do
```
axch, line 585: Should we make the same change to QueryType?

axch (approved):

apaszke:
I forgot about this PR, but have merged all the fixes in other PRs already 🤦‍♂️

PR #968, opened 2022-06-17
Remove the `IxTy` constructor for `Atom`.

dougalm:
Instead, `For` and `Seq` take the index type and the dict as separate argumnets.

### DIFF
--- a/src/lib/Types/Primitives.hs
+++ b/src/lib/Types/Primitives.hs
@@ -54,7 +54,6 @@ data PrimTC e =
       | EffectRowKind
       | LabeledRowKindTC
       | LabelType
-      | IxTypeKind
         deriving (Show, Eq, Generic, Functor, Foldable, Traversable)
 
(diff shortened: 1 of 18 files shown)


src/lib/Types/Primitives.hs
```diff
@@ -151,7 +150,7 @@ traversePrimOp = inline traverse
 {-# INLINABLE traversePrimOp #-}
 
 data PrimHof e =
-        For ForAnn e e        -- Ix type, body lambda
+        For ForAnn e e e        -- index type, ix dict, body lambda
```
apaszke, line 153: We could probably query the index type from the dict, couldn't we? 🤔

apaszke (approved):
LGTM!

src/lib/Types/Primitives.hs
dougalm: Good point. Just pushed a change to do it that way instead. At the very least it should reduce IR sizes.

PR #985, opened 2022-07-07
Implement a parser for effect and handler definitions.

### LINKED ISSUE #979: Implementing a parser for effect and handler definitions.
## Syntax for effect declarations

The syntax for effect declarations will closely mirror that of `interface` declarations. However, each signature must be preceded by a keyword that indicates how any handler must behave with respect to resumption. These keywords are:

1. `jmp` - the operation never resumes.
2. `def` - the operation tail-resumes exactly once and behaves like a function from the caller's perspective (a "linear" effect)
3. `ctl` - either of the above cases (down the line, can be extended to mean "any number of resumptions")

Specifying the resumption policy in the operation type mimics [Koka](https://koka-lang.github.io/koka/doc/book.html#sec-handlers). However, Koka does not have an analogue to `jmp`.

For example:

```
effect Exn
  jmp raise : a: Type ?-> Unit -> a
```

After @axch's parser rewrite, it will (hopefully) be possible to write a signature that more closely mirrors the `def` keyword:

```
effect Exn
  jmp raise : {a} Unit -> a
```

Effects are not currently parameterized and do not carry state. This will likely need to be relaxed somehow since second-class resumptions cannot simulate state.

## Syntax for handler definitions

Handlers exist at the top level and can be called like functions to handle an effectful expression. Their syntax will mirror that of `instance` definitions. Handlers may be type-parameterized and must specify their return type. For example:

```
handler {r} catch : Exn returning (Maybe r)
  jmp raise = \_. Nothing
  -- jmp raise {a} (_:Unit) = Nothing  -- after parsing changes are merged
  return = \x. Just x
```

The parameter `r` is matched against the type of the expression the handler is handling. The special `return` function must always be defined to lift a value of type `r` to the type given after the `returning` keyword, in this case `Maybe r`. This is the type of the handled expression. Each operation must return a value of the `returning` type or resume in tail position. `resume` shall be a keyword-function that accepts a single parameter of type equal to the _return type of the operation_ as given in the `effect` declaration.

We repeat the `jmp`, `def`, or `ctl` keyword to remind the author of a handler about the resumption policy of the effect operation.

Note: because `raise` returns any type parametrically and does not take any values of that type as an argument, it is impossible to write a resuming handler for `raise`, unless some bottom value is added to Dex.

## Calling a handler

For an example of calling a handler, see:

```
catch do
  for i:(Fin 5).
    if ordinal i > 3
      then raise ()
      else 23
```
The type of the `do` block should be `{Exn} Int` and the instantiated type of `catch` is something like `{Exn|e} Int -> {e} Maybe Int`. This particular example will return `Nothing :: Maybe Int`. If `raise` were never called at runtime (say the `> 3` were instead `> 10`), then the value `23` would be passed to the handler's `return` function.

## New keywords

In all, this proposal adds the following keywords:

* `effect` to declare new effects
  * `jmp` to indicate that an effect operation shall never return
  * `ctl` to indicate that an effect operation might do anything
* `handler` to declare new handlers
  * `return` to define the value-lifting function of a handler
  * `returning` to give the return type of a handler
  * `resume` to call the resumption

alexreinking:
This PR will add parser support for frontend data structures to describe user-defined effects and their handlers. Type inference is out of scope for this PR. 

Draft TODOs:
* [x] This is still missing the `resume` keyword for use in effect handler operation bodies.
* [x] Fix broken tests that use the name `return`

Resolves #979


src/lib/Parser.hs
```diff
@@ -298,6 +300,27 @@ interfaceDef = do
...
+    policy <- resumePolicy
+    v <- anyName
+    ty <- annot uType
+    return (fromString v, UEffectOpType policy ty)
+  return (toNest methodNames, methodTys)
+
+effectDef :: Parser (UDecl VoidS VoidS)
+effectDef = do
+  keyWord EffectKW
+  effName <- upperName <|> symName
```
apaszke, line 320: I think ideally we should allow `anyName`

src/lib/Types/Source.hs
```diff
@@ -366,33 +395,42 @@ instance Pretty (SourceMap n) where
     fold [pretty v <+> "@>" <+> pretty x <> hardline | (v, x) <- M.toList m ]
 
 instance GenericE UVar where
-  type RepE UVar = EitherE5 (Name AtomNameC)    (Name TyConNameC)
-                            (Name DataConNameC) (Name ClassNameC)
-                            (Name MethodNameC)
+  type RepE UVar = EitherE8 (Name AtomNameC)     (Name TyConNameC)
+                            (Name DataConNameC)  (Name ClassNameC)
+                            (Name MethodNameC)   (Name EffectNameC)
+                            (Name EffectOpNameC) (Name HandlerNameC)
```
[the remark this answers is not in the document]
alexreinking: I had previously discussed widening this with @dougalm. Plus, 8 is a power of 2 so it feels nicer 😉

src/lib/Types/Source.hs
```diff
@@ -55,11 +55,14 @@ type UEffect = EffectP (SourceNameOr (Name AtomNameC))
...
- | UClassVar   (Name ClassNameC   n)
- | UMethodVar  (Name MethodNameC  n)
+   UAtomVar     (Name AtomNameC     n)
+ | UTyConVar    (Name TyConNameC    n)
+ | UDataConVar  (Name DataConNameC  n)
+ | UClassVar    (Name ClassNameC    n)
+ | UEffectVar   (Name EffectNameC   n)
+ | UMethodVar   (Name MethodNameC   n)
+ | UEffectOpVar (Name EffectOpNameC n)
+ | UHandlerVar  (Name HandlerNameC  n)
```
alexreinking, line 65: delete

alexreinking: oops, no need

dougalm (approved):

PR #986, opened 2022-07-08
Add support for declaring custom function linearizations

apaszke:
This change adds a new syntax for
```
custom-linearization var expr
```
which declares `expr` to be the expression defining the custom
linearization of function defined by `var`. If the type of `var` is
`a -> b`, then `expr` should have type `a -> (b, a -> b)`. I didn't use
linear arrows, because linearity is a bit sketchy in the surface
language at the moment.

It is however still limited, and is mostly meant to be a prototype
implementation to figure out most annoyances. I will make it more
general in following changes (adding support for nary functions and
polymorphism).

---

This required quite a bit of care around simplification, because now the
identities (names) of functions have meaningful effects on AD, so we
can't just inline them all over the place. As such, we now refrain from
inlining bodies of functions defined at the top level. They only get
inlined when they appear on the lhs of an application, except for when
we're under `linearize` and we know that the function has a custom rule
defined.

Note that this inlining now sometimes has to go through multiple hops.
Consider the following program:
```
f = \x. ...
g = f
g 1.0
```
Before, when we began analyzing the expression `g 1.0`, the environment
roughly looked like this:
```
f @> \x. ...
g @> \x. ...  -- inlined body of f
```
However, after this change, `f` will not get inlined eagerly into the
definition of `g`, and the env will look as follows:
```
f @> \x. ...
g @> f
```
meaning that inlining of `g` into `g 1.0` requires us to go through a
chain of lookups.

The impact of that change on performance is unclear, but I would expect
it to be a net positive. We might end up doing more lookups, but they're
generally cheap, and we generally don't write long chains of function
aliases. At the same time, we actually can gain quite a bit by inlining
less eagerly. If we were to serialize the environment after evaluating
the example program ablve, the new one would have a much more compact
representation than the old.

### DIFF
--- a/src/lib/Linearize.hs
+++ b/src/lib/Linearize.hs
@@ -20,6 +20,8 @@ import MTL1
 import QueryType
 import Util (bindM2)
 import PPrint
+import Types.Core
+import Core
 
 -- === linearization monad ===
 
@@ -319,6 +321,14 @@ linearizeDecls (Nest (Let b (DeclBinding ann _ expr)) rest) cont = do
 linearizeExpr :: Emits o => Expr i -> LinM i o Atom Atom
 linearizeExpr expr = case expr of
   Atom x -> linearizeAtom x
+  App (Var f) xs -> do
+    f' <- substM f
+    lookupCustomRules f' >>= \case
+      Nothing -> error "not implemented"
+      Just (CustomLinearize cl) -> do
+        wts <- forM (toList xs) linearizeAtom
+        (ans, flin) <- fromPair =<< naryApp cl (wts <&> \(WithTangent p _) -> p)
+        return $ WithTangent ans $ naryApp (sink flin) =<< sequence (wts <&> \(WithTangent _ t) -> t)
   App _ _ -> error "not implemented"
   TabApp x idxs -> do
     zipLin (linearizeAtom x) (pureLin $ ListE $ toList idxs) `bindLin`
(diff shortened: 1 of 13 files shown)


src/lib/Linearize.hs
```diff
@@ -319,6 +321,14 @@ linearizeDecls (Nest (Let b (DeclBinding ann _ expr)) rest) cont = do
 linearizeExpr :: Emits o => Expr i -> LinM i o Atom Atom
 linearizeExpr expr = case expr of
   Atom x -> linearizeAtom x
+  App (Var f) xs -> do
+    f' <- substM f
+    lookupCustomRules f' >>= \case
+      Nothing -> error "not implemented"
```
axch, line 327: Is this really "not implemented"?  Could you say what, exactly, is not implemented?

I assume this case mostly doesn't happen because the way an `App (Var f)` would happen here is only if simplify didn't inline the function, and the only way that would happen is if there were rules for it, and right now the only rules are linearization rules; but it would still be good to have a reminder of what this case means.

axch (approved):
Glad to see this!

src/lib/Linearize.hs
apaszke: It's the same message that we had before. Linearization runs on simplified code, so most applications should be gone by now. The only exceptions are: (1) the newly added custom linearizations (2) `@noinline` functions. We still don't support differentiation of (2), which is why it's not implemented.

PR #991, opened 2022-07-20
Add support for symbolic representation of zero tangents in custom linearization rules

apaszke:
When a `custom-linearization-symbolic` directive is used to declare a
custom linearization rule, the linearized function is expected to take
`SymbolicTangent a` as the type of tangents of `a`. This makes it
possible to branch based on the activity of function arguments and use
that information to choose the most efficient linearization rule.


axch (approved):

apaszke:
Yes, agreed! I think that even in the current form, AD will call the rule with concrete constructors, which should get simplified away as soon as they reach a case statement. So we might be getting the specialization we want just out of inlining + simplification as it is today.

PR #994, opened 2022-07-25
Implement dead code elimination

apaszke:
This change reimplements dead code elimination after safer names. I
think that it's the first of the passes that utilize information about
binder usage, so it was a good exercise that will make the
implementation of other similar passes (e.g. inlining) easier.

Thanks to the way we represent bindings in Dex Core, DCE is fundamentally
about traversing the whole IR in search of blocks, and then performing
DCE on every block. I thought about using a generic traversal for that
purpose, but ultimately decided against it. It wasn't a good match,
because it is written with forward-traversals in mind, and requires us
to carry a substitution, which is not necessary in the case of DCE.

The implementation strategy I ended up with is very similar to the way
we generate code for substitutions and free variable computation. Which
makes sense, because DCE also wants to compute free variables. To avoid
repeated traversals of subtrees, we fuse this computation with the
block-searching-traversal. Note that the hoisting component is a bit
unsafe, but we verify the invariants in debug mode (by paying for the
repeated traversals).

I also verified that all the generics get inlined and optimized away, so
this approach ends up yielding code similar to one we would get by
writing explicit `case` expressions over our IR. However, this approach has
the benefit of automatically adapting to any changes in the Core IR.

### DIFF
--- a/src/lib/Core.hs
+++ b/src/lib/Core.hs
@@ -287,6 +287,15 @@ instance (Color c, SinkableE ann, ToBinding ann c) => BindsEnv (BinderP c ann) w
   toEnvFrag (b:>ann) = EnvFrag (RecSubstFrag (b @> toBinding ann')) Nothing
     where ann' = withExtEvidence b $ sink ann
 
+instance (SinkableE ann, ToBinding ann AtomNameC) => BindsEnv (NonDepNest ann) where
+  toEnvFrag (NonDepNest topBs topAnns) = toEnvFrag $ zipNest topBs topAnns
+    where
+      zipNest :: Distinct l => Nest AtomNameBinder n l -> [ann n] -> Nest (BinderP AtomNameC ann) n l
+      zipNest Empty [] = Empty
+      zipNest (Nest b bs) (a:anns) = withExtEvidence b $ withSubscopeDistinct bs $
+        Nest (b:>a) $ zipNest bs $ sinkList anns
+      zipNest _ _ = error "Mismatched lengths in NonDepNest"
+
 instance BindsEnv EffectBinder where
   toEnvFrag (EffectBinder effs) = EnvFrag emptyOutFrag $ Just effs
 
@@ -316,6 +325,12 @@ instance BindsEnv EnvFrag where
 
 instance BindsEnv (RecSubstFrag Binding) where
   toEnvFrag frag = EnvFrag frag mempty
+
+-- This is needed to be able to derive generic traversals over Atoms, and is
+-- ok to be left unimplemented for as long as it's _dynamically_ unreachable.
+-- Since references are only used in Imp lowering, we're generally ok.
+instance BindsEnv DataConRefBinding where
+  toEnvFrag = error "not implemented"
 
 instance (BindsEnv b1, BindsEnv b2)
          => (BindsEnv (PairB b1 b2)) where
--- a/src/lib/Optimize.hs
+++ b/src/lib/Optimize.hs
@@ -4,9 +4,13 @@
 -- license that can be found in the LICENSE file or at
 -- https://developers.google.com/open-source/licenses/bsd
 
-module Optimize (earlyOptimize) where
+{-# LANGUAGE UndecidableInstances #-}
 
+module Optimize (earlyOptimize, optimize) where
+
+import Data.Functor
 import Control.Monad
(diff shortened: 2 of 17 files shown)


src/lib/Types/Core.hs
```diff
@@ -235,6 +234,12 @@ instance GenericB EffectBinder where
   fromB (EffectBinder effs) = LiftB effs
   toB   (LiftB effs) = EffectBinder effs
 
+data BoxPtr n = BoxPtr (Atom n) (Block n)  -- ptrptr, size
+                deriving (Show, Generic)
+
+data NonDepNest ann n l = NonDepNest (Nest AtomNameBinder n l) [ann n]
```
axch, line 240: What here is not supposed to depend on what?  Could you add a code comment?

src/lib/Core.hs
```diff
@@ -287,6 +287,18 @@ instance (Color c, SinkableE ann, ToBinding ann c) => BindsEnv (BinderP c ann) w
...
+instance (SinkableE ann, ToBinding ann AtomNameC) => BindsEnv (NonDepNest ann) where
+  toEnvFrag (NonDepNest topBs topAnns) = toEnvFrag $ zipNest topBs topAnns
+    where
+      zipNest :: Distinct l => Nest AtomNameBinder n l -> [ann n] -> Nest (BinderP AtomNameC ann) n l
+      zipNest Empty [] = Empty
+      zipNest (Nest b bs) (a:anns) = withExtEvidence b $ withSubscopeDistinct bs $
+        Nest (b:>a) $ zipNest bs $ sinkList anns
+      zipNest _ _ = error "Mismatched lengths in NonDepNest"
+
+instance BindsEnv DataConRefBinding where
```
axch, line 299: Why is this instance necessary if it's not implemented?  Under what circumstances might this error occur at runtime?

src/lib/Core.hs
```diff
@@ -317,6 +329,9 @@ instance BindsEnv EnvFrag where
 instance BindsEnv (RecSubstFrag Binding) where
   toEnvFrag frag = EnvFrag frag mempty
 
+instance BindsEnv SuperclassBinders where
```
axch, line 332: Same question here.

axch (changes requested):

src/lib/Core.hs
apaszke: Well it might occur if you even use this method, but at the moment it is guaranteed to be _dynamically_ (but not _statically_) dead. This is because right now we only use it in the code generated for DCE, but at all stages when DCE happens, the `XYZRef` constructors are not allowed to appear in the IR. It's basically an unfortunate effect of putting all pieces that might ever appear in Dex Core in a single AST.

src/lib/Types/Core.hs
apaszke: Will do. In a `Nest b n l` the free vars of a binder half way through the nest can depend on all binders above. In a `NonDepNest` the annotations of each binder cannot depend on the names bound previously. You can think of it as a parallel binding construct, that introduces multiple names into the scope concurrently. But safer names always reasons about linear scope extensions, so we encode it in this way instead.

src/lib/Optimize.hs
```diff
@@ -61,3 +68,165 @@ instance GenericTraverser UTLS where
...
+      below <- dceNest bs' ans'
+      case isPureDecl of
+        False -> return $ ElimFailure b' below
+        True  -> do
+          hoistUsingCachedFVs b' below <&> \case
+            HoistSuccess below' -> ElimSuccess below'
+            HoistFailure _ -> ElimFailure b' below
+    case dceAttempt of
+      ElimSuccess below' -> return below'
+      ElimFailure (Let b' decl') (Abs bs'' ans'') -> do
```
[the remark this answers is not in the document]
apaszke: `b'` is only in scope in the `Abs` that appears here on the right, and the free vars are removed when we exit the `refreshAbs` above. `decl'` cannot refer to `b'`, so even though we traverse it, it will not add _this `b'`_ into the free vars. But I just realized I forgot to add the free vars of `b'` itself (which was ok, because a `NameBinder` never has any free vars).

src/lib/Core.hs
apaszke: I added a comment.

axch (approved):
LGTM.

PR #995, opened 2022-07-25
Avoid specializing on the "data" components of types.

dougalm:
The idea is that we don't want to specialize so aggressively that `Fin 4` and
`Fin 5` yield different specializations. In both cases, we only want to
specialize at `Fin n` "for all n". For now, we've special-cased this to `Fin`,
but we should be able to generalize to any type parameter that has a runtime
representation (i.e. anything we consider "data").

### DIFF
--- a/src/lib/Builder.hs
+++ b/src/lib/Builder.hs
@@ -58,6 +58,7 @@ import Control.Monad.Writer.Strict hiding (Alt)
 import Control.Monad.State.Strict (MonadState, StateT (..), runStateT)
 import qualified Data.Set        as S
 import qualified Data.Map.Strict as M
+import Data.Foldable (toList)
 import Data.Functor ((<&>))
 import Data.Graph (graphFromEdges, topSort)
 import Data.Text.Prettyprint.Doc (Pretty (..), group, line, nest)
@@ -769,7 +770,9 @@ buildNaryLam binderNest body = do
           naryAbsToNaryLam $ Abs bs block
 
 asNaryLam :: EnvReader m => NaryPiType n -> Atom n -> m n (NaryLamExpr n)
-asNaryLam = undefined
+asNaryLam ty f = liftBuilder do
+  buildNaryLamExpr ty \xs ->
+    naryApp (sink f) (map Var $ toList xs)
 
 buildNaryLamExpr
   :: ScopableBuilder m
--- a/src/lib/Builder.hs
+++ b/src/lib/Builder.hs
@@ -21,7 +21,7 @@ module Builder (
   getClassDef, getDataCon,
   Emits, EmitsEvidence (..), buildPi, buildNonDepPi,
   buildLam, buildTabLam, buildLamGeneral,
-  buildAbs, buildNaryAbs, buildNaryLam, buildNullaryLam, buildNaryLamExpr,
+  buildAbs, buildNaryAbs, buildNaryLam, buildNullaryLam, buildNaryLamExpr, asNaryLam,
   buildAlt, buildUnaryAlt, buildUnaryAtomAlt,
   buildNewtype,
   emitDataDef, emitClassDef, emitInstanceDef, emitDataConName, emitTyConName,
@@ -256,16 +256,14 @@ emitImpFunBinding hint f = emitBinding hint $ ImpFunBinding f
 
 emitSpecialization
   :: (Mut n, TopBuilder m)
-  => NameHint -> SpecializationSpec n -> m n (AtomName n)
-emitSpecialization hint specialization = do
- case specialization of
-  AppSpecialization fname args -> do
-    TopFunBound unspecializedTy _ <- lookupAtomName fname
-    specializedTy <- instantiateNaryPi unspecializedTy args
-    let binding = TopFunBound specializedTy $ SpecializedTopFun (Var fname) args
-    name <- emitBinding hint $ toBinding binding
-    extendSpecializationCache specialization name
-    return name
+  => SpecializationSpec n -> m n (AtomName n)
+emitSpecialization s = do
+  let hint = getNameHint s
+  specializedTy <- specializedFunType s
+  let binding = TopFunBound specializedTy $ SpecializedTopFun s
+  name <- emitBinding hint $ toBinding binding
+  extendSpecializationCache s name
+  return name
 
 bindModule :: (Mut n, TopBuilder m) => ModuleSourceName -> ModuleName n -> m n ()
 bindModule sourceName internalName = do
@@ -769,6 +767,9 @@ buildNaryLam binderNest body = do
       Nest (b:>ty) bs ->
         AtomicBlock $ Lam $ LamExpr (LamBinder b ty PlainArrow Pure) $
           naryAbsToNaryLam $ Abs bs block
+
+asNaryLam :: EnvReader m => NaryPiType n -> Atom n -> m n (NaryLamExpr n)
+asNaryLam = undefined
 
 buildNaryLamExpr
   :: ScopableBuilder m
--- a/src/lib/QueryType.hs
+++ b/src/lib/QueryType.hs
@@ -182,7 +182,7 @@ naryLamExprType (NaryLamExpr (NonEmptyNest b bs) eff body) = liftTypeQueryM idSu
     binderToPiBinder (nameBinder:>ty) = PiBinder nameBinder ty PlainArrow
 
 specializedFunType :: EnvReader m => SpecializationSpec n -> m n (NaryPiType n)
-specializedFunType (AppSpecialization ~(Var f) ab) = liftEnvReaderM $
+specializedFunType (AppSpecialization f ab) = liftEnvReaderM $
   refreshAbs ab \extraArgBs (ListE staticArgs) -> do
     let extraArgBs' = fmapNest plainPiBinder extraArgBs
     lookupAtomName (sink f) >>= \case
--- a/src/lib/QueryType.hs
+++ b/src/lib/QueryType.hs
@@ -13,7 +13,7 @@ module QueryType (
   getMethodIndex,
   instantiateDataDef, instantiateNaryPi, instantiateDepPairTy, instantiatePi, instantiateTabPi,
   litType, lamExprTy,
-  numNaryPiArgs, naryLamExprType,
+  numNaryPiArgs, naryLamExprType, specializedFunType,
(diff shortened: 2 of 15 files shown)


src/lib/TopLevel.hs
```diff
@@ -593,10 +590,11 @@ execUDecl mname decl = do
...
 compileTopLevelFun
   :: (Topper m, Mut n)
-  => AtomName n -> NaryPiType n -> Atom n -> [Atom n] -> m n ()
-compileTopLevelFun fname naryPi f staticArgs = do
-  f' <- forceDeferredInlining f
-  fSimp <- simplifyTopFunction naryPi f' staticArgs
+  => AtomName n -> m n ()
+compileTopLevelFun fname = do
+  TopFunBound fTy (SpecializedTopFun s) <- lookupAtomName fname
+  fPreSimp <- specializedFunPreSimpDefinition fTy s
```
apaszke, line 596: It feels weird to do the look up above and then pass in `fTy` and `s` below (what if you pass in a type that doesn't match the specialization?!). Perhaps move the lookup into `specializedFunPreSimpDefinition`?

src/lib/Builder.hs
```diff
@@ -770,6 +768,9 @@ buildNaryLam binderNest body = do
         AtomicBlock $ Lam $ LamExpr (LamBinder b ty PlainArrow Pure) $
           naryAbsToNaryLam $ Abs bs block
 
+asNaryLam :: EnvReader m => NaryPiType n -> Atom n -> m n (NaryLamExpr n)
+asNaryLam = undefined
```
apaszke, line 772: Dead code?

src/lib/QueryType.hs
```diff
@@ -181,6 +181,17 @@ naryLamExprType (NaryLamExpr (NonEmptyNest b bs) eff body) = liftTypeQueryM idSu
...
     binderToPiBinder (nameBinder:>ty) = PiBinder nameBinder ty PlainArrow
 
+specializedFunType :: EnvReader m => SpecializationSpec n -> m n (NaryPiType n)
+specializedFunType (AppSpecialization ~(Var f) ab) = liftEnvReaderM $
+  refreshAbs ab \extraArgBs (ListE staticArgs) -> do
+    let extraArgBs' = fmapNest plainPiBinder extraArgBs
+    lookupAtomName (sink f) >>= \case
+      TopFunBound fTy _ -> do
+        NaryPiType dynArgBs effs resultTy <- instantiateNaryPi fTy staticArgs
+        let allBs = fromJust $ nestToNonEmpty $ extraArgBs' >>> nonEmptyToNest dynArgBs
```
apaszke, line 191: Is that really guaranteed to be empty? What if somebody decorates a piece of data with `@noinline`?

src/lib/Simplify.hs
```diff
@@ -367,18 +350,62 @@ simplifyAtomAndInline atom = confuseGHC >>= \_ -> case atom of
...
+generalizeDataComponentsNest Empty [] = return (Abs Empty (ListE []), [])
+generalizeDataComponentsNest (Nest b bs) (x:xs) = do
+  (ab, newArgs) <- case binderType b of
+    TyKind -> generalizeType x
+    DictTy dTy -> do
+      x' <- generalizeInstance dTy x
+      return (Abs Empty x', [])
+    _ -> error "not implemented"
+  refreshAbs ab \newBs x' -> do
+    Abs bs' UnitE <- applySubst (b @> SubstVal x') $ Abs bs UnitE
```
apaszke, line 379: This looks quadratic. Perhaps carry the SubstFrag in parameters and apply it right before you actually use a binder?

src/lib/Simplify.hs
```diff
@@ -367,18 +350,62 @@ simplifyAtomAndInline atom = confuseGHC >>= \_ -> case atom of
...
+    (abRest, newArgsRest) <- generalizeDataComponentsNest bs' (map sink xs)
+    ListE newArgsRest' <- return $ ignoreHoistFailure $ hoist newBs $ ListE newArgsRest
+    refreshAbs abRest \newBsRest (ListE xs') -> do
+      return (Abs (newBs >>> newBsRest) (ListE (sink x':xs')), (newArgs ++ newArgsRest'))
```
apaszke, line 383: I feel like it would be simpler to just keep accumulating the binders into an RNest that would be passed to recursive calls. Then you would also have a permanent reference to the namespace above the abstracted binders and you could avoid doing the repetitive hoisting of newArgsRest. In general, it seems like this function could be restructured to do a lot less work.

apaszke (changes requested):
Mostly LGTM, but please let's make `generalizeDataComponentsNest` have reasonable performance.

> For now, we've special-cased this to Fin, but we should be able to generalize to any type parameter that has a runtime representation.

But it will not be an easy extension, will it?

apaszke (changes requested):
Mostly LGTM, but please let's make `generalizeDataComponentsNest` have reasonable performance.

> For now, we've special-cased this to Fin, but we should be able to generalize to any type parameter that has a runtime representation.

But it will not be an easy extension, will it?

dougalm:
> But it will not be an easy extension, will it?
We'll see! It'll be tricky enough that I don't want to do it in this PR, but here's my plan for the record (also for my future self because it might be a little while before I come back to it with paternity leave and all).

There are two challenges, (1) generalizing the types themselves and (2) generalizing the dictionaries to match those newly generalized types. For (1) I was planning to just traverse the type, and ask `isData` of each argument to a type constructor. If "yes" then put a new variable in its place, if "no" then traverse it. For (2), I want to do a sort of efficient dictionary synthesis using the existing dictionary as a template. Our function is like `generalizeInstance :: EnvReader m => DictType n -> Atom n -> m n (Atom n)`. We only have instances at this point, so we go over the arguments to the instance. If they're class params themselves then we recur. If they're implicit params then we infer them using ordinary type inference based on the desired (generalized) type.

dougalm:
> let's make generalizeDataComponentsNest have reasonable performance.

Yes, good idea. I was imagining that the list of arguments would be small, but it's a bad habit of mine.

src/lib/Simplify.hs
dougalm: Good idea. I just tried it and it worked out pretty well. It makes the code a bit less clear but it's definitely an efficiency win.

src/lib/Simplify.hs
dougalm: Good idea

src/lib/Builder.hs
dougalm: No, just untested (oops!)

apaszke (changes requested):
Note that `generalizeDataComponentsNestRec` is almost tail recursive now (you'd only need an accumulator for the `args`!

But more importantly, this is completely untested! Can you please check in some test code?

src/lib/Types/Core.hs
```diff
@@ -455,16 +455,19 @@ data AtomBinding (n::S) =
...
    deriving (Show, Generic)
 
 -- TODO: extend with AD-oriented specializations, backend-specific specializations etc.
 data SpecializationSpec (n::S) =
-  AppSpecialization (AtomName n) [Type n]
+  -- The additional binders are for "data" components of the specialization types, like
+  -- `n` in `Fin n`.
+  -- XXX: the `Atom n` is actually always a `Var n`, but we can't use `AtomName
+  -- n` because inference needs a `SubstE AtomSubstVal` instance for zonking. We
+  -- should fix this!
```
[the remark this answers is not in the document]
dougalm: Yes, good idea

src/lib/QueryType.hs
dougalm: The intention is that `@noinline` is only used for functions. If somebody tries to use it with data they should hit the `Not a valid @noinline function type` error in `execUDecl`. I'll add a test for it though.

src/lib/Simplify.hs
```diff
@@ -364,23 +364,37 @@ determineSpecializationSpec fname staticArgs = do
...
+    case ty of
+      TyKind -> generalizeType x'
+      DictTy dTy -> do
+        xGeneral <- generalizeInstance dTy x'
+        return (Abs Empty xGeneral, [])
+      _ -> error "not implemented"
+  ListE newArgs' <- return $ ignoreHoistFailure $ hoist newBs $ ListE newArgs
+  (ab', newerArgs) <- refreshAbs ab \newerBs x' -> do
+    extendSubst (b @> SubstVal x') $
+      generalizeDataComponentsNestRec (Abs bs UnitE) xs (newBs >>> newerBs)
```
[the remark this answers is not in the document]
dougalm: Yep, good idea. I also made the whole thing tail recursive like you suggested.

apaszke (approved):
Looks great! Thanks!

src/lib/Simplify.hs
```diff
@@ -369,31 +370,40 @@ generalizeDataComponentsNest
...
-  ListE newArgs' <- return $ ignoreHoistFailure $ hoist newBs $ ListE newArgs
-  (ab', newerArgs) <- refreshAbs ab \newerBs x' -> do
-    extendSubst (b @> SubstVal x') $
-      generalizeDataComponentsNestRec (Abs bs UnitE) xs (newBs >>> newerBs)
-  return (ab', newArgs' ++ newerArgs)
+  ListE newerArgs' <- return $ ignoreHoistFailure $ hoist newBs $ ListE newerArgs
+  refreshAbs ab \newerBs x' -> do
+    extendSubst (b @> SubstVal x') do
+      let newAccum = ( newBs >>> toRNest newerBs
+                     , fmap sink generalizedArgs `snoc` x'
```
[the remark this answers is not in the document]
dougalm: I can just do the `ListE xs' <- sinkM $ ListE xs` trick for now

src/lib/Util.hs
```diff
@@ -282,6 +283,30 @@ getAlternative :: Alternative m => [a] -> m a
...
+snoc (ReversedList xs) x = ReversedList (x:xs)
+
+emptySnocList :: SnocList a
+emptySnocList = ReversedList []
+
+unsnoc :: SnocList a -> [a]
+unsnoc (ReversedList x) = reverse x
+
+toSnocList :: [a] -> SnocList a
+toSnocList xs = ReversedList $ reverse xs
```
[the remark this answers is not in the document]
dougalm: yes

PR #996, opened 2022-07-26
Refactor evidence fabrication for marker classes (Ext, Distinct, Mut, Emits, EmitsInf)

axch:
Working on the safer names paper makes me both want to and understand enough to clean this up.

Specifically
- `FooEvidence` should actually carry the class `Foo`.
- `withFooEvidence` should just unpack the class `Foo` from the evidence and run the continuation.
- Fabrication has to be done by coercing class dictionaries, because the class has to exist eventually; name this operation `withFabricatedFoo`.
- `withFabricatedFoo` doesn't need a value-level argument, since it's fabricating anyway; rely on type application to disambiguate the type at which the class is to be fabricated.
- Where called for, also provide `fabricatedFoo` to create a fabricated GADT carrying the desired class.

This obviates `ExtWitness` (because now `ExtEvidence` does that job), so delete it (was unused anyway).

### DIFF
--- a/src/lib/Name.hs
+++ b/src/lib/Name.hs
@@ -2577,9 +2577,11 @@ class ExtEnd (n::S)
 
 getExtEvidence :: Ext n l => ExtEvidence n l
 getExtEvidence = ExtEvidence
+{-# INLINE getExtEvidence #-}
 
 absurdExtEvidence :: ExtEvidence VoidS n
 absurdExtEvidence = fabricateExtEvidence
+{-# INLINE absurdExtEvidence #-}
 
 -- We give this one a ' because the more general one defined in Name is the
 -- version we usually want to use.
@@ -2606,8 +2608,6 @@ data ExtEvidence (n::S) (l::S) where
 instance Category ExtEvidence where
   id = ExtEvidence
   {-# INLINE id #-}
-  -- Unfortunately, we can't write the class version of this transitivity axiom
-  -- because the intermediate type would be ambiguous.
   ExtEvidence . ExtEvidence = ExtEvidence
   {-# INLINE (.) #-}
 
--- a/src/lib/Name.hs
+++ b/src/lib/Name.hs
@@ -1551,7 +1551,7 @@ extendTrivialInplaceT
 extendTrivialInplaceT d =
   UnsafeMakeInplaceT \env decls -> do
     let env' = unsafeCoerceE $ extendOutMap env d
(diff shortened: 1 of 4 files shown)


src/lib/Name.hs
```diff
@@ -2600,30 +2576,39 @@ instance (ExtEnd n => ExtEnd l) => Ext n l
 class ExtEnd (n::S)
 
 getExtEvidence :: Ext n l => ExtEvidence n l
-getExtEvidence = FabricateExtEvidence
+getExtEvidence = ExtEvidence
```
apaszke, line 2579: Mark INLINE?

src/lib/Name.hs
```diff
@@ -2600,30 +2576,39 @@ instance (ExtEnd n => ExtEnd l) => Ext n l
...
-data ExtEvidence (n::S) (l::S) = FabricateExtEvidence
+data ExtEvidence (n::S) (l::S) where
+  ExtEvidence :: (Ext n l) => ExtEvidence n l
 
 instance Category ExtEvidence where
-  id = FabricateExtEvidence
+  id = ExtEvidence
   {-# INLINE id #-}
   -- Unfortunately, we can't write the class version of this transitivity axiom
   -- because the intermediate type would be ambiguous.
```
apaszke, line 2610: I feel like this comment is false? We have the trick that allows GHC to infer transitivity

src/lib/Types/Core.hs
```diff
@@ -572,7 +572,7 @@ bindersTypes :: (Distinct l, ProvesExt b, BindsNames b, BindsOneAtomName b)
              => Nest b n l -> [Type l]
 bindersTypes Empty = []
 bindersTypes (Nest b bs) = ty : bindersTypes bs
-  where ty = withExtEvidence b $ withSubscopeDistinct bs $ sink (binderType b)
+  where ty = withExtEvidence (Nest b bs) $ sink (binderType b)
```
apaszke, line 575: nit: perhaps bind the nest to a var in the pattern (`n@(Nest b bs)`)?

apaszke (approved):

src/lib/Name.hs
axch: Well, it's certainly quite old, so it could be stale.  Removed.

src/lib/Name.hs
axch: Done.

PR #998, opened 2022-07-29
Add a loop unrolling optimization

apaszke:
We like to write n-ary helpers such as `split_key`, but they often end
up being used with very little arities (especially binary). In those
simple cases, we end up introducing unnecessary intermediate arrays,
index casts, and array indexing operations, that could have been
simplified to a direct evaluation of the body.

This pass adds a simple form of loop unrolling that tries to counter
those issues, under a potential limited code blowup.

### DIFF
--- a/misc/file-check
+++ b/misc/file-check
@@ -14,7 +14,7 @@ if [[ -z "$FILECHECK" ]]; then
   exit 1
 fi
 
-if ${@:2} $1 | $FILECHECK $1 ; then
+if ${@:2} $1 --outfmt result-only | $FILECHECK $1 ; then
   echo "OK"
   exit 0
 else
--- a/src/lib/GenericTraversal.hs
+++ b/src/lib/GenericTraversal.hs
@@ -36,6 +36,9 @@ class (SinkableE s, HoistableState s) => GenericTraverser s where
   traverseExpr :: Emits o => Expr i -> GenericTraverserM s i o (Expr o)
   traverseExpr = traverseExprDefault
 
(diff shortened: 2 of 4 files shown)


apaszke:
A good way to visualize the difference is to consider the program
```
:p
  [k1, k2] = split_key k
  (k1, k2)
```
which has the following simplified form:
```
v#0:((Fin 2) => Word64) =
  for i:(Fin 2).
    v#0:Nat = %cast Nat i
    v#1:Word64 = %cast Word64 v#0
    v#2:Word64 = threefr.1 0x375f238fcddb151d v#1
    v#2
v#1:Word64 = v#0.(0@Fin 2)
v#2:Word64 = v#0.(1@Fin 2)
(v#1, v#2)
```
that turns into just that after optimizations:
```
v#0:Word64 = threefr.1 0x375f238fcddb151d 0x0
v#1:Word64 = threefr.1 0x375f238fcddb151d 0x1
(v#0, v#1)
```

apaszke:
And of course, the optimized form exposes a nice opportunity for constant folding threefry completely, but that's for later (+ we might want to only do this if we consider the body of threefry "cheap enough").

axch (approved):
I assume this still leaves room for later inlining loops that are only used once, etc; but it seems reasonable to merge now without that.

apaszke:
Right. I think that inlining will have to run before this pass (I don't think we ever want to roll things back into loops). But small loops that haven't been fused with anything (like the ones generated by `split_key`) are still worth unrolling.

PR #1006, opened 2022-08-10
Infer implicit arguments for jax.jit(dex.eval()).

axch:
To wit, we already had code in `dex_call_abstract_eval_with_shape` for
inferring implicit arguments; this change just reuses it in
`dex_apply_lowering` to construct appropriate MLIR constants to pass
as arguments to the compiled function.

The catch is that this whole system relies heavily on the
restrictedness of types that can travel across the Jax-Dex boundary,
to wit, only scalars and rectangular arrays; and on the fact that the
only allowed implicit arguments are the sizes of the dynamic
dimensions of said rectangular arrays.

However, that restriction is already present in the status quo before
this change.

### DIFF
--- a/python/dex/interop/jax/apply.py
+++ b/python/dex/interop/jax/apply.py
@@ -45,29 +45,8 @@ def dex_call_abstract_eval_with_shape(*args, func_atom):
   native_func = get_compiled(func_atom)
   arg_sig = native_func.explicit_argument_signature
   res_sig = native_func.result_signature
-  if len(args) != len(arg_sig):
-    raise RuntimeError(f"Dex function expects {len(arg_sig)} arguments, but was given {len(args)}")
-  if not all(isinstance(arg, jax.core.ShapedArray) for arg in args):
-    raise RuntimeError("Cannot perform evaluation of Dex functions without known shapes")
-  # Check arguments and infer shape parameters
-  shape_vars = {}
-  for i, (arg, b) in enumerate(zip(args, arg_sig)):
-    expected_dtype = np.dtype(b.type.ctype)
-    if arg.dtype != expected_dtype:
-      raise RuntimeError(f"dtype mismatch in arg {i}: expected {expected_dtype}, got {arg.dtype}")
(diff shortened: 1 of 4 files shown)


python/dex/interop/jax/apply.py
```diff
@@ -110,11 +115,28 @@ def dex_apply_lowering(ctx, *args, func_atom):
...
+      ctx.avals_in, native.explicit_argument_signature)
+
+  name_to_mlir_arg = {}
+  for arg, binder in zip(args, native.explicit_argument_signature):
+    name_to_mlir_arg[binder.name] = arg
+
+  mlir_args = []
+  for binder in native.argument_signature:
+    if binder.implicit:
+      if binder.name in shape_vars:
```
apaszke, line 132: Missing an else that raises an error?

python/dex/interop/jax/apply.py
```diff
@@ -110,11 +115,28 @@ def dex_apply_lowering(ctx, *args, func_atom):
...
+        mlir_arg = mlir.ir_constant(
+            IdxRepTy(shape_vars[binder.name]).value,
+            canonicalize_types=False)
+        mlir_args.append(mlir_arg)
+    else:
+      mlir_args.append(name_to_mlir_arg[binder.name])
```
apaszke, line 138: I feel like this is a bit roundabout. Could we basically do this?
```py
shape_vars = ...
name_to_mlir_arg = {}
for b, arg in zip(native.explicit_argument_signature, args):
  name_to_mlir_arg[b.name] = arg
for n, s in shape_vars.items():
  name_to_mlir_arg[n] = mlir.ir_constant(...)
mlir_args = [name_to_mlir_arg[n] for n in native.argument_signature]
```

I think that every implicit (shape) variable will always be an argument of the native function, so there's no loss in making that constants eagerly.

apaszke (approved):
LGTM!

python/dex/interop/jax/apply.py
axch: Good idea!  Done.

PR #1007, opened 2022-08-11
Teach Dex-side implementation of vmap to annotate types and implicit arguments

### LINKED ISSUE #1005: Jax vmap over Dex function with implicit arguments
Consider exporting a dynamically-shaped Dex function to Jax, such as
```
def sqr {n: Nat} (x:(Fin n => Float)) : Fin n => Float =
  for i. x.i * x.i
```
The dynamicity of the shape is captured as an implicit argument giving the size of the array.  So far so good.

Now, if the enclosing Jax computation applies `vmap` to this, the Jax-Dex interop will generate the following expression to implement it, which doesn't compile:
```
\ x0. for i:(Fin 2). sqr x0.i
Type error:Ambiguous type variable

\ x0. for i:(Fin 2). sqr x0.i
                     ^^^^
```

Indeed, as written, this call site of `sqr` does not have enough information to fill in the array size `n`.  Naturally, that has to come from the caller to this anonymous function.

Do we want to try to code the Dex compiler to automatically infer what implicit arguments a lambda expression is missing (implicit implicit arguments?)?  Is there a good way to do that?

Alternately, do we want to adjust the interop to generate an explicit implicit argument?  Along the lines of
```
\{m} (x0:(Fin 2) => (Fin m) => Float). for i:(Fin 2). sqr x0.i
```

axch:
Jax vmap of a Dex primitive used to form an expression like
```
\ x0 x1 x2. for i:(Fin batch_size). f x0.i x1.i x2.i
```
which would not compile if `f` had implicit arguments (e.g., dynamic
array size).

Now we add type annotations to the `xi`, which may mention size
parameters; and also implicit arguments to the lambda form, which bind
said size parameters.  The result is an expression like
```
\ {n m}
  (x:(Fin batch_size) => (Fin n) => Float)
  (y:(Fin batch_size) => (Fin m) => Float)
  (z:(Fin batch_size) => (Fin 4) => Float)
  . for i:(Fin batch_size). f x.i y.i z.i
```
(here `batch_size` is a metavariable standing for the actual batch size).

Fixes #1005.

### DIFF
--- a/python/dex/interop/jax/apply.py
+++ b/python/dex/interop/jax/apply.py
@@ -192,16 +192,33 @@ def dex_call_batched(batched_args, batched_dims, func_atom):
   func_name = func_atom.name
   assert func_name is not None
 
+  # TODO: Make it possible to get the signature without compiling the function
+  native = get_compiled(func_atom)
+  expl_args = native.explicit_argument_signature
+  if len(batched_args) != len(expl_args):
+    raise RuntimeError(f"Dex function expects {len(expl_args)} arguments, but was given {len(batched_args)}")
+
+  batched_args = []
+  batched_dims_it = iter(batched_dims)
+  for binder in native.argument_signature:
+    if binder.implicit:
+      batched_args.append("{" + binder.name + "}")
+    else:
+      ty = binder.type.dex_annotation()
+      if next(batched_dims_it) is not batching.not_mapped:
+        ty = f"(Fin {batch_size}) => {ty}"
+      batched_args.append(f"({binder.name} : {ty})")
+
   # Only index into the arguments which are batched. `i` is the index used for
   # the Dex for loop constructor.
   batched_fn_params = [
-      f"x{param_idx}" if dim is batching.not_mapped else f"x{param_idx}.i"
-      for param_idx, dim in enumerate(batched_dims)
+      binder.name if dim is batching.not_mapped else f"{binder.name}.i"
+      for dim, binder in zip(batched_dims, expl_args)
   ]
 
   # This is the actual batching expression
   batched_fn = module.eval(
-      r"\ " + " ".join(f"x{i}" for i in range(len(batched_args))) + ". "
+      r"\ " + " ".join(batched_args) + ". "
       + f"for i:(Fin {batch_size}). {func_name} "
       + " ".join(batched_fn_params))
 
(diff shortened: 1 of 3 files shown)


python/dex/native_function.py
```diff
@@ -19,6 +19,20 @@
...
+def dex_annotation(ctype):
+  dex_types = {
+      ctypes.c_int64: 'Int64',
+      ctypes.c_int32: 'Int32',
+      ctypes.c_uint8: 'Word8',
+      ctypes.c_uint32: 'Word32',
+      ctypes.c_uint64: 'Word64',
+      ctypes.c_double: 'Float64',
+      ctypes.c_float: 'Float32',
+  }
```
apaszke, line 32: This creates the map from scratch every time you call the function, only to look up one of the elements. Perhaps just make the map global? and inline the expression since then the remaining body is so trivial?

python/dex/interop/jax/apply.py
```diff
@@ -192,16 +192,39 @@ def dex_call_batched(batched_args, batched_dims, func_atom):
...
+
+  def batch_type_annotation(param_idx, binder):
+    if batched_dims[param_idx] is batching.not_mapped:
+      return binder.type.dex_annotation()
+    else:
+      return f"(Fin {batch_size}) => {binder.type.dex_annotation()}"
+  batched_explicit_args = {}
+  for param_idx, binder in enumerate(expl_args):
+    ty = batch_type_annotation(param_idx, binder)
+    batched_explicit_args[binder.name] = f"({binder.name} : {ty})"
```
apaszke, line 209: Is there any reaosn why we can't inline this loop below? explicit_argument_signature is just argument_signature with implicit args filtered out. But you do already have a conditional on those in the loop below.

apaszke: If the issue is that you have to zip `batched_dims` with `explicit_params`, you can just do
`batched_dims_it = iter(batched_dims)` and then consume `next(batched_dims_it)` inside the loop. That gets rid of the extra map indirection and is IMO much easier to understand.

apaszke (approved):

python/dex/interop/jax/apply.py
axch: Yeah, zipping with batched_dims was the issue.  Thanks, that's a reasonably clean way to handle it.

python/dex/native_function.py
axch: Done.

PR #1008, opened 2022-08-12
Support implicit arguments and dynamic shapes under jax jvp in dex.eval.

axch:
Same issue and same solution as vmap: explicitly mention those arguments in generated lambda expressions.

### DIFF
--- a/python/dex/interop/jax/apply.py
+++ b/python/dex/interop/jax/apply.py
@@ -245,33 +245,67 @@ def dex_call_jvp(arg_values, arg_tangents, func_atom):
   Returns:
      A pair of the primal output and the tangent.
   """
-  assert len(func_atom.compile().result_signature) == 1
+  # TODO: Make it possible to get the signature without compiling the function
+  native = get_compiled(func_atom)
+  assert len(native.result_signature) == 1
   num_args = len(arg_values)
   module = func_atom.module.copy()
 
-  # Helper functions to build strings of primal and tangent inputs.
-  def arg_string(prefix):
-    return " ".join(f"{prefix}{i}" for i in range(num_args))
+  # TODO: Support explicit arguments that are not being differentiated wrt
+  # TODO: What if some explicit primal argument occurs in the type annotation of later one?
+  # - If this happens, I have to either retain the name of the primal, or rename the annotation.
+  # - Retaining the name of the primal is a problem, though, because then a constructed tangent name
+  #   might clash with it.  (Solvable by maintaining an explicit name mapping.)
+  implicit_args = []
+  primals = []
+  tangents = []
+  name_to_ty = {}
+  for binder in native.argument_signature:
+    if binder.implicit:
+      implicit_args.append("{" + binder.name + "}")
+    else:
+      annot = binder.type.dex_annotation()
+      p_name = f"p{binder.name}"
+      primals.append(p_name)
+      name_to_ty[p_name] = annot
+      t_name = f"t{binder.name}"
+      tangents.append(t_name)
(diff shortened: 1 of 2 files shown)


python/dex/interop/jax/apply.py
```diff
@@ -245,48 +245,81 @@ def dex_call_jvp(arg_values, arg_tangents, func_atom):
...
+  for binder in native.argument_signature:
+    if binder.implicit:
+      implicit_args.append("{" + binder.name + "}")
+    else:
+      p_name = f"x{binder.name}"
+      primals.append(p_name)
+      name_to_ty[p_name] = binder.type.dex_annotation()
+      t_name = f"u{binder.name}"
+      tangents.append(t_name)
+      name_to_ty[t_name] = binder.type.dex_annotation()
```
apaszke, line 272: nit: perhaps compute the annoation once?

python/dex/interop/jax/apply.py
```diff
@@ -245,48 +245,81 @@ def dex_call_jvp(arg_values, arg_tangents, func_atom):
...
+  tangents = []
+  name_to_ty = {}
+  for binder in native.argument_signature:
+    if binder.implicit:
+      implicit_args.append("{" + binder.name + "}")
+    else:
+      p_name = f"x{binder.name}"
+      primals.append(p_name)
+      name_to_ty[p_name] = binder.type.dex_annotation()
+      t_name = f"u{binder.name}"
```
apaszke, line 270: nit: why are tangents prefixed with u? Perhaps prefix primals with p and tangents with t?

python/dex/interop/jax/apply.py
```diff
@@ -245,48 +245,81 @@ def dex_call_jvp(arg_values, arg_tangents, func_atom):
...
+      tangents.append(t_name)
+      name_to_ty[t_name] = binder.type.dex_annotation()
+
+  def juxt_string(items):
+    return " ".join(items)
 
-  def tuple_string(prefix):
-    return "(" + ", ".join(f"{prefix}{i}" for i in range(num_args)) + ")"
+  def juxt_arg_string(items):
```
apaszke, line 277: nit: items -> names? and the same in the functions below

python/dex/interop/jax/apply.py
```diff
@@ -245,48 +245,81 @@ def dex_call_jvp(arg_values, arg_tangents, func_atom):
...
+    annotated = [f"({item} : {name_to_ty[item]})" for item in items]
+    return juxt_string(annotated)
+
+  def tuple_arg_string(items):
+    ty = tuple_ty_ref_string([name_to_ty[item] for item in items])
+    return f"({tuple_ref_string(items)} : {ty})"
+
+  def tuple_ref_string(items):
```
apaszke, line 285: What does "ref" mean here?

apaszke (approved):

python/dex/interop/jax/apply.py
axch: Done.

python/dex/interop/jax/apply.py
axch: Because that's how it was in the old code, but sure.  Changed.

python/dex/interop/jax/apply.py
axch: Done.

python/dex/interop/jax/apply.py
axch: Variable reference as opposed to formal parameter.  Could change it to "var", but this seems ok.

PR #1010, opened 2022-08-12
Support implicit arguments under Jax vjp.

axch:
Same issue and same as vmap and jvp: explicitly mention them as
arguments to generated lambda expressions.

### DIFF
--- a/python/dex/interop/jax/apply.py
+++ b/python/dex/interop/jax/apply.py
@@ -272,23 +272,6 @@ def dex_call_jvp(arg_values, arg_tangents, func_atom):
       tangents.append(t_name)
       name_to_ty[t_name] = annot
 
-  def juxt_string(names):
-    return " ".join(names)
-
-  def juxt_arg_string(names):
-    annotated = [f"({name} : {name_to_ty[name]})" for name in names]
-    return juxt_string(annotated)
-
-  def tuple_arg_string(names):
-    ty = tuple_ty_ref_string([name_to_ty[name] for name in names])
-    return f"({tuple_ref_string(names)} : {ty})"
-
-  def tuple_ref_string(names):
-    return "(" + ", ".join(names) + ")"
-
-  def tuple_ty_ref_string(names):
-    return "(" + " & ".join(names) + ")"
-
   # Add the current function atom as a variable in the context, so that we can
   # use it to apply jvp.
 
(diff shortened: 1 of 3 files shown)


python/dex/interop/jax/apply.py
```diff
@@ -361,16 +361,22 @@ def dex_call_evaluate_linearized_transpose(cotangents, *args, func_atom):
...
-  primals, tangents = args[:num_primals], args[num_primals:]
-
-  # Helper functions to build strings of primal and tangent inputs.
-  def arg_string(prefix, index_set):
-    return " ".join(f"{prefix}{i}" for i in index_set)
+  implicit_args = []
+  name_to_ty = {}
+  for binder in native.argument_signature:
+    if binder.implicit:
+      implicit_args.append("{" + binder.name + "}")
```
apaszke, line 373: perhaps check that `not name_to_ty`? Otherwise you will change the order of implicit and explicit args

apaszke (approved):

python/dex/interop/jax/apply.py
axch: Thought about it, and added (i) a simple test case where it changes the order harmlessly, and (ii) a comment explaining why I think such reordering is safe in all cases that won't also break elsewhere in the interop.

apaszke: I agree that this is harmless right now, but this is only because of the current limitations of export. It's not a fundamental restriction though, and I can imagine having us lift it at some point. While I agree that disallowing that permutation might be too strong, the right thing to do here would be to do hoisting checks (intersect the free vars of type annotations of implicit binders with the explicit binders).

axch: Added hoisting checks.

python/dex/interop/jax/apply.py
```diff
@@ -339,38 +322,79 @@ def tuple_ty_ref_string(names):
...
+  # infer anyway).  As a defensive measure, we perform explicit
+  # hoisting checks when doing the reordering.
+  implicit_args = []
+  name_to_ty = {}
+  for binder in native.argument_signature:
+    if binder.implicit:
+      if hoistable(binder.type, name_to_ty.keys()):
+        implicit_args.append("{" + binder.name + "}")
+      else:
+        raise RuntimeError(f"Hoist check failed: implicit {binder.name} of type {binder.type} depends on a previous explicit binder")
```
apaszke, line 391: A NotImplementedError might have been better, since it's not user's fault.

PR #1011, opened 2022-08-17
Add support for partial destination passing

apaszke:
Our destination passing implementation used to only work with full
destinations. Which is generally fine, for as long as the result atoms
of blocks mention bare variables. However, many functions such as
`yieldAccum` end up with a projection in the result, which completely
trips up our method and results in many unnecessary copies.

In this patch I add a "projected dest" and make sure that we can e.g.
inherit a destination for an accumulator from the result of its
surrounding for loop.


src/lib/Lower.hs
```diff
@@ -84,72 +84,120 @@ traverseFor maybeDest dir ixDict lam@(LamExpr (LamBinder ib ty arr eff) body) =
...
       ans' <- traverseDeclNest decls $ traverseAtom ans
-      place dest ans'
+      place (FullDest dest) ans'
+      return ans'
     Just destMap -> do
       s <- getSubst
       case isDistinctNest decls of
         Nothing -> error "Non-distinct decls?"
         Just DistinctBetween -> do
           -- Note that we ignore ans! Its components are written inplace through destMap.
```
axch, line 128: It seems this comment is now incorrect?

axch (commented):
Well, I tried to read through this, but couldn't really follow what's happening.

Can we (create and) reference a writeup of the destination passing design in these traversals, so they're easier to follow?

(But my ignorance may not be a reason not to merge the PR.)

apaszke:
I added some comments. Let's talk when we meet and try to write up more docs!

PR #1014, opened 2022-08-20
A reasoned proposal for operator precedence in Dex.

axch:
As a side-effect of working on the parser reorganization #1013, I
spent some effort on thinking about operator precedence.  The result
is a proposal to adjust Dex's operator precedence table to the
structure and for the reasons given in this document.

The parser reorganization partially implements this propsal, because
it makes various formerly-special syntactic entities (like the equals
character `=` and the pipe character `|`) be operators subject to
precedence rules.

However, this proposal goes beyond the changes in the parser reorg.
Mainly, I propose defining explicit (and distinct) precedences for a
large number of oprators that are currently defined in user-space in
the prelude and not mentioned in the parser at all; and I also propose
changing the precedence of several extant operators relative to each
other.

The proposal document is written in the present tense so that if
accepted it can just become documentation; but everything in it
(especially the rationales that use the word "arbitrary") is meant, at
this stage, as a proposal open to discussion.

I will interpret a "LGTM" review on this PR as support for the syntax
changes proposed (as opposed to simply code-reviewing the proposal
document qua proposal).

### DIFF
--- a/doc/syntax-philosophy.md
+++ b/doc/syntax-philosophy.md
@@ -1,0 +1,276 @@
+# Dex Syntax Overview
+
+Dex has a lot of syntax.  It can be hard to keep track of, but to make it easier
+Dex tries to be uniform about its grouping structure.  For instance, the colon
+`:` means "I have a name (or pattern) on my left, whose type is on my right".
+But sometimes that name is a variable binding, sometimes it's a function
+argument, and sometimes it's a record field name.  However, in all of these
+cases, the binding precedence of `:` is consistent (and quite loose), so you can
+tell what other operators are part of the type vs part of the enclosing context
+just by parsing the precedence and the grouping parens, without having to
+remember the enclosing syntactic form.
+
+In fact, Dex achieves this with a two-stage parser: Dex resolves keywords,
+grouping, and operator precedence first, uniformly, and then constructs its
+abstract syntax from that.  That's why you see two different kinds of errors.  A
+"Parse error" comes from the first pass, effectively inserting parentheses in
+the right places around all the infix operators.  At this stage, Dex detects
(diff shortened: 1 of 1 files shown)


doc/syntax-philosophy.md
```diff
@@ -0,0 +1,276 @@
...
+  - The standard comparisons `>`, `<`, `>=`, `<=` are given the same precedence
+    by convention
+
+  - The equality operators `==` and `!=` get the next precedence, looser than
+    comparison because `4 < 5 == True` is a reasonable thing to write (if a bit
+    redundant)
+
+3. Boolean arithmetic follows, because it can consume the results of comparisons
+
+  - `&&` is tighter than `||` because the or-of-ands form for Boolean formulas
```
apaszke, line 117: Wouldn't it be better to make them have the same precedence and error if not explicitly parenthesized?

apaszke (approved):
I don't have any strong opinions on precedence, so the change sounds good to me. I'm not sure how helpful this actually is to newcomers, but it certainly doesn't hurt.

Something I'd really like to see, are precedence rules defined in the user language, not in the parser. But it's far from the critical path, so no need to do now.

doc/syntax-philosophy.md
axch: No?  `2 < x && x < 5 || x == 7` seems like a perfectly sensible expression?

alexreinking: I also read `&&` like multiplication and `||` like addition. 🤷

apaszke: I feel like it's better to be explicit in cases like this one, but I guess this precedence is common in other languages, so I won't veto it.

dougalm (approved):
Beautiful! Very well argued from first principles.

PR #1017, opened 2022-08-21
Add Binomial and Poisson samplers.

emilyfertig:
(As discussed with @axch )

The Poisson sampler uses the same algorithm TFP uses for low rate: https://github.com/tensorflow/probability/blob/main/tensorflow_probability/python/distributions/poisson.py#L439


axch (approved):
Thanks @emilyfertig!

It would be nice to eventually get the rejection samplers for the large-argument cases, but we can start here.

PR #1021, opened 2022-08-23
Add a newtype constructor

apaszke:
The Core IR has quite a bit of redundancy in its representations.
Records, one constructor ADTs and products are all... just products,
with different types. And we should represent them that way!

This change adds a "newtype" atom constructor that can take a single
value payload to construct another type. At the moment it is only used
for `Fin` values (letting us remove the `FinVal` constructor). `Nat`s
(as a type different from Word32) are the next step.

### DIFF
--- a/src/lib/Imp.hs
+++ b/src/lib/Imp.hs
@@ -473,11 +473,6 @@ toImpOp maybeDest op = case op of
       (BaseTy _, BaseTy bt) -> do
         x' <- fromScalarAtom x
         returnVal =<< toScalarAtom =<< cast x' bt
-      (TC (Fin _), IdxRepTy) -> do
-        let Con (FinVal _ xord) = x
-        returnVal xord
-      (IdxRepTy, TC (Fin n)) ->
-        returnVal $ Con $ FinVal n x
       _ -> error $ "Invalid cast: " ++ pprint sourceTy ++ " -> " ++ pprint destTy
   BitcastOp destTy x -> do
     case destTy of
@@ -1045,7 +1040,7 @@ makeDestRec idxs depVars ty = confuseGHC >>= \_ -> case ty of
     ProdType tys  -> (Con . ConRef) <$> (ProdCon <$> traverse rec tys)
     Fin n -> do
       x <- rec IdxRepTy
-      return $ Con $ ConRef $ FinVal n x
+      return $ Con $ ConRef $ Newtype (TC $ Fin n) x
     _ -> error $ "not implemented: " ++ pprint con
   _ -> error $ "not implemented: " ++ pprint ty
   where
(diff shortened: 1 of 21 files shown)


src/lib/Simplify.hs
```diff
@@ -729,7 +729,6 @@ simplifyOp op = case op of
     return $ Con $ Lit $ Int32Lit $ fromIntegral val
   CastOp (BaseTy (Scalar Word32Type)) (Con (Lit (Word64Lit val))) ->
     return $ Con $ Lit $ Word32Lit $ fromIntegral val
-  CastOp IdxRepTy (Con (FinVal _ i)) -> return i
```
axch, line 732: Is there a corresponding simplification for newtypes that should appear here?

src/lib/QueryType.hs
```diff
@@ -374,6 +374,8 @@ instance HasType Atom where
                                      (Var v') -> return v'
                                      _ -> error "!?"
                                    instantiateDepPairTy t (ProjectElt (0 NE.:| is) v')
+        -- Newtypes
```
axch, line 377: I don't understand this comment.  Is it saying that TC (Fin _) is possible because of newtypes?  Also, should there be an arm for newtypes in this case statement?

src/lib/Interpreter.hs
```diff
@@ -209,10 +209,6 @@ evalOp expr = mapM evalAtom expr >>= \case
     let failedCast = error $ "Cast not implemented: " ++ pprint sourceTy ++
                              " -> " ++ pprint destTy
     case (sourceTy, destTy) of
-      (IdxRepTy, TC (Fin n)) -> return $ Con $ FinVal n x
-      (TC (Fin _), IdxRepTy) -> do
-        let Con (FinVal _ ord) = x
-        return ord
```
axch, line 215: Why are these no longer needed?

src/lib/Imp.hs
```diff
@@ -473,11 +473,6 @@ toImpOp maybeDest op = case op of
       (BaseTy _, BaseTy bt) -> do
         x' <- fromScalarAtom x
         returnVal =<< toScalarAtom =<< cast x' bt
-      (TC (Fin _), IdxRepTy) -> do
-        let Con (FinVal _ xord) = x
-        returnVal xord
-      (IdxRepTy, TC (Fin n)) ->
-        returnVal $ Con $ FinVal n x
```
axch, line 480: Why are these no longer needed?

axch (approved):
Sadly, I can't say I quite follow this change, but it looks like an improvement.  I left a couple of questions which would help me understand it.

src/lib/Simplify.hs
apaszke: Nope. We no longer need to explicitly cast the payload out of a newtype, because I made it so that it can be a projection (`ProjectElt [0] newtype_atom_name` is an atom that represents the payload).

src/lib/Interpreter.hs
apaszke: You no longer need a `CastOp` to construct a `Fin` value: it is now `Con (Newtype (TC (Fin n)) ordinal_atom`. And to extract the payload you use the projection, as explained in the comment above.

src/lib/Imp.hs
apaszke: As above.

src/lib/QueryType.hs
apaszke: This is meant to query the type of a projection. Before this change, projections out of `Fin` were not allowed and you had to use a `CastOp`. Now, I allowed projections out of newtypes, and `Fin` is currently the only newtype we support.

PR #1022, opened 2022-08-23
Clarify `if` syntax

axch:
Both in a new documentation notebook that explains with words and examples what the syntax is, and in a better error message that detects misplaced `else` clauses and tells the programmer how to fix them.


apaszke (approved):
I love the neat error messages!

PR #1025, opened 2022-08-24
Allow underscore and single quote in Dex type signature parser

### LINKED ISSUE #1024: Calling Dex from Jax fails when Dex parameter name has an underscore
Calling
```
dex_sqr = primitive(dex.eval(
    r'\{n} foo_bar:((Fin n) => Float). for i. foo_bar.i * foo_bar.i'))
```
produces
```
AssertionError: (b'?n:u32,foo_bar:f32[n]', 10, ':')

python/dex/native_function.py:233: AssertionError
```

axch:
Fixes #1024.

### DIFF
--- a/python/tests/dexjit_test.py
+++ b/python/tests/dexjit_test.py
@@ -1,0 +1,128 @@
+# Copyright 2022 Google LLC
+#
+# Use of this source code is governed by a BSD-style
+# license that can be found in the LICENSE file or at
+# https://developers.google.com/open-source/licenses/bsd
+
+import unittest
+import numpy as np
+from functools import partial
+from contextlib import contextmanager
+
+import jax
+import jax.numpy as jnp
+from jax import lax
+from jax.experimental import enable_x64
+
+from dex.interop.jax import dexjit
+
+def lax_test(prim, arg_thunk, **kwargs):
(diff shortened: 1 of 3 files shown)


python/tests/dexjit_test.py
```diff
@@ -0,0 +1,128 @@
...
+            lax_test(prim, lambda: (rn(10, 10), rn())))
+check_broadcasting_pointwise(lax.add, full=True)
+check_broadcasting_pointwise(lax.mul)
+check_broadcasting_pointwise(lax.sub)
+check_broadcasting_pointwise(lax.div)
+check_broadcasting_pointwise(lax.max)
+
+
+if __name__ == "__main__":
+  unittest.main()
```
apaszke, line 128: Any particular reason to split off this file?

apaszke (approved):

python/tests/dexjit_test.py
axch: Theme -- calling Dex from Jax and transpiling Jax to Dex register as quite different interop mechanisms (in particular at quite different stages of completeness), so it seems sensible to have their tests live in different files (especially as the amount of code therein has been growing).

PR #1026, opened 2022-08-25
Render and index all available Dex documentation

axch:
All tested examples and explicit documentation are already rendered; this PR also auto-detects and renders all bundled Dex libraries.

In addition, this PR auto-generates an index file that links to all of the above renderings, so we do not force curious users (and search engines) to do URL surgery to discover them.

### DIFF
--- a/misc/build-web-index
+++ b/misc/build-web-index
@@ -1,0 +1,49 @@
+#!/bin/bash
+
+# This script constructs a Markdown file (on standard output) that constitutes
+# an index of all web-rendered pages of Dex examples, documentation, and
+# libraries.
+
+# The rendered HTML of the index is meant to be placed at the root of the Dex
+# page tree, so that relative links work.
+
+# The script accepts three arguments:
+# - One argument containing a space-separated list of documentation files
+# - One argument containing a space-separated list of example files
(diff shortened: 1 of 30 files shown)


misc/build-web-index
```diff
@@ -0,0 +1,49 @@
...
+ | sed -e 's/^\(.*\).dx/- [\1.dx](\1.html)/' \
+ | sed -e "s/'#//"
+
+echo ""; echo "## Libraries"; echo ""
+
+echo "- [lib/prelude.dx](prelude.html): The Dex Prelude (automatically imported)"
+
+grep -F "'# " $libraries \
+ | sed -e 's/^\(.*\).dx/- [\1.dx](\1.html)/' \
+ | sed -e "s/'#//"
```
apaszke, line 49: Wouldn't it be better to write a Python script? This is neat, but looks a little difficult to maintain and modify.

apaszke (approved):

misc/build-web-index
axch: Yeah, we could write a Python script.  I defaulted to shell because it involves less ceremony, but if the maintenance gets annoying we can certainly move to Python.

PR #1027, opened 2022-08-25
Print full localhost URL

adrhill:
Small quality of life improvement:
When running `dex web`, this should make the localhost URL clickable in most terminals, opening a browser tab to `localhost:8000`.

I'm not super familiar with Haskell, so I hope this is a valid string.


axch (approved):
Thanks for the tweak!  Yes, this is a valid Haskell string (escapes are `\`-prefixed, like most programming languages).

PR #1029, opened 2022-08-25
Generalize brownian motion example to arbitrary dimensions, and add text

duvenaud:
The `brownian_motion.dx` example is currently an optimized version of the virtual Brownian tree algorithm from [Scalable gradients for SDEs](https://arxiv.org/pdf/2001.01328.pdf).  However as it stands it has no text explanation.  This PR adds a bit of background and some comments.

It also extends the algorithm to work anywhere on the real line, and in arbitrarily-many input dimensions.  This is based on currently-unpublished research.  However, since I used Dex to check intermediate results as it was derived, I have this demo ready to go before the paper.

Thanks to @dougalm for writing the first version of this algorithm (which is a faster variant of what's in the paper), and for suggesting the use of typeclasses to implement recursion over input dimensions.

### DIFF
--- a/examples/brownian_motion.dx
+++ b/examples/brownian_motion.dx
@@ -106,12 +106,14 @@ xs = linspace pixels 0.3 2.1
 tolerance = xs.(1@_) - xs.(0@_)
 fs = map (sample_brownian_motion tolerance randn (new_key 0)) xs
 :html show_plot $ xy_plot xs fs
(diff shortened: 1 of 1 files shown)


axch (approved):
Thanks for the nice writeup!

apaszke:
This is awesome. Thanks David!

PR #1034, opened 2022-08-26
Some notes about the seq lowering pass from talking with Adam

axch:


src/lib/Lower.hs
```diff
@@ -30,13 +30,45 @@ import Util (foldMapM)
...
+-- this pass, and the `Seq`, `AllocDest`, `Place`, `RememberDest`, and
+-- `Freeze` constructors appear.
+--
+-- All the `Place`s in the resulting IR are non-conflicting: every
+-- array element is only written to once.
+--
+-- The traversal for a block or subexpression returns an `Atom`
+-- representing the returned value.  If a destination was passed in,
+-- the traversal is also responsible for arranging to write that value
+-- into that destination (possibly by forwaring (a part of) the
```
duvenaud, line 58: typo: "forwaring"

apaszke (approved):

PR #1037, opened 2022-08-30
Tweak the quine script to suppress differences due to Dex timing pragmas

### LINKED ISSUE #910: Define policy for %time and %bench in examples
It seems useful to include `%time` and `%bench` commands in committed Dex examples, both to show users those commands exist, and to show how fast (or slow) Dex is.

However, and example like that cannot be added to the `check-quine` test suite, because the latter will never see the same timing as when the example was committed, and thus will always fail.

What should we do?
- Avoid automatically checking such examples?
- Add a --conceal-timings=False flag to Dex to make quine tests ignore timings?
- Define some clever scheme to parse saved timing information and use it to prevent performance regressions?  (Though, need to be robust to different CPUs and loading levels.)

Speaking of which, do we trust the automatic example compilation Action to provide decent resources for timings?

axch:
Also remove the Levenshtein distance example from the block-list,
because it can now safely be quine tested.

Fixes #910.

### DIFF
--- a/.github/workflows/haskell-ci.yaml
+++ b/.github/workflows/haskell-ci.yaml
@@ -24,7 +24,7 @@ jobs:
             install_deps: brew install llvm@12 pkg-config wget gzip coreutils
             path_extension: $(brew --prefix llvm@12)/bin
           - os: ubuntu-20.04
-            install_deps: sudo apt-get install llvm-12-tools llvm-12-dev pkg-config wget gzip
+            install_deps: sudo apt-get install llvm-12-tools llvm-12-dev pkg-config wget gzip wamerican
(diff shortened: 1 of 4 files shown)


apaszke (approved):

apaszke:
Seems like the size of the dictionary varies between systems, so maybe let's skip printing it?

axch:
Done.

axch:
Actually, it turns out that the dictionary on MacOS is significantly different from Ubuntu, such that most of the words found also change.  Instead of trying to code a dictionary-agnostic version of this demo, I'm just going to take it out of the MacOS tests.

apaszke (approved):

PR #1038, opened 2022-08-31
Use newtypes to represent variants

apaszke:
Just like with records. Variants do not add extra expressive power and
degenerate to sum types once the set of cases is statically known.

### DIFF
--- a/src/lib/Simplify.hs
+++ b/src/lib/Simplify.hs
@@ -15,9 +15,10 @@ module Simplify
 import Control.Category ((>>>))
 import Control.Monad
 import Control.Monad.Reader
+import Data.Maybe
+import Data.List (elemIndex)
 import Data.Foldable (toList)
 import Data.Text.Prettyprint.Doc (Pretty (..), hardline)
-import qualified Data.Map.Strict as M
 import qualified Data.List.NonEmpty as NE
 import qualified Data.Set as S
 import GHC.Exts (inline)
@@ -498,10 +499,6 @@ simplifyAtom atom = confuseGHC >>= \_ -> case atom of
   RecordTy _ -> substM atom >>= cheapNormalize >>= \atom' -> case atom' of
     StaticRecordTy items -> StaticRecordTy <$> dropSubst (mapM simplifyAtom items)
     _ -> error $ "Failed to simplify a record with a dynamic label: " ++ pprint atom'
-  Variant types label i value -> do
-    types' <- fromExtLabeledItemsE <$> substM (ExtLabeledItemsE types)
-    value' <- simplifyAtom value
-    return $ Variant types' label i value'
   VariantTy (Ext items ext) -> VariantTy <$> do
     items' <- mapM simplifyAtom items
     ext' <- liftM fromExtLabeledItemsE $ substM $ ExtLabeledItemsE $ Ext NoLabeledItems ext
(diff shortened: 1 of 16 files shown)


src/lib/Simplify.hs
```diff
@@ -698,38 +695,49 @@ simplifyOp op = case op of
...
       -- Emit a case statement (ordered by the arg type) that lifts the type.
       buildCase right (VariantTy fullRow) \caseIdx [v] -> do
-          let (label, i) = labels !! caseIdx
-          let idx = case M.lookup label litems of Nothing  -> i
-                                                  Just tys -> i + length tys
-          let fullRow' = fromExtLabeledItemsE $ sink $ ExtLabeledItemsE fullRow
-          return $ Variant fullRow' label idx v
+        -- TODO: This is extremely slow! Optimize this! Use a prefix sum?
+        let (label, i) = labels !! caseIdx
+        let idx = fromJust $ elemIndex (label, i + length (lookupLabel leftTys label)) fullLabels
```
axch, line 708: Why can't we compute idx the same way as before?

src/lib/Simplify.hs
```diff
@@ -698,38 +695,49 @@ simplifyOp op = case op of
...
+      let fullLabels = toList $ reflectLabels fullTys
       let labels = toList $ reflectLabels rightTys
       -- Emit a case statement (ordered by the arg type) that lifts the type.
       buildCase right (VariantTy fullRow) \caseIdx [v] -> do
-          let (label, i) = labels !! caseIdx
-          let idx = case M.lookup label litems of Nothing  -> i
-                                                  Just tys -> i + length tys
-          let fullRow' = fromExtLabeledItemsE $ sink $ ExtLabeledItemsE fullRow
-          return $ Variant fullRow' label idx v
+        -- TODO: This is extremely slow! Optimize this! Use a prefix sum?
```
axch, line 706: Could you spell out what's slow and how to make it better at one more level of detail?

src/lib/Simplify.hs
```diff
@@ -698,38 +695,49 @@ simplifyOp op = case op of
...
+        let (label, i) = labels !! caseIdx
+        let idx = fromJust $ elemIndex (label, i + length (lookupLabel leftTys label)) fullLabels
+        let fullRow'@(NoExt fullRowItems') = fromExtLabeledItemsE $ sink $ ExtLabeledItemsE fullRow
+        return $ Con $ Newtype (VariantTy fullRow') $ SumVal (SumTy $ toList fullRowItems') idx v
     _ -> error "not a variant"
-  VariantSplit leftTys@(LabeledItems litems) full -> getType full >>= \case
-    VariantTy (NoExt fullTys@(LabeledItems fullItems)) -> do
+  VariantSplit leftTys full -> getType full >>= \case
+    VariantTy (NoExt fullTys) -> do
+      -- TODO: This is extremely slow! Optimize this!
```
axch, line 714: Here, too.

axch (approved):

src/lib/Simplify.hs
apaszke: Because the idx we had before was an index _within_ that given label, while the idx we have now is an index across labels (it's an index to an entry in `toList labeledTypes`).

src/lib/Simplify.hs
apaszke: Done.

src/lib/Simplify.hs
apaszke: Done.

PR #1047, opened 2022-09-02
Eight more benchmark names

axch:
Preparation for adding those benchmarks to `main`.


performance.html
```diff
@@ -25,6 +25,14 @@ <h1 class="display-4">Dex continuous benchmarking</h1>
...
       fluidsim: 'Fluid simulation example',
       fused_sum: 'Fused summation micro-benchmark',
+      gaussian: 'Generating 1,000,000 Gaussian random variates micro-benchmark',
+      jvp_matmul 'Matrix multiply as generated by forward AD (500x500 times 500x500) micro-benchmark',
+      matmul_big: 'Matrix multiply (500x500 times 500x500) micro-benchmark',
+      matmul_small: 'Matrix multiply (1000-batch of 10x10 times 10x10) micro-benchmark',
+      matvec_big: 'Matrix-vector multiply (10000x10000 times 10000) micro-benchmark',
+      matvec_small: 'Matrix-vector multiply (10000-batch of 10x10 times 10) micro-benchmark',
+      poly: 'Pointwise operations (polynomial evaluation) micro-benchmark',
+      vjp_matmul 'Matrix multiply as generated by reverse AD (500x500 times 500x500) micro-benchmark',
```
apaszke, line 35: reverse-mode?

performance.html
```diff
@@ -25,6 +25,14 @@ <h1 class="display-4">Dex continuous benchmarking</h1>
       regression: 'Regression example',
       fluidsim: 'Fluid simulation example',
       fused_sum: 'Fused summation micro-benchmark',
+      gaussian: 'Generating 1,000,000 Gaussian random variates micro-benchmark',
+      jvp_matmul 'Matrix multiply as generated by forward AD (500x500 times 500x500) micro-benchmark',
+      matmul_big: 'Matrix multiply (500x500 times 500x500) micro-benchmark',
+      matmul_small: 'Matrix multiply (1000-batch of 10x10 times 10x10) micro-benchmark',
+      matvec_big: 'Matrix-vector multiply (10000x10000 times 10000) micro-benchmark',
+      matvec_small: 'Matrix-vector multiply (10000-batch of 10x10 times 10) micro-benchmark',
```
apaszke, line 33: I feel like we should have unbatched benchmarks too, since that's an even more common case.

apaszke (approved):

performance.html
axch: Done.

PR #1049, opened 2022-09-06
Use newtypes to represent ADTs

apaszke:
ADTs now desugar to newtypes over sums of (dependent) products. Of
course, when there's a single constructor we elide the sum. Similarly,
when a constructor has a single field, we elide the product. Finally,
constructors with no dependent fields do not use dependent products in
their representation.

This patch already seems to simplify the implementation quite a bit, but
there are still a few improvements left on the table:
- `Alt`s don't need binder nests anymore
- `DataConDef`s shouldn't need to carry the binder nest, just the
  representation type
- `SumAsProd` type is now always a sum type

And beyond the simplifications, it also generalizes our support for
dependent ADTs! We used to only allow dependent fields in
single-constructor types, but this requirement is now lifted (with the
implementation of dependent pair types doing all the heavy lifting).

### DIFF
--- a/src/lib/Builder.hs
+++ b/src/lib/Builder.hs
@@ -24,7 +24,6 @@ module Builder (
   buildLam, buildTabLam, buildLamGeneral,
   buildAbs, buildNaryAbs, buildNaryLam, buildNullaryLam, buildNaryLamExpr, asNaryLam,
   buildAlt, buildUnaryAlt, buildUnaryAtomAlt,
-  buildNewtype,
   emitDataDef, emitClassDef, emitInstanceDef, emitDataConName, emitTyConName,
   buildCase, emitMaybeCase, buildSplitCase,
   emitBlock, emitDecls, BuilderEmissions, emitAtomToName,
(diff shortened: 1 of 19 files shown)


dougalm (approved):
Looks great! Amazing to be able to delete all that redundant code.

One issue with this representation of the newtype data constructor is that since it includes the full representation of its type, the size of the atom is quadratic in the number of newtype layers. A record of records of records has to describe the leaf types three times. I'm still wondering whether the newtype data constructor could take a special ADT to describe its type instead, which could lead to a more compact representation. Something like this: 

```haskell
data Atom n =
    NewtypeDataCon (NewtypeTag n) (Atom n)
    ...

data NewtypeTag n =
  | RecordTyTag  [Label]
  | VariantTyTag [Label]
  | FinTag (Atom n)
  | NatTag
    ...
```

We might even be able to arrange it so that we can use `NewtypeTag`, along with the representation type, to represent the types too, and then we'd be able to delete the type constructors from `Atom`, like `RecordTy`, `VariantTy` and `TyCon`.

Anyway, no need to block merging this PR for that.

apaszke:
I agree and was already aware about the duplication, but I figured we can tackle that later. I also considered emitting types as let bindings (and adjusting our type querying infra to allow that), but your way would work too I think!

PR #1054, opened 2022-09-08
Rename NonNegFloat to LogSpace in stats library

darrenjw:
Rename of NonNegFloat to a generic LogSpace type, as discussed in #1039

### DIFF
--- a/lib/stats.dx
+++ b/lib/stats.dx
@@ -1,43 +1,50 @@
 '# Stats
 Probability distributions and other functions useful for statistical computing.
 
-'## Non-negative floating point numbers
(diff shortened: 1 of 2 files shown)


axch (approved):
As written, this hard-codes the representation type (and therefore precision) of both probabilities and distribution parameters.  There is precedent: the closest by coverage Haskell library I know about does the same (except that author chose double precision).

We could, if we wanted, do better, by leaving these instances parameterized by the underlying probability type.  For example,
```
data BernoulliDist p = Bernoulli p

instance {p} [VSpace p, Floating p] Dist (BernoulliDist p) Bool p
  density = \(Bernoulli prob) b.
    if b
     then Exp $ log prob
     else Exp $ log1p (-prob)
```

What do you think?  We can merge as-is, but if you want to take a stab at it, I think the parametric version would be even nicer.

By the way, if you tire of writing out repetitive class constraints, you can hack a "interface alias" of sorts by creating an interface with no content of its own and providing an instance for it:
```
interface [VSpace p, Floating p] ProbLike p
  dummy: p -> p

instance {p} [VSpace p, Floating p] ProbLike p
  dummy = \x. x
```

I'm not necessarily recommending this, just mentioning it as an option.  One down-side is that it might lead to less-precise class constraints on some methods.

darrenjw:
Thanks. I did start to try and do this, but due to the strong Float-centric nature of the prelude, it gets a bit messy. The first issues I hit were functions like `rand` and `randn` returning Floats. I could work around this, but I suspect I'll hit other related snags, and I'm a bit worried it might lead to some "ambiguous type parameter" grief for library users. So, since Float64 is quite impractical to use right now (it doesn't have a VSpace instance, none of the linear algebra operations work, etc.), I'd prefer to merge as-is right now and revisit the issue if and when Float64s become more practical. Is that OK?

axch:
SGTM.

PR #1058, opened 2022-09-09
Just enough algeff typechecking to start implementing backend

alexreinking:
Per @dougalm's suggestion, I'm opening this PR to get some eyes on the partial progress towards implementing typechecking for the serial algebraic effects implementation.

The next steps will be to implement the backend parts for effects that do not resume and then work on the typechecking rules for resume and then implementing resume, and so on.


alexreinking:
I just noticed the software gore in the merge for the GenericE stuff... will fix.

src/lib/Types/Core.hs
```diff
@@ -42,7 +42,7 @@ import Util (FileHash)
 
 import Types.Primitives
 import Types.Source
-import Types.Imp
+import Types.Imp ( IExpr(ILit), IType, ImpFunName, ImpFunction )
```
apaszke, line 45: Why the extra imports?

src/lib/Simplify.hs
```diff
@@ -490,6 +490,7 @@ simplifyAtom atom = confuseGHC >>= \_ -> case atom of
   Con con -> Con <$> (inline traversePrimCon) simplifyAtom con
   TC tc -> TC <$> (inline traversePrimTC) simplifyAtom tc
   Eff eff -> Eff <$> substM eff
+  EffOp eff idx ty -> EffOp <$> substM eff <*> pure idx <*> substM ty  -- TODO(alex): check correctness
```
apaszke, line 493: This might be a good place for an `error "unsupported"` :)

src/lib/Linearize.hs
```diff
@@ -274,6 +274,7 @@ linearizeAtom atom = case atom of
   DepPairTy _     -> emitZeroT
   TC _            -> emitZeroT
   Eff _           -> emitZeroT
+  EffOp _ _ _     -> emitZeroT  -- TODO(alex): check correctness
```
apaszke, line 277: This feels sketchy... Are you sure there's no way to have a value we're differentiating in the free vars of an `EffOp`?

apaszke: Also, I guess this should just be unhandled. AD runs on simplified code, and simplified code should not contain any effectful operations.

apaszke (commented):
This is an initial review. I haven't analyzed the inference code in depths, because I'm already too tired for today. Will take another look tomorrow.

src/lib/Types/Core.hs
alexreinking: Oops, I think these were automatically added.

alexreinking:
@apaszke -- @dougalm and I got rid of `EffOp` in the next PR by replacing it with a `Perform` PrimOp and two new Atoms: `HandlerDictCon` and `HandlerDictTy`, by analogy with the instance implementation. We chose to create new atoms because we're concerned about fusing them with the instance equivalents at this stage. The concepts are distinct. But if you feel strongly about not introducing any new atoms, we can make that change, too.

I'm still working on writing the design documents you requested offline.

src/lib/Types/Core.hs
alexreinking: Fixed in #1070

apaszke:
Thanks for removing `EffOp`! But I think it would be better to hold off on adding the handler dicts until we actually need them, i.e. when you start working on eliminating handlers in simplification. This change should really focus on type checking only, and `Perform` should be sufficient for that.

src/lib/Types/Primitives.hs
```diff
@@ -73,7 +73,9 @@ data PrimCon e =
       | ConRef (PrimCon e)
       -- Misc hacks
       | ExplicitDict e e  -- Dict type, method. Used in prelude for `run_accum`.
-      | DictHole (AlwaysEqual SrcPosCtx) e -- Only used during type inference
+      -- Only used during type inference
+      | DictHole (AlwaysEqual SrcPosCtx) e
+      | HandlerHole (AlwaysEqual SrcPosCtx) e  -- e is HandlerDictTy
```
apaszke, line 78: In DictHoles we need the SrcPosCtx in case we need to report an error message about failing to find a valid instance. But type-checking should ensure that we can always resolve the handlers, so it should be unnecessary here, right?

src/lib/Transpose.hs
```diff
@@ -286,6 +287,8 @@ transposeAtom atom ct = case atom of
   TabLam _        -> notTangent
   DictCon _       -> notTangent
   DictTy _        -> notTangent
+  HandlerDictCon _ -> notTangent  -- TODO(alex): check correctness
+  HandlerDictTy _ -> notTangent   -- TODO(alex): check correctness
```
apaszke, line 291: Yeah this seems reasonable. AD should run on simplified IR, which I imagine should not have any user-defined effects anymore

src/lib/QueryType.hs
```diff
@@ -615,6 +625,11 @@ getTypePrimOp op = case op of
   OutputStreamPtr ->
     return $ BaseTy $ hostPtrTy $ hostPtrTy $ Scalar Word8Type
     where hostPtrTy ty = PtrType (Heap CPU, ty)
+  Perform hndDict i -> do
```
apaszke, line 628: Huh, I thought that `Perform` should also take an argument and actually perform the effect. But it seems that it really acts more like a projection of the method impl from the handler, right? In that case, maybe rename it to `GetEffOp` or something else along those lines

src/lib/PPrint.hs
```diff
@@ -949,6 +958,7 @@ prettyPrecPrimCon con = case con of
   LabelCon name -> atPrec ArgPrec $ "##" <> p name
   ExplicitDict _ _ -> atPrec ArgPrec $ "ExplicitDict"
   DictHole _ e -> atPrec LowestPrec $ "synthesize" <+> pApp e
+  HandlerHole _ e -> atPrec LowestPrec $ "synthesize" <+> pApp e
```
apaszke, line 961: perhaps call it something else than `synthesize` to avoid confusion?  "resolve"?

src/lib/QueryType.hs
```diff
@@ -236,9 +240,9 @@ sourceNameType v = do
       UDataConVar v' -> lookupEnv v' >>= \case DataConBinding _ _ e -> getType e
       UClassVar   v' -> lookupEnv v' >>= \case ClassBinding  def -> return $ getClassTy def
       UMethodVar  v' -> lookupEnv v' >>= \case MethodBinding _ _ e  -> getType e
-      UEffectVar   _ -> error "not implemented: sourceNameType::UEffectVar"
-      UEffectOpVar _ -> error "not implemented: sourceNameType::UEffectOpVar"
-      UHandlerVar  _ -> error "not implemented: sourceNameType::UHandlerVar"
+      UEffectVar  v' -> getType $ userEffect v'
```
apaszke, line 243: This rule is a little surprising to me. As far as I understand the rhs of `->` always evaluated to `EffectRowKind`. But the var that names an effect is not an effect row, it's an effect. We might need to add another kind in primitives to express that

src/lib/Inference.hs
```diff
@@ -125,10 +125,105 @@ inferTopUDecl decl@(ULet _ (UPatAnn p ann) rhs) result = do
...
+  let effSourceName = uBinderSourceName effName
+  let effOpSourceNames = nestToList uBinderSourceName opNames
+  let effDef = EffectDef effSourceName (zip effOpSourceNames opTys')
+  effName' <- emitEffectDef effDef
+  opNames' <-
+    forM (enumerate effOpSourceNames) \(i, opName) -> do
+      emitEffectOpDef effName' i opName
+  let subst = effName @> effName' <.> opNames @@> opNames'
+  UDeclResultDone <$> applySubst subst result
+-- TODO(alex): handle retEff
```
apaszke, line 139: What is retEff?

src/lib/Inference.hs
```diff
@@ -1309,9 +1404,35 @@ inferUVar = \case
...
-  UHandlerVar _ -> throw NotImplementedErr "inferUVar::UHandlerVar"
+  UEffectVar _ -> do
+    throw NotImplementedErr "inferUVar::UEffectVar"  -- TODO(alex): implement
+  UEffectOpVar v -> do
+    -- the following should never fail due to operation index since
+    -- users cannot name the "return" operation
+    EffectOpBinding (EffectOpDef effName (OpIdx i)) <- lookupEnv v
+    EffectDef effSourceName _ <- lookupEffectDef effName
+    let holeTy = HandlerDictTy (HandlerDictType effSourceName effName)
+    let holeCon = Con (HandlerHole (AlwaysEqual Nothing) holeTy)
```
apaszke, line 1415: Ok, if you're not even preserving any useful source location, then we should definitely get rid of it (see my comment in Types/Core.hs)

src/lib/Inference.hs
```diff
@@ -1309,9 +1404,35 @@ inferUVar = \case
   UMethodVar v -> do
     MethodBinding _ _ f <- lookupEnv v
     return f
-  UEffectVar _ -> throw NotImplementedErr "inferUVar::UEffectVar"
-  UEffectOpVar _ -> throw NotImplementedErr "inferUVar::UEffectOpVar"
-  UHandlerVar _ -> throw NotImplementedErr "inferUVar::UHandlerVar"
+  UEffectVar _ -> do
+    throw NotImplementedErr "inferUVar::UEffectVar"  -- TODO(alex): implement
+  UEffectOpVar v -> do
```
apaszke, line 1409: This reads to me a bit as if the effectful operations were first-class functions too. Are they like that? Is this allowed?
```
f = throw
case color of
  Red -> f ()
  Green -> f ()
  Blue -> 1.0
```

I thought that even any mention of `throw` would perform the throw operation

src/lib/Inference.hs
```diff
@@ -1309,9 +1404,35 @@ inferUVar = \case
...
+    -- the following should never fail due to operation index since
+    -- users cannot name the "return" operation
+    EffectOpBinding (EffectOpDef effName (OpIdx i)) <- lookupEnv v
+    EffectDef effSourceName _ <- lookupEffectDef effName
+    let holeTy = HandlerDictTy (HandlerDictType effSourceName effName)
+    let holeCon = Con (HandlerHole (AlwaysEqual Nothing) holeTy)
+    Var <$> (emit $ Op (Perform holeCon i))
+  UHandlerVar v -> liftBuilder $ buildHandlerLam v
```
apaszke, line 1417: Instead of rebuilding the lambdas at every use site, it would be better to construct it once, when you're building the handler, and then to stash it in the `HandlerBinding`

src/lib/Inference.hs
```diff
@@ -125,10 +125,105 @@ inferTopUDecl decl@(ULet _ (UPatAnn p ann) rhs) result = do
...
+     Abs Empty (Abs bodyTyArg' (Abs tyArgs' (retTy' `PairE` retOp' `PairE` ListE ops'))) <-
+      liftInfererM $ solveLocal $ buildDeclsInf do
+        checkHandlerBodyTyArg bodyTyArg \r -> do
+          checkHandlerArgs tyArgs do
+            effName'' <- sinkM effName'
+            retTy' <- checkUType retTy
+            (retOp, ops') <- checkHandlerBody effName'' (sink r) retTy' opDefs
+            return (retTy' `PairE` retOp `PairE` ListE ops')
+     -- TODO(alex): replace Pure with actual effects
+     let handler = HandlerDef effName' bodyTyArg' tyArgs' Pure retTy' ops' retOp'
```
apaszke, line 152: What does this `Pure` mean? Is this even sound to merge without this fix?

src/lib/Inference.hs
```diff
@@ -125,10 +125,105 @@ inferTopUDecl decl@(ULet _ (UPatAnn p ann) rhs) result = do
...
+  return (retOp, ops'')
+
+checkHandlerOp :: EmitsInf o
+               => EffectName o
+               -> Type o
+               -> UEffectOpDef i
+               -> InfererM i o (Int, Block o)
+checkHandlerOp effName retTy ~(UEffectOpDef pol (InternalName _ opName) opBody) = do
+  opName' <- substM opName
+  -- TODO(alex): consider removing following redundancy
```
apaszke, line 191: elaborate on what the redundancy is?

apaszke (approved):
LGTM. I could still use some more info in the wiki page, but I think I have a reasonable understanding of what's going on now.

src/lib/Inference.hs
alexreinking: handlers can, themselves, add effects to the caller of the handler because the effect-op definitions might call effect operations. `retEff` is that effect row

src/lib/Transpose.hs
alexreinking: Indeed -- however, these are removed by #1071, which I merged into this branch

src/lib/Types/Primitives.hs
alexreinking: Removed in #1071

src/lib/PPrint.hs
alexreinking: Handler holes removed in #1071

src/lib/QueryType.hs
alexreinking: Maybe I'm confused about the context of this code? I thought these names would appear only in expressions like `MyEff` in `{MyEff|effs}`. I need a Type here but the only constructor that can take a user effect is `Eff`, which expects an effect row. So I construct just `{MyEff|}` as a singleton effect row. I didn't see anything obviously wrong with this, but maybe there is and so we _do_ need another kind?

src/lib/Inference.hs
alexreinking: Handler holes are gone. #1071

src/lib/Inference.hs
alexreinking: I'm not sure why this would be a problem? Once everything gets inlined (which we're assuming will indeed happen), we would reduce to
```
case color of
  Red -> throw ()
  Green -> throw ()
  Blule -> 1.0
```
no?

src/lib/Inference.hs
alexreinking: Yeah that would be better... we weren't sure up front how large/expensive this would be and adding data fields is... annoying.

src/lib/Inference.hs
alexreinking: This is unsound for handlers whose operations themselves call effect functions... we're just trying to push through to make `Exn` work and are ignoring this case (for now).

src/lib/Inference.hs
alexreinking: It's gone now, but the comment wasn't 🙂

src/lib/Types/Core.hs
```diff
@@ -63,6 +63,8 @@ data Atom (n::S) =
  | Con (Con n)
  | TC  (TC  n)
  | Eff (EffectRow n)
+ | EffOp (Effect n) Int (EffectOpType n)
```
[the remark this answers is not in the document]
alexreinking: EffOp removed in #1070

src/lib/PPrint.hs
```diff
@@ -250,6 +253,7 @@ instance PrettyPrec (Atom n) where
     TC  e -> prettyPrec e
     Con e -> prettyPrec e
     Eff e -> atPrec ArgPrec $ p e
+    EffOp _ _ ty -> atPrec ArgPrec $ p ty
```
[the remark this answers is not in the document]
alexreinking: EffOp removed in #1070

src/lib/Types/Core.hs
```diff
@@ -418,13 +425,103 @@ data Binding (c::C) (n::S) where
...
+instance SubstE AtomSubstVal EffectDef
+deriving instance Show (EffectDef n)
+deriving via WrapE EffectDef n instance Generic (EffectDef n)
+
+data HandlerDef (n::S) where
+  HandlerDef :: EffectName n
+             -> PiBinder n r -- body type arg
+             -> Nest PiBinder r l
+               -> EffectRow l
+               -> Type l     -- return type
```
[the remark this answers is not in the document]
alexreinking: The body type arg above is a parameter name like `r` that refers to the type of the block that is handled. The _return type_ here is like `Maybe r` (for `Exn`), which is the type that the handler ultimately returns.

src/lib/Simplify.hs
alexreinking: EffOp removed in #1070

src/lib/Linearize.hs
alexreinking: EffOp removed in #1070

PR #1061, opened 2022-09-11
Upgrade dependencies to use consistent resolver

alexreinking:
Trying to use Template Haskell with the existing dependency overrides raised some difficult-to-resolve errors. Upgrading our pinned dependencies to use a consistent resolver (lts-18.23) should make TH experiments simpler.

Basically all that needed to happen was to replace `Data.Text.Prettyprint.Doc` with `Prettyprinter` in imports.

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -15,11 +15,6 @@ extra-deps:
     subdirs:
       - llvm-hs
       - llvm-hs-pure
-  - megaparsec-8.0.0
-  - prettyprinter-1.6.2
-  - store-0.7.8@sha256:0b604101fd5053b6d7d56a4ef4c2addf97f4e08fe8cd06b87ef86f958afef3ae,8001
-  - store-core-0.4.4.4@sha256:a19098ca8419ea4f6f387790e942a7a5d0acf62fe1beff7662f098cfb611334c,1430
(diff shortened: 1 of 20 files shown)


stack.yaml
```diff
@@ -15,11 +15,6 @@ extra-deps:
     subdirs:
       - llvm-hs
       - llvm-hs-pure
-  - megaparsec-8.0.0
```
apaszke, line 18: I think those might have been pinned for performance reasons, or to work around bugs that were present in the older versions. Are they causing any issues?

apaszke (changes requested):
The pins were added for a reason, see e.g. https://github.com/google-research/dex-lang/commit/9ff97ec936b86880df25d12fdb541ec25be01414, so I'd like to make sure we're not causing any regressions here.

I'd also like to keep Dex free of TH. We've had our share of problems with it in the past (linking LLVM, increased compile times, worse caching behavior for incremental rebuilds, ...)

alexreinking:
> I'd also like to keep Dex free of TH. We've had our share of problems with it in the past (linking LLVM, increased compile times, worse caching behavior for incremental rebuilds, ...)

Ok, I had a different impression from discussions with @dougalm 

Maybe I just abandon this thread.

dougalm:
> Ok, I had a different impression from discussions with @dougalm

Yes, that's my fault, sorry. I had encouraged Alex to experiment with TH as a way to cut down on the boilerplate every time we define a new E-kinded thing. But since Adam has had bad experiences using TH, let's avoid it for now.

PR #1062, opened 2022-09-12
Add `erf` and `erfc`.  Implement `cumulative`  and `survivor`  for `NormalDist`.

emilyfertig:
The new stats library looks great!

It looks like Cuda has `erfinv` but LLVM and libm don’t, so I didn’t add the quantile function (I’m happy to, if someone has thoughts on where to get it, or a preferred approximation). 

`cumulative` and `survivor` could have better numerics. Do you think we should emulate Cephes/TFP’s ndtr? https://github.com/tensorflow/probability/blob/main/tensorflow_probability/python/internal/special_math.py If so (or if you have another preferred impl) I’ll update the PR.

I also punted on `Linearize` for now.

### DIFF
--- a/lib/prelude.dx
+++ b/lib/prelude.dx
@@ -934,6 +934,8 @@ interface Floating a
   sqrt   : a -> a
   pow    : a -> a -> a
   lgamma : a -> a
+  erf    : a -> a
+  erfc   : a -> a
 
 def lbeta {a} [Sub a, Floating a] : a -> a -> a = \x y. lgamma x + lgamma y - lgamma (x + y)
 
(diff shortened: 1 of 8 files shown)


apaszke (commented):
The implementation LGTM, but I'll leave the decisions about numerics to @axch

axch (approved):
Thanks!  I was hoping someone would pick this up :)

Re: ndtr: If you think it would be fun to port ndtr to Dex, we would be happy to have it; but this change is an improvement over the status quo, so I'm happy to merge it as-is.

I took the liberty of recording the remaining todos for good Gaussian support as Issue #1063; feel free to bite off as much or as little of that as you would like.

PR #1064, opened 2022-09-12
Write down what simplification does, and the IR that results.

axch:
This is based on a conversation with @apaszke.

### DIFF
--- a/src/lib/Simplify.hs
+++ b/src/lib/Simplify.hs
@@ -39,7 +39,61 @@ import LabeledItems
 import Types.Primitives
 import Types.Core
 
--- === simplification monad ===
+-- === Simplification ===
+
+-- The purpose of simplification is that we want first-class functions
+-- in the surface language, but we don't want to deal with them when
+-- generating code.  To that end, simplification inlines almost all
+-- functions everywhere they appear.
+
+-- In particular, simplification also carries out AD by discharging
+-- the `Linearize` and `Transpose` constructors of `PrimHof`.
+
+-- The exception is functions explicitly tagged @noinline by the
+-- programmer.  Those, however, are second-class: they are all
+-- toplevel, and get specialized until they are first order.
+
+-- Currently, simplification also discharges `CatchException` by
+-- elaborating the expression into a Maybe-style monad.  Note: the
+-- plan is for `CatchException` to become a user-defined effect, and
+-- for simplification to discharge all of them.
+
+-- Simplification also discharges bulk record and variant operations
+-- by converting them into individual projections and constructors.
+
+-- Simplification also opportunistically does peep-hole optimizations:
+-- some constant folding, case-of-known-constructor, projecting known
+-- elements from products, etc; but is not guaranteed to find all such
+-- opportunities.
+
+-- Simplification happens after dictionary synthesis (`synthTopBlock`)
+-- and before lowering.  Its input invariants are therefore
+-- - PrimCon has no DictHole constructors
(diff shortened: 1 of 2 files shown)


src/lib/Simplify.hs
```diff
@@ -39,6 +39,62 @@ import LabeledItems
...
+-- toplevel, and get specialized until they are first order.
+
+-- Currently, simplification also discharges `CatchException` by
+-- elaborating the expression into a Maybe-style monad.  Note: the
+-- plan is for `CatchException` to become a user-defined effect, and
+-- for simplification to discharge all of them.
+
+-- Simplification also discharges bulk record and variant operations
+-- by converting them into individual projections and constructors.
+-- TODO(reviewer): Can it always discharge all of them?
```
apaszke, line 63: Yes. They are forbidden from the IR after simplification

src/lib/Simplify.hs
```diff
@@ -39,6 +39,62 @@ import LabeledItems
...
+--   - Destination operations
+--   - Vector operations
+-- - PrimHof has no Dex abstract machine ops (currently Seq or RememberDest)
+
+-- The output invariants of simplification are all of the input
+-- invariants, and
+-- - Lambdas are only allowed in
+--   - the result of a top-level Block, or
+--   - immediately as arguments to `PrimHof` constructors
+-- - Non-table applications in decls can only refer to @noinlin
```
apaszke, line 84: type: noinline

src/lib/Simplify.hs
```diff
@@ -39,6 +39,62 @@ import LabeledItems
...
+--   functions
+-- - PrimHof has no CatchException, Linearize, or Transpose constructors
+-- - PrimOp has no
+--   - ThrowException constructors
+--   - RecordCons or RecordSplit constructors
+--   - VariantLift or VariantSplit constructors
+--   - TODO(reveiwer): Under what circumstances can RecordConsDynamic,
```
apaszke, line 91: Under no circumstances

src/lib/Simplify.hs
```diff
@@ -39,6 +39,62 @@ import LabeledItems
...
+-- elements from products, etc; but is not guaranteed to find all such
+-- opportunities.
+
+-- Simplification happens after dictionary synthesis (`synthTopBlock`)
+-- and before lowering.  Its input invariants are therefore
+-- - PrimCon has no DictHole constructors
+-- - PrimOp has no
+--   - Low-level memory operations
+--   - Destination operations
+--   - Vector operations
```
apaszke, line 76: I think vector operations are inessential here (simplification would handle them as written), while low-level memory and destination ops all belong to the "Dex abstract machine" umbrella IMO

src/lib/Simplify.hs
```diff
@@ -39,6 +39,62 @@ import LabeledItems
...
+--   functions
+-- - PrimHof has no CatchException, Linearize, or Transpose constructors
+-- - PrimOp has no
+--   - ThrowException constructors
+--   - RecordCons or RecordSplit constructors
+--   - VariantLift or VariantSplit constructors
+--   - TODO(reveiwer): Under what circumstances can RecordConsDynamic,
+--     RecordSplitDynamic, or VariantMake survive simplification?
+
+-- TODO(reviewer): Any other invariants?  Especially on input?
```
apaszke, line 94: I think ExplicitDict disappears after simplification too

apaszke (approved):

PR #1068, opened 2022-09-13
Docfix: Add camera.ppm download to docs.yaml

niklasschmitz:
The rendered dither.dx example at https://google-research.github.io/dex-lang/examples/dither.html currently throws a runtime error due to missing camera.ppm. 

This PR adds the download link of camera.ppm from the makefile also to the docs.yaml GH action.


apaszke:
I think that a better fix would be to adjust the makefile so that the camera download shows up as the dependency of the `docs` target (probably indirectly, through the dithering docs). I think that it might be enough to add
```
pages/examples/dither.html: dither-data
```
just as you already did for the example:
```
run-examples/dither: dither-data
```

If you apply the change, could you also remove the download from the CI script and add a similar dependency to the tutorial doc target?

niklasschmitz:
That seems like a better fix indeed, I moved the handling now into the makefile both for dither and tutorial data.

apaszke (approved):

apaszke:
This is great! Thanks a lot.

srush:
this example is so cool.

PR #1074, opened 2022-09-19
Add a pass that emits Ix dicts from types

apaszke:
Reifying Ix dicts as explicit decls makes it much more straightforward to perform simplifications on those as we apply them to the program.

### DIFF
--- a/src/lib/Lower.hs
+++ b/src/lib/Lower.hs
@@ -4,13 +4,17 @@
 -- license that can be found in the LICENSE file or at
 -- https://developers.google.com/open-source/licenses/bsd
 
-module Lower (lowerFullySequential, DestBlock, vectorizeLoops) where
+{-# LANGUAGE UndecidableInstances #-}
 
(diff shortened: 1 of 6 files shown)


src/lib/Lower.hs
```diff
@@ -29,7 +33,325 @@ import GenericTraversal
...
+      -- includes b. It's unclear to me how to get rid of those casts, since we do
+      -- want the scope after all emissions coming from b's annotations, but before
+      -- b itself. And there's no easy way to get it with the way the code is set up
+      -- right now.
+      scope <- unsafeGetScope
+      bodyEms <- emitIxScoped $ emitIxE body
+      -- We only emit decls while traversing Blocks, and the implementation for
+      -- emitIxE should handle that locally. No other place should emit!
+      refreshAbs bodyEms \(EmitIxEmissions ds REmpty) body' -> do
+        (PairB ds' b'', s, _) <- return $ generalizeIxDicts (unsafeCoerceE scope) b' ds
```
apaszke, line 133: @dougalm please let me know if you have any suggestions about this use of unsafe APIs.

PR #1081, opened 2022-09-22
Diagonal convolution benchmark

axch:


benchmarks/continuous.py
```diff
@@ -269,14 +280,23 @@ def save(commit, results: Sequence[Result], datapath, commitpath):
...
-  print('Building baseline: {BASELINE}')
-  baseline_path = build(BASELINE)
+  if len(argv) < 3:
+    raise ValueError("Expected at least three arguments!")
+  datapath, commitpath, commit = argv[:3]
+  benchmark_names = argv[3:]
+  benchmarks = []
+  for test in BENCHMARKS:
+    if not benchmark_names or test.name in benchmark_names:
+      benchmarks.append(test)
```
apaszke, line 290: This seems clearer to me?
```
benchmarks = BENCHMARKS if not benchmark_names else [b for b in BENCHMARKS if b.name in benchmark_names]
```

apaszke (approved):

benchmarks/continuous.py
axch: Done.

PR #1083, opened 2022-09-22
Print the effect row along the Pi type

apaszke:
For some reason we weren't printing the effect rows, at all! This led to a fairly long debugging session for me, when the type checker complained that `Nat -> Nat != Nat -> Nat`, because the first type was actually `Nat -> {IO} Nat`...

### DIFF
--- a/src/lib/PPrint.hs
+++ b/src/lib/PPrint.hs
@@ -336,7 +336,10 @@ instance Pretty (PiType n) where
     prettyBody = case body of
       Pi subpi -> pretty subpi
       _ -> pLowest body
-    in prettyBinder <> (group $ line <> p arr <+> prettyBody)
+    prettyEff = case eff of
+      Pure -> space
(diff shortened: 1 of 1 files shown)


src/lib/PPrint.hs
```diff
@@ -336,7 +336,10 @@ instance Pretty (PiType n) where
     prettyBody = case body of
       Pi subpi -> pretty subpi
       _ -> pLowest body
-    in prettyBinder <> (group $ line <> p arr <+> prettyBody)
+    prettyEff = case eff of
+      Pure -> space
+      _    -> pretty eff <> space
```
axch, line 341: Should this be `space <> pretty eff <> space`, or do we know that `pretty eff` always starts with enough space?

axch (approved):

src/lib/PPrint.hs
apaszke: Good point! Fixed.

PR #1084, opened 2022-09-22
Stop using Simplify in Imp

apaszke:
This change adds a pass that replaces the explicit `Ix` dict decls with newtypes that contain simplified methods in their bodies. It also switches over Imp to stop calling into simplify, since it can just look up the simplified definitions directly.

### DIFF
--- a/src/lib/Imp.hs
+++ b/src/lib/Imp.hs
@@ -24,11 +24,11 @@ import GHC.Exts (inline)
 import Err
 import MTL1
 import Name
-import Builder
+-- If you want to touch Ix methods, use the *FromSimp variants defined in this module
+import Builder hiding (unsafeFromOrdinal, ordinal, indexSetSize)
 import Syntax
 import CheckType (CheckableE (..))
 import Lower (IxDestBlock)
-import Simplify (buildBlockSimplified, dceApproxBlock, emitSimplified)
 import LabeledItems
 import QueryType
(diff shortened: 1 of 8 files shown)


src/lib/Imp.hs
```diff
@@ -1538,25 +1535,61 @@ liftBuilderImp cont = do
...
+
+ordinalPreSimp :: Emits n => IxType n -> Atom n -> BuilderM n (Atom n)
+ordinalPreSimp (IxType _ dict) i =
+  liftM unwrapNewtype $ liftEnvReaderM (projPreSimpDictThunk dict 1) >>= (`appReduce` i)
+
+unsafeFromOrdinalPreSimp :: Emits n => IxType n -> Atom n -> BuilderM n (Atom n)
+unsafeFromOrdinalPreSimp (IxType _ dict) o =
+  liftEnvReaderM (projPreSimpDictThunk dict 2) >>= (`appReduce` (Con $ Newtype NatTy o))
+
+projPreSimpDictThunk :: Atom n -> Int -> EnvReaderM n (Atom n)
```
dougalm, line 1564: I found this "preSimp" terminology confusing. I think it's meant to be read as "presimplified" or "previously simplified", but I read it as "pre simplification" as in "this works with IxTypes that haven't been simplified yet. Maybe change to "projDictThunkFromSimp" or something?

dougalm (approved):

PR #1085, opened 2022-09-22
Deduplicate Ix dicts in the emitIx pass

apaszke:
This extends the env and emissions of the emitIx pass so that we can avoid re-emitting the same dict multiple times. Most Dex programs (even the large ones) only use a handful of index types!

I initially thought that caching the emissions in the env would be sufficient for this, but it turns out it's not.
Some dict emissions might start out unique and become duplicates after generalization. As an example consider:
```
x = for i:(Fin 2) j:(..i). 1.0
for i j. x.i.j + x.i.j
```
While traversing the first `for`, we'll emit a dict `Ix (Fin 2)` and a generalized dict `(i:Fin 2) -> Ix (..i)`. We'd expect that the second `for` should not cause any extra emissions. However, once we reach `j` `Ix (..i)` is not a dict that appears in the `Env`, so it will be emitted. But, once we try to hoist that emission above the `i` binder, we will generalize it exactly to `(i:Fin 2) -> Ix (..i)`, which we already have.

With this change, the dict cache in the env acts as a preliminary filter that lets us avoid emitting unnecessary decls that would become aliases to other dicts once emission deduplication happens.

### DIFF
--- a/src/lib/Lower.hs
+++ b/src/lib/Lower.hs
@@ -13,6 +13,7 @@ import Data.Word
 import Data.Functor
 import Data.Set qualified as S
 import Data.List.NonEmpty qualified as NE
+import Data.HashMap.Strict qualified as HM
 import Data.Text.Prettyprint.Doc (pretty, viaShow, (<+>))
 import Control.Category
 import Control.Monad.Reader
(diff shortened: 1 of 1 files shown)


dougalm (approved):
lgtm!

PR #1086, opened 2022-09-23
Rewrite misc/build-web-index in Python and make it complain loudly if a document has no title

axch:
Add a test to the Makefile (and therefore to CI) checking that every decoument has a title, so that they are all included in the auto-generated InDex.

Add a title to `lib/complex.dx` so it is included in same.

### DIFF
--- a/misc/build-web-index
+++ b/misc/build-web-index
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env python3
 
 # This script constructs a Markdown file (on standard output) that constitutes
 # an index of all web-rendered pages of Dex examples, documentation, and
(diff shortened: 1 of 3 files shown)


misc/build-web-index
```diff
@@ -17,33 +17,45 @@
...
+    with open(fname, 'r') as f:
+      found_title = False
+      for line in f.readlines():
+        title = re.match(r"' *# ?(.*)", line)
+        if title:
+          print(f"- [{fname}]({link_name}.html) {title.group(1)}")
+          found_title = True
+          break
+      if not found_title:
+        raise ValueError(f"File {fname} did not have a title (top-level Markdown heading)")
```
apaszke, line 38: The `found_title` + `if` can actually be replaced by an [`else` block to the `for` loop](https://docs.python.org/3/tutorial/controlflow.html#break-and-continue-statements-and-else-clauses-on-loops).

misc/build-web-index
```diff
@@ -17,33 +17,45 @@
...
-libraries=$3
+def file_block(files):
+  for fname in files:
+    if fname.startswith("doc/"):
+      link_name = fname[len("doc/"):]
+    else:
+      link_name = fname
+    with open(fname, 'r') as f:
+      found_title = False
+      for line in f.readlines():
```
apaszke, line 31: Is there any benefit/need to search through the whole file? I would expect that the title should be the first thing that appears in it. Otherwise, we might treat a heading placed halfway as a title, but that seems wrong.

apaszke (approved):

misc/build-web-index
axch: Changed to just check the first line.

PR #1091, opened 2022-09-27
Scope discipline for link-time names

dougalm:
In the LLVM/C world, linking happens in a global name space. That clashes with
Dex's scoped naming system, which assumes that all names are abstractable and
alpha-renamable. Previously we limped along with a hack, but now that we're
about to add even more link-time names (specifically pointer constants, which
currently don't work with standalone functions) it's time to fix it.

With this change, each top-level standalone function is compiled as a speparate
LLVM module, and the functions it calls are given arbitrary local names that are
only used within that particular module. We maintain a mapping between these
names and our usual well-scoped "E-kinded" names. At link time, we provide the
linker with an explicit mapping from the local names to the actual function
pointers in the current process (thanks to [Adam's recent patch to llvm-hs](https://github.com/llvm-hs/llvm-hs/pull/404)).
Linking has to be redone when we start a new process, but everything else
can be stored on disk.

This way, object files contain no names outside of our own naming system. We're
free to cache them, merge separately compiled functions, use different versions
of a single library, and so on.


apaszke (commented):
We should think about Export too!

src/lib/Types/Core.hs
```diff
@@ -392,26 +400,32 @@ data Cache (n::S) = Cache
...
 type ObjectFileName = Name ObjectFileNameC
-
+type ObjectFileContents = BS.ByteString
 data ObjectFile n = ObjectFile
-  { objectFileContents :: BS.ByteString
-  , objectFileFunsDefined :: [CFunName]
-  -- XXX: direct dependencies only
-  , objectFileDeps :: [ObjectFileName n] }
+  { objectFileContents :: ObjectFileContents
+  , objectFileNameMap  :: CNameMap ObjectFileName n }
```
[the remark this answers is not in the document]
dougalm: Good point. I'll rename `ObjectFile` to `FunObjCode` which happily has the same number of characters :)

src/Dex/Foreign/JIT.hs
```diff
@@ -36,7 +36,6 @@ import Name
 import Logging
```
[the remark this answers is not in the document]
dougalm: As discussed, I'll use the single LLVM JIT we store in the DexJIT instead of create a fresh one each time.

src/lib/LLVMExec.hs
```diff
@@ -81,27 +88,47 @@ type DexExitCode = Int
...
+  tm <- LLVM.Shims.newHostTargetMachine R.PIC CM.Large CGO.Aggressive
+  jit <- LLVM.JIT.createJIT tm
+  return $ DexJIT jit tm opt
+
+compileLLVMModule :: DexJIT -> L.Module -> LocalCName -> IO FunObjCodeContents
+compileLLVMModule jit ast exportName = do
+  let tm = jitTargetMachine jit
+  withContext \c -> do
+    Mod.withModuleFromAST c ast \m -> do
+      internalize [exportName] m  -- TODO(dougalm): I don't know what this does. Is it needed?
```
apaszke, line 102: I think we talked about this. Perhaps remove this comment?

src/lib/JIT.hs
```diff
@@ -74,7 +81,7 @@ data CompileState = CompileState
   , globalDefs  :: [L.Definition]
   , curDevice   :: Device
   -- TODO: use safe names for object files
```
apaszke, line 83: I think this is resolved?

src/lib/JIT.hs
```diff
@@ -405,52 +441,52 @@ compileInstr instr = case instr of
...
+      FunctionSubstVal f' lTy (IFunType cc _ _) -> do
+        case cc of
+          CEntryFun -> do
+            exitCode <- emitCallInstr lTy f' args' >>= (`asIntWidth` i1)
+            compileIf exitCode throwRuntimeError (return ())
+            return []
+          CInternalFun -> do
+            exitCode <- emitCallInstr lTy f' args' >>= (`asIntWidth` i1)
+            compileIf exitCode throwRuntimeError (return ())
+            return []
```
apaszke, line 455: Since those two are identical, perhaps you could do this:
```
case cc of 
  CEntryFun -> return ()
  CInternalFun -> return ()
  _ -> error "..."
exitCode <- emit ...
...
```

src/lib/JIT.hs
```diff
@@ -47,15 +50,19 @@ import qualified RawName as R
...
-import Util (IsBool (..), bindM2)
+import Util (IsBool (..), bindM2, enumerate)
 
 -- === Compile monad ===
 
-type OperandSubstVal = SubstVal AtomNameC (LiftE Operand)
+data OperandSubstVal (c::C) (n::S) where
+  OperandSubstVal  :: L.Operand -> OperandSubstVal AtomNameC n
+  FunctionSubstVal :: L.Operand -> LLVMFunType -> IFunType -> OperandSubstVal ImpFunNameC   n
+  RenameOperandSubstVal :: Name c n -> OperandSubstVal c n
```
apaszke, line 63: We never emit those during the traversal, so perhaps we could call them top-level subst vals? or ffi subst vals?

apaszke (commented):
This generally LGTM, but I would really like us to make sure we're not using the JIT from multiple threads in the evaluator. We should figure out some sort of a locking strategy, until we carefully work out that it's actually ok to be used this way.

It would be also nice to get rid of the JITs on the Python side, but I guess that can be done in another PR.

src/lib/JIT.hs
dougalm: Quite right. I didn't notice!

src/lib/JIT.hs
dougalm: Added a comment explaining that that's the only way in which we currently use it

src/lib/LLVM/HEAD/JIT.hs
```diff
@@ -101,15 +105,39 @@ compileModule moduleJIT@JIT{..} objFiles ast compilationPipeline = do
 foreign import ccall "dynamic"
   callDtor :: FunPtr (IO ()) -> IO ()
 
+loadNativeModule :: JIT -> ExplicitLinkMap -> ObjectFileContents -> IO NativeModule
+loadNativeModule moduleJIT linkingMap objFileContents = do
+  moduleDylib <- newDylib moduleJIT
+  let cl = compileLayer moduleJIT
+  addExplicitLinkMap cl moduleDylib linkingMap
+  OrcJIT.addDynamicLibrarySearchGeneratorForCurrentProcess cl moduleDylib
+  let moduleDtors = [] -- TODO(dougalm): handle destructors (with help from Adam)
```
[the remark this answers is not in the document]
dougalm: done

apaszke (commented):
We should still fix the multithreading issue before we merge this!

src/lib/Types/Core.hs
```diff
@@ -392,25 +401,32 @@ data Cache (n::S) = Cache
...
+-- `dexCalledFunction_0_bar`, `dexCalledFunction_1_baz` (The `_foo` suffixes are
+-- just for readability and don't need to be unique.) The only way we use these
+-- names is to link the object code against known pointers to the functions it
+-- calls, and then to obtain its own pointer.
+
+type LocalCName = String -- has no meaning beyond a single LLVM module or object file
+data CNameMap (e::E) (n::S) = CNameMap
+  { mainFunName    :: LocalCName
+  , calledFunNames :: M.Map LocalCName (e n)
+  , dtorList       :: [LocalCName] }
```
apaszke, line 421: It feels weird to have the main func name and destructors be part of a "name map", since they don't map to anything. Perhaps we should move them to some extra fields of a function object code and keep this a map for just its dependencies

src/lib/LLVM/HEAD/JIT.hs
```diff
@@ -61,55 +62,68 @@ type CompilationPipeline = LLVM.Module -> IO ()
...
   -- TODO: Clear the dylib
   forM_ moduleDtors callDtor
 {-# SCC unloadNativeModule #-}
 
-withNativeModule :: JIT -> [ObjectFileContents] -> LLVM.AST.Module -> CompilationPipeline -> (NativeModule -> IO a) -> IO a
-withNativeModule jit objs m p = bracket (compileModule jit objs m p) unloadNativeModule
+withNativeModule
+  :: JIT -> ExplicitLinkMap -> [String]
+  -> LLVM.AST.Module -> CompilationPipeline -> (NativeModule -> IO a) -> IO a
+withNativeModule jit linkMap dtors m p =
```
apaszke, line 125: This shouldn't be taking the extra destructors argument since we can recover those form the LLVM.AST.Module. This is duplicating the data and might end up being error prone in case they disagree

PR #1092, opened 2022-09-27
Perform loop invariant code motion (and DCE) after lowering

apaszke:
Most importantly, to lift allocs of arrays that have stable (but not necessarily statically known) shapes out of the hot inner loops. This dramatically improves the performance (40x !) of the `conv` benchmark when it is not specialized on the `size` parameter (you can replace the `3` by `sum [3]` and see for yourself).

### DIFF
--- a/src/lib/Optimize.hs
+++ b/src/lib/Optimize.hs
@@ -6,7 +6,7 @@
 
 {-# LANGUAGE UndecidableInstances #-}
 
-module Optimize (earlyOptimize, optimize, peepholeOp) where
+module Optimize (earlyOptimize, optimize, peepholeOp, hoistLoopInvariant, dceIxDestBlock) where
 
 import Data.Functor
 import Data.Word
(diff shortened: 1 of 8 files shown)


axch (approved):

src/lib/Optimize.hs
```diff
@@ -211,6 +211,92 @@ instance GenericTraverser ULS where
...
+instance GenericTraverser LICMS where
+  traverseExpr = \case
+    Hof (Seq dir ix (ProdVal dests) (Lam (LamExpr b body@(Block ann _ _)))) -> do
+      Distinct <- getDistinct
+      ix' <- traverseAtom ix
+      dests' <- traverse traverseAtom dests
+      let numCarries = length dests
+      Abs hdecls destsAndBody <- traverseLamBinder b \b' -> do
+        -- First, traverse the block, to allow any Hofs inside it to hoist their own decls.
+        Block _ decls ans <- traverseGenericE body
```
dougalm, line 241: Is this going to be quadratic in the loop nesting depth? Not a blocking problem if so. Just curious.

dougalm (approved):
Looks great! I like that the rule is as simple as "hoist everything you can, so long as it's pure", which makes it easy to reason about at the user level. It's not perfectly work-preserving because loop bodies might be executed zero times dynamically. But I think it's still the right thing to do, rather than some complicated alternative that checks for the zero case. In my mental cost model, `n >> 1`, even though `n=0`, `n=1` and `n=2` are all technically possible.

src/lib/Optimize.hs
apaszke: I think it's still linear. This `traverseGenericE` is the first traversal of the body, while the cost of the `seqLICM` call below is equal to the number of decls _in the returned block_. Any decls nested in blocks inside are not re-traversed, since we never open up the expressions.

Well, I guess for as long as we don't hoist the allocs and the substitution is the identity substitution... Otherwise I think we do need to re-traverse to substitute and that is potentially exponential in the nesting depth...

apaszke:
Yes, I considered the issues with speculating the pure expressions a bit too much, but in the end decided that it might be a good enough first try. I'm less worries about work preservation and more about hoisting expressions that contain `error` or `unsafe_io`. But I guess in the end that's what's unsafe in `unsafe_io`!

apaszke:
At some point we might also want to make this thing less aggressive, to avoid blowing up memory usage. Code like this one:
```
for _:(Fin m).
  x1 = for i:(Fin n). ...
  x2 = for i:(Fin n). ...
  ...
  xk = for i:(Fin n). ...
  ()  -- maybe this loop is executed for effects
```
has a memory footprint O(n), but after this LICM it will rise to O(nk) since we don't try to deduplicate the allocs

PR #1093, opened 2022-09-27
Educational: built-in `map` primitive

normanrink:
This is educational work.
DO NOT MERGE.

Implement a built-in `map` primitive for mapping a function over an array.

### DIFF
--- a/src/lib/Simplify.hs
+++ b/src/lib/Simplify.hs
@@ -839,6 +839,9 @@ projectDictMethod d i = do
 
 simplifyHof :: Emits o => Hof i -> SimplifyM i o (Atom o)
 simplifyHof hof = case hof of
+  Map f -> do
+    f' <- simplifyAtom f
+    liftM Var $ emit $ Hof $ Map f'
   For d ixDict lam -> do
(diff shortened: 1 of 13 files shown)


src/lib/Simplify.hs
```diff
@@ -835,9 +835,19 @@ projectDictMethod d i = do
...
+    (fun', Abs b recon) <- simplifyLam fun
+    array' <- simplifyAtom array
+    ans <- liftM Var $ emit $ Hof $ Map fun' array'
+    case recon of
+      IdentityRecon -> return ans
+      LamRecon reconAbs -> do
+        TabPi (TabPiType (_:>ixTy) _) <- getType array'
+        buildTabLam noHint ixTy \i -> do
+          locals <- tabApp (sink ans) $ Var i
+          ithArg <- emitAtomToName =<< (tabApp (sink array') $ Var i)
```
apaszke, line 848: Instead of emitting to name, you could `b @> SubstVal ithArg` below

apaszke (commented):
Looks good!

src/lib/Simplify.hs
normanrink: Thanks.

PR #1101, opened 2022-09-28
Occurrence analysis data structures and unit tests

axch:
This change adds a series of internal representations for binding occurrence information and transformations between them.  The proximate goal is to support inlining array-valued bindings into loop bodies, where safe; but in the future some evolution of these could perhaps be used for assessing e.g., when a reference is written to exactly once in each cell so that it can be replaced with a producer loop.

The occurrence analysis structures here are coded to be able to see through algebraic index types, to detect that e.g.
```
  for i. xs.(Left i)
  for j. xs.(Right j)
```
in fact reads `xs` only once per element.

This change also adds a `make watch-tests` target that asks `stack` to rerun the Haskell unit tests in response to source file edits.

A couple open issues with this change I want to record for posterity:

- These structures only keep track of work (e.g., FLOP) preservation, not locality.  Inlining a locality-sensitive producer into transposition could be problematic; if we care about preventing that, these structures would need to be updated to keep some notion of locality in mind.

- There is a known precision problem: the pipeline is conservative about the number of dimensions of the binding that need to be apparent for inlining to be safe.  For instance, if the use is for j. xs.j.j then it's not work-preserving to inline `xs` if its definition cannot be immediately cancelled against one level of indexing. However, the analysis will conservatively assume two levels of indexing are required.

### DIFF
--- a/makefile
+++ b/makefile
@@ -271,6 +271,9 @@ tests: opt-tests unit-tests lower-tests quine-tests repl-test module-tests doc-f
 unit-tests:
 	$(STACK) test --work-dir .stack-work-test $(STACK_FLAGS)
 
+watch-unit-tests:
+	$(STACK) test --work-dir .stack-work-test $(STACK_FLAGS) --file-watch
+
 opt-tests: just-build
 	misc/file-check tests/opt-tests.dx $(dex) -O script
 
--- a/src/lib/Occurrence.hs
+++ b/src/lib/Occurrence.hs
@@ -5,6 +5,7 @@
 -- https://developers.google.com/open-source/licenses/bsd
 
 {-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE StrictData #-}
 
 module Occurrence where
 
@@ -13,6 +14,7 @@ import Prelude qualified
 
 import Control.Monad.State.Strict
 import Data.IntMap.Strict qualified as M
+import Data.List (foldl')
 
 import Name
 
@@ -360,8 +362,8 @@ interp :: Access iter -> Use iter
 interp = \case
   Location [] ct -> Const ct
   Location (expr:exprs) ct -> At expr $ interp $ Location exprs ct
-  Any accesses -> foldl  max zero $ map interp accesses
-  All accesses -> foldl plus zero $ map interp accesses
+  Any accesses -> foldl'  max zero $ map interp accesses
+  All accesses -> foldl' plus zero $ map interp accesses
   ForEach b access -> iterate False b $ interp access
 
 -- The interesting function is resolution of iteration-space loops.
@@ -390,7 +392,7 @@ iterate injective ib (At expr use) = At expr' use' where
   use' = iterate injective' ib use
 iterate injective ib (Max uses) =
   -- TODO This can be an underapproximation if the upstream structure
-  -- is expressive enough, but I think it's OK for now.  To wit,
+  -- becomes expressive enough, but it's OK for now.  For example,
   -- consider the access pattern
   --   for i.
   --     if ...
@@ -412,12 +414,16 @@ iterate injective ib (Max uses) =
   -- However, the current Access structure cannot actually _prove_
   -- that inlining `xs` into the latter expression is safe, because
   -- the addition will register as an unknown function of `i`.  The
-  -- distribution bug is therefore masked.  I further conjecture that
-  -- there are as yet no Access-representable expressions that
-  -- actually trigger a problem here.
-  foldl max zero $ map (iterate injective ib) uses
+  -- distribution bug is therefore masked.
+  --
+  -- In fact, that currently always happens.  All distinct functions
+  -- that produce indices of type n that Access can currently
+  -- represent are either `Unknown` or have disjoint ranges, so this
+  -- kind of underapproximation cannot yet lead to a mistaken
+  -- assertion that some binding is safe to inline.
+  foldl' max zero $ map (iterate injective ib) uses
 iterate injective ib (Sum uses) =
-  foldl plus zero $ map (iterate injective ib) uses
+  foldl' plus zero $ map (iterate injective ib) uses
 
 iterateIxExpr :: NameBinder AtomNameC iter iter'
               -> IxExpr iter' -> State Bool (IxExpr iter)
@@ -463,6 +469,12 @@ iterateIxExpr _ IxAll = return IxAll
 -- latter case, the object is not accessed outside that subset.  The
 -- Subset structure maintains a map of further subsets, and how later
 -- dimensions are accessed in each of them.
+
+-- Note that we care not just about which elements are used, but also
+-- the indexing depth of the object.  That is `CConst One` and
+-- `CSubset (SAll (CConst One))` both mean "used everywhere exactly
+-- once", but the latter also means "do not inline unless you have an
+-- explicit `for` to cancel against indexing".
 data CollapsedUse =
     CConst Count
   | CSubset (Subset CollapsedUse)
@@ -494,8 +506,8 @@ collapse (At expr use) =
   let use' = collapse use in
   if use' == zero then zero
   else CSubset $ exprToSubset expr use'
-collapse (Sum uses) = foldl plus zero $ map collapse uses
-collapse (Max uses) = foldl max zero $ map collapse uses
+collapse (Sum uses) = foldl' plus zero $ map collapse uses
+collapse (Max uses) = foldl' max zero $ map collapse uses
 
 exprToSubset :: IxExpr n -> a -> Subset a
 -- An iteration-space variable approximates to "all", because at this
@@ -631,7 +643,7 @@ instance ApproxConst a => ApproxConst (Subset a) where
   approxConst (Prod _ use) = approxConst use
 
 instance ApproxConst v => ApproxConst (M.IntMap v) where
-  approxConst uses = foldl max zero $ map approxConst $ M.elems uses
+  approxConst uses = foldl' max zero $ map approxConst $ M.elems uses
 
 -- === Notes ===
 
--- a/src/lib/Occurrence.hs
+++ b/src/lib/Occurrence.hs
@@ -1,0 +1,671 @@
+-- Copyright 2022 Google LLC
+--
+-- Use of this source code is governed by a BSD-style
+-- license that can be found in the LICENSE file or at
+-- https://developers.google.com/open-source/licenses/bsd
+
+{-# LANGUAGE DeriveAnyClass #-}
+
+module Occurrence where
+
+import Prelude hiding (abs, iterate, max, sum)
+import Prelude qualified
+
+import Control.Monad.State.Strict
+import Data.IntMap.Strict qualified as M
+
+import Name
+
+-- === Use Analysis ===
+
+-- We want to know how a binding is used in a Dex program, so as to
+-- know whether inlining it is guaranteed to conserve work and code
+-- size.
+
+-- Conservation of code size is easy.  Coservation of work, however,
+-- is more subtle.  Inlining the computation of x into
+--   for i. x
+-- does not conserve work (unless the index set has size 0 or 1), but
+-- inlining x into
+--   for i. x.i
+-- does conserve work -- provided we have access to the loop body
+-- that constructs it so we can beta-reduce against the indexing
+-- operation.
+
+-- The main difficulty is that Dex allows indexing by expressions
+-- other than referencing a loop binder, so to solve subtle cases we
+-- need to translate iteration-space expressions into data-space.
+-- For example, inlining x into
+--   for i. x.(Left i)
+--   for j. x.(Right j)
+-- conserves work, because the two injections have disjoint ranges.
+
+-- We proceed with a four-phase translation.
+--
+-- - Phase 1 translates Dex source into the Access structure.  An
+--   Access mirrors the control pattern of the Dex source, abstracting
+--   away most Dex syntax but still referring to iteration space.
+--   Access retains nearly complete usage information.
+--
+-- - Phase 2 transposes Access into Use by turning iteration space
+--   into data space.  (Use is like Access, except organized in data
+--   space.)  Use retains about the same amount of usage information
+--   as Access.
+--
+-- - Phase 3 collapses Use into CollapsedUse.  This process normalizes
+--   the subsets of the data space that are actually accessed, such as
+--   realizing that the subsets of all `Left` values and all `Right`
+--   values cannot intersect.  This is where most of the
+--   overapproximations are; CollapsedUse forgets much of the
+--   detailed usage information Use retained.
+--
+-- - Phase 4 actually extracts the inlining-relevant information from
+--   CollapsedUse.
+
+-- === Algebra ===
+
+-- The algebraic structure we want in order to represent uses is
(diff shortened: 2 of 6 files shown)


makefile
```diff
@@ -271,6 +271,9 @@ tests: opt-tests unit-tests lower-tests quine-tests repl-test module-tests doc-f
 unit-tests:
 	$(STACK) test --work-dir .stack-work-test $(STACK_FLAGS)
 
+watch-tests:
```
apaszke, line 274: `watch-unit-tests`? Since we use `tests` to mean all tests, not just those.

src/lib/Occurrence.hs
```diff
@@ -0,0 +1,623 @@
...
+-- that constructs it so we can beta-reduce against the indexing
+-- operation.
+
+-- The main difficulty is that Dex allows indexing by expressions
+-- other than referencing a loop binder, so to solve subtle cases we
+-- need to translate iteration-space expressions into data-space.
+-- For example, inlining x into
+--   for i. x.(Left i)
+--   for j. x.(Right j)
+-- conserves work, because the two injections have disjoint ranges.
```
apaszke, line 41: Note that this depends on the Ix instance. I think that an `Ix ((Fin n)|(Fin n))` that collapses both cases onto a single `Fin n` is lawful (`size = n` and let's say that `unsafe_from_ordinal` always picks the left case), but would not be work-preserving in this case!

src/lib/Occurrence.hs
```diff
@@ -0,0 +1,623 @@
...
+  plus :: a -> a -> a
+
+-- === Access ===
+
+-- We represent an upper bound on the pattern by which (the object
+-- bound to) a name is accessed.  The goal is to determine whether
+-- inlining the name could duplicate work, so this is an upper bound
+-- (as opposed to an estimate or a lower bound).
+
+-- Each leaf or subtree of the object may be accessed a known or
```
apaszke, line 92: What is a "leaf or subtree of the object"?

src/lib/Occurrence.hs
```diff
@@ -0,0 +1,623 @@
...
+--         ...
+--
+-- - `All parts` turns into sequentialization
+--     \ref.
+--       compile part1 ref
+--       compile part2 ref
+--       ...
+--
+-- - `ForEach b sub` is iteration
+--     \ref. for b in (type of b).
```
apaszke, line 199: The `b` in `ForEach` is not annotated by its type, making this correspondence a little difficult to understand. Is it like that, because all `b`s are really instances of some index sets, but for the purpose of this analysis it doesn't matter which ones? If so, it might be worth adding that to the comment

src/lib/Occurrence.hs
```diff
@@ -0,0 +1,623 @@
...
+--
+-- Note that a Use can still depend on iteration-space indices (from
+-- the `iter` scope)---it just doesn't have internal binders that are
+-- meant to be interpreted in iteration space.
+--
+-- A Use implicitly iterates over as many dimensions of the accessed
+-- object as there are nesting levels of At constructors in the Use.
+-- We do not use explicit binders for this because they always appear
+-- in the same nesting order and can always be pulled to the outermost
+-- layer, so it's simplest to leave them implicit.
```
apaszke, line 215: I don't understand this comment and the previous one

src/lib/Occurrence.hs
```diff
@@ -0,0 +1,623 @@
...
+
+data Use (iter::S) =
+  -- The object is accessed in its entirety
+    Const Count
+  -- The object is accessed according to the third argument, but only
+  -- where the first argument equals the second.
```
apaszke, line 221: There is no third argument here

src/lib/Occurrence.hs
```diff
@@ -0,0 +1,623 @@
...
+-- how many times each index is used.
+--
+-- Note that now IxExpr permits the `IxAll` constructor, so an
+-- `IxExpr` represents a subset of indices rather than a single index.
+--
+-- - `Const ct` becomes accumulation
+--     \ref. ref += ct
+--     Also polymorphic in the ref type, applying elementwise
+--
+-- - `Max alts` becomes case
```
apaszke, line 238: It is a little weird that `Max` corresponds to non-deterministic choice and not to a max

src/lib/Occurrence.hs
```diff
@@ -0,0 +1,623 @@
...
+--
+-- - `Sum parts` becomes sequencing
+--     \ref.
+--       compile alt1 ref
+--       compile alt2 ref
+--       ...
+--
+-- - `At expr use` becomes testing and slicing
+--     \ref. for b in <next object dim>.
+--       if b in expr then (compile use ref!b)
```
apaszke, line 252: what does `b in expr` mean?

src/lib/Occurrence.hs
```diff
@@ -0,0 +1,623 @@
...
+--
+-- Another way to see this is that the compilation of At is the only
+-- thing that cares about and alters the type of the reference being
+-- passed around.  Ergo, all the At constructors must occur in the
+-- same order.
+
+-- === Use smart constructors ===
+
+-- We maintain a normal form for Use data with these smart
+-- constructors.  The main invariant is
```
apaszke, line 273: There are no smart constructors here, only the `MaxPlus` instance. But its methods don't seem sufficient to me? I think `Const One` is also a good normal form for a `Use`

src/lib/Occurrence.hs
apaszke: I think the interpreter below uses the max, as I would expect

src/lib/Occurrence.hs
```diff
@@ -0,0 +1,623 @@
...
+-- which represents a data-space loop that checks that the data-space
+-- variable equals the indexing expression.
+interp :: Access iter -> Use iter
+interp (Location [] ct) = Const ct
+interp (Location (expr:exprs) ct) = At expr rest where
+  rest = interp $ Location exprs ct
+interp (Any accesses) = foldl max zero uses where
+  uses = map interp accesses
+interp (All accesses) = foldl plus zero uses where
+  uses = map interp accesses
```
apaszke, line 342: Perhaps just `$ map interp accesses` and similarly in all cases above, to avoid the extra hang? It might also be more readable if you use lambda case (since you won't need to mention `interp` so many times):
```
interp = \case
  Location [] ct -> Const ct
  Location (expr: exprs) ct -> ...
  ...
```

apaszke (commented):
I would love to understand this better, but I was still unable to reconstruct what this code does even when carefully following the great comments. I got lost around `iterate`, mainly because I don't fully understand `ForEach` and `At`.

src/lib/Occurrence.hs
```diff
@@ -0,0 +1,623 @@
...
+  -- on a more fine-grained basis than "was this used more than once?"
+  if ct == zero then zero else Const Unbounded
+iterate injective ib (At expr use) = At expr' use' where
+  (expr', injective') =
+    flip runState injective $ iterateIxExpr ib expr
+  use' = iterate injective' ib use
+iterate injective ib (Max uses) =
+  foldl max zero $ map (iterate injective ib) uses
+iterate injective ib (Sum uses) =
+  foldl plus zero $ map (iterate injective ib) uses
```
apaszke, line 372: This rule essentially reassociates the sum coming from the iteration over the sum in the use, right?

The previous one seems like it might be an underapproximation? `sum for i. (max xs.i) > max (sum for i. xs.i)` and we go from the first one to the second one

src/lib/Occurrence.hs
```diff
@@ -0,0 +1,623 @@
...
+-- binders, potentially severely.
+
+-- A CollapsedUse is either a count of the accesses to the whole
+-- object, or a subset of the leading dimension of the object.  In the
+-- latter case, the object is not accessed outside that subset.  The
+-- Subset structure maintains a map of further subsets, and how later
+-- dimensions are accessed in each of them.
+data CollapsedUse =
+    CConst Count
+  | CSubset (Subset CollapsedUse)
```
apaszke, line 420: Is `CSubset (SAll (CConst c))` equivalent to `CConst c`?

src/lib/Occurrence.hs
```diff
@@ -0,0 +1,623 @@
...
+-- intersection as efficiently as is practicable.
+-- We include an optimization: The Inj constructor carries an optional
+-- payload associated with the whole type, instead of broadcasting it
+-- over all the cases.
+data Subset a =
+  -- The index could be any type, and all of it is accessed.
+    SAll a
+  -- The index is a sum.  We maintain a map from constructors to
+  -- subsets of their respective types.
+  | Inj  (Maybe a) (M.IntMap (Subset a))
```
apaszke, line 436: What is the `Maybe`?

makefile
axch: Done.

src/lib/Occurrence.hs
axch: I was under the impression that we expect every `Ix n` instance to implement a bijection between `n` and a prefix of the non-negative integers.  Is that not so?

src/lib/Occurrence.hs
axch: Until we generalize to tuple-typed accessees, this just means element or subarray.  Comment adjusted.

src/lib/Occurrence.hs
axch: In the explanation I'm assuming that if necessary, I have the Dex environment lying around, where the type that `b` indexes can be looked up.  But the code currently doesn't use that type (only the semantics do), so there is no need to explicitly store it.  Updated the comments.

src/lib/Occurrence.hs
axch: Tried to clarify by reordering text and giving examples.

src/lib/Occurrence.hs
axch: Stale documentation refreshed.

src/lib/Occurrence.hs
axch: Well, if I am trying to get the smallest value I can, offering my adversary a free choice computes a max between those options.  But happy to rewrite if there's a clearer code expression?

src/lib/Occurrence.hs
axch: The expression in general represents a set of positions.  Comment added to the `At` constructor of `Use`.

src/lib/Occurrence.hs
axch: Outdated wording fixed.

src/lib/Occurrence.hs
axch: Thanks!

src/lib/Occurrence.hs
axch: Good one!  I'm not sure how to deal with this, because we really do need to push `iterate` as far down as we can, to get rid of the iteration-space binders.

For now I think it's actually fine, due to an accident of the upstream Access structure; left a detailed comment in the code.

src/lib/Occurrence.hs
axch: Almost.  They both represent the same amount of access, but `CConst One` corresponds to `xs`, which is work-preserving to inline even if the creation expression is completely opaque; whereas `CSubset (SAll (CConst c))` corresponds to `for i. xs.i`, which is only work-preserving to inline if the outermost dimension is immediately accessible to cancel against the indexing expression.

src/lib/Occurrence.hs
axch: It's an optional value that's associated with all the cases, as an optimization over explicitly broadcasting it.  Tried to clarify the comment.

axch:
Thanks for the engaged review!  I hope the updated comments are a bit clearer; do you want to get together and talk about what's happening and how to best explain it?

src/lib/Occurrence.hs
apaszke: Good question. Technically, it would still work even if it was a surjection onto the prefix given by the size. So it's really a question of the contract we pick for `Ix`.

src/lib/Occurrence.hs
apaszke: Yeah I think I agree. Consider `sum for i. (max x y)` (`x` and `y` are uses that might contain `i`). My take on this would be that for this to be an underapproximation, we'd need `x` to be sometimes larger than `y` and sometimes smaller than `y`. But I think that `Use` expressions cannot be monotonic in this way, at least not given the current set of constructors? You understand this better, so could you perhaps finish this argument? I'm pretty sure it will be much simpler than whatever is in the code right now.

src/lib/Occurrence.hs
apaszke: Ah, it might be worth adding this as a comment then. This data structure is not just about the use, but also about the number of indices applied to a value.

src/lib/Occurrence.hs
axch: Yes.  But in a case like `expr == Inject 0 IxAll`, this if is really testing "is `b` a `Left`?"

src/lib/Occurrence.hs
axch: Comment added.

src/lib/Occurrence.hs
```diff
@@ -0,0 +1,671 @@
...
+--       compile alt2 ref
+--       ...
+--
+-- - `At expr use` becomes testing and slicing
+--     \ref. for b in <next object dim>.
+--       if b in expr then (compile use ref!b)
+
+-- This semantics implies a transformation that we perform eagerly:
+-- - It's safe to distribute Sum into Max, so that Max ends up on the
+--   outside.
```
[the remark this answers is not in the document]
axch: With `Sum` it is safe, because we take the combinatorial hit to actually distribute it.  That is,
`max(a, b) + max(c, d) <= max(a + c, a + d, b + c, b + d)`.  In the `for` case we were doing the equivalent of turning `max(a, b) + max(c, d)` into `max(a + c, b + d)`, which indeed is not safe.

src/lib/Occurrence.hs
```diff
@@ -0,0 +1,671 @@
...
+-- The main interpreter is pretty straightforward.  The main trick is
+-- just semantics: an indexing expression becomes a (nested) `At`,
+-- which represents a data-space loop that checks that the data-space
+-- variable equals the indexing expression.
+interp :: Access iter -> Use iter
+interp = \case
+  Location [] ct -> Const ct
+  Location (expr:exprs) ct -> At expr $ interp $ Location exprs ct
+  Any accesses -> foldl  max zero $ map interp accesses
+  All accesses -> foldl plus zero $ map interp accesses
```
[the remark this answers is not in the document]
axch: Done.

src/lib/Occurrence.hs
apaszke: I don't follow your counterargument. Think of `x` and `y` as a count of uses in the two case branches. If one branch is consistently larger than the other then commuting a max with a sum is actually sound and will return in the exact same value being returned.

src/lib/Occurrence.hs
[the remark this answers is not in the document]
apaszke: Ok, I agree that if you take a combinatorial hit then it is indeed safe, but that's not what I would understand from this comment.

apaszke (approved):
LGTM, although I would like to clarify the comment around sum-of-max to use the simpler example I proposed

PR #1106, opened 2022-10-05
Split up compilation, linking and execution into separate steps.

dougalm:
Previously, we rolled these all together in `compileAndEval`, while in other
places we separated them, like in compiling top-level functions. This had a few
problems:
  1. We ended up with different compilation paths for top-level functions,
     evaluated expressions, and export/foreign functions.
  2. Long-running linking meant a long-running compilation jit, which led to
     concerns about thread safety of LLVM APIs (see Adam's comments on PR
     #1091). Long-running state also makes caching tricky and casts shadows in
     lots of places.
  3. We exposed a lot of LLVM surface area because of all the different ways in
     which we interacted with it. We were worried that this would make it hard
     to e.g. switch to MLIR as an alternative backend.

Main changes
  * All compilation now goes through `LLVM.Compile`, which produces object code
  * All linking now goes through `LLVM.Link`, which yields a pointer to
    executable code and a callback for cleanup. We don't even need to use LLVM
    for this, and we could swap it out for `ld` or something.
  * All execution goes through `Runtime`, which doesn't call into LLVM at all!
  * There is no persistent LLVM state at all. We create a temporary jit every
    time we need to compile a function and we destroy it once we have the object
    code. We create separate temporary jits each time we need to link/load some object
    code. We keep these jits alive until we destroy them by calling the teardown
    callback for the linked function. (We might want to be more thoughtful about
    how we manage resources for these.)
  * `Dex.Foreign` stores the top env inside an `MVar` rather than passing it
    around as a (pointer to a) value. This takes care of threading issues
    (because MVars work as a lock) and ensures that we don't accidentally use
    the wrong env.
  * Renamed the imp-to-llvm translation pass to `ImpToLLVM` instead of `JIT` (we
    had so many files called `JIT.hs`. In this change, we remove three of them!)

Still to do
  * Handle benchmark path (required before merging)
  * Handle teardown properly (required before merging)
  * The standalone function path is still partly separate from the top-level
    expression path because the latter operates on blocks instead of functions.
    And some dex-level optimizations don't even get run on the standalone
    function path currently because they're only set up to work on blocks. I
    think we should merge them completely, and treat top-level expressions as
    nullary standalone functions that get immediately called. We can do that in
    a follow-up PR.


src/lib/TopLevel.hs
```diff
@@ -629,17 +630,25 @@ compileTopLevelFun fname = do
...
+      extendLoadedObjects fname f
+      return f
+
+linkFunObjCode :: FunObjCode -> ExplicitLinkMap -> IO NativeFunction
+linkFunObjCode obj depPtrs = do
+  l <- createLinker
+  addExplicitLinkMap l depPtrs
+  addObjectFile l (objectFileContents obj)
+  ptr <- getFunctionPointer l (objectFileMainFunName obj)
+  let destructor = destroyLinker l
```
apaszke, line 650: What about the destructors?

src/lib/TopLevel.hs
```diff
@@ -665,42 +674,50 @@ forceDeferredInlining v =
...
   resultValsNoPtrs <- mapM litValToPointerlessAtom resultVals
   applyNaryAbs reconAtom $ map SubstVal resultValsNoPtrs
 {-# SCC evalLLVM #-}
 
+compileToObjCode :: Topper m => LLVM.AST.Module -> LocalCName -> CNameMap e n -> m n FunObjCode
+compileToObjCode llvmAST mainFunName (CNameMap names) = do
+  logger  <- getFilteredLogger
+  opt <- getLLVMOptLevel <$> getConfig
+  objCodeBytes <- liftIO $ compileLLVM logger opt llvmAST mainFunName
+  return $ FunObjCode objCodeBytes mainFunName (M.keys names) [] -- TODO: destructors!
```
apaszke, line 713: We should resolve this TODO before merging. We have all the relevant code!

src/Dex/Foreign/JIT.hs
```diff
@@ -123,9 +67,9 @@ dexFreeFunctionSignature sigPtr = do
...
-dexUnload jitPtr funcPtr = do
-  ForeignJIT{..} <- fromStablePtr jitPtr
-  addrTable <- readIORef addrTableRef
-  LLVM.JIT.unloadNativeModule $ nativeModule $ addrTable M.! funcPtr
-  modifyIORef addrTableRef $ M.delete funcPtr
+dexUnload :: Ptr Context -> ExportNativeFunctionAddr -> IO ()
+dexUnload ctxPtr funcPtr = do
+  Context _ _ ptrTabMVar <- fromStablePtr ctxPtr
+  addrTable <- takeMVar ptrTabMVar
+  putMVar ptrTabMVar $  M.delete funcPtr addrTable
```
apaszke, line 74: Since `insertIntoNativeFunctionTable` is a method of `Context`, perhaps make deletion from the table another method, instead of inlining it here? Otherwise it's inconsistent

python/dex/api.py
```diff
@@ -82,34 +80,23 @@ def dex_func(name, *signature):
...
 
 init()
-jit = createJIT()
-nofree = False
-@atexit.register
-def _teardown():
-  global nofree
-  destroyJIT(jit)
-  fini()
-  nofree = True  # Don't destruct any Haskell objects after the RTS has been shutdown
```
apaszke, line 111: Why is this gone? This is actually very important

python/dex/__init__.py
```diff
@@ -32,8 +34,7 @@ def _from_ptr(cls, ptr):
     return self
 
   def __del__(self):
-    if api is None or api.nofree: return
-    api.destroyContext(self)
+    pass  # TODO: free
```
apaszke, line 37: Why can't we call `destroyContext` anymore?

src/lib/Types/Core.hs
```diff
@@ -415,19 +421,27 @@ data Cache (n::S) = Cache
...
-type FunObjCodeContents = BS.ByteString
-data FunObjCode n = FunObjCode
-  { objectFileContents :: FunObjCodeContents
-  , objectFileNameMap  :: CNameMap FunObjCodeName n }
+type FunObjCodeNameMap = CNameMap FunObjCodeName
+data FunObjCode = FunObjCode
+  { objectFileContents    :: BS.ByteString
+  , objectFileMainFunName :: LocalCName
+  , objectFileFreeVars    :: [LocalCName]
+  , objectFileDtorList    :: [LocalCName] }
```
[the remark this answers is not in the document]
dougalm: I initially had them bundled, but this turned out to be a useful object by itself. Think of it as a closed lambda term, which accepts these link-time objects as parameters. Being closed means we don't need to worry about free names. I'll add some explanation though. And we actually do use it that way. We use the same object code with different link-type values when we start a new process.

python/dex/__init__.py
```diff
@@ -60,8 +59,6 @@ class Prelude(Module):
   __slots__ = ()
   def __init__(self):
     self._as_parameter_ = api.createContext()
-    if not self._as_parameter_:
-      api.raise_from_dex()
```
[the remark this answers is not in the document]
dougalm: thanks, that was a mistake

python/dex/__init__.py
dougalm: also a mistake

src/lib/LLVM/Link.hs
```diff
@@ -0,0 +1,87 @@
...
+  getDataLayout l = OrcJIT.getDataLayout $ linkerIRLayer l
+  getMangler    l = OrcJIT.getMangler    $ linkerIRLayer l
+
+type CName = String
+
+type ExplicitLinkMap = M.Map CName (Ptr ())
+
+createLinker :: IO Linker
+createLinker = do
+  -- TODO: cleanup for TargetMachine?
```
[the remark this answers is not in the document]
dougalm: We don't store the `TargetMachine` in `Linker` directly. It's an argument to `OrcJIT.createIRCompileLayer`, and the `IRCompiler` layer gets cleaned up along with the execution session, I think?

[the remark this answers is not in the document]
apaszke: The `IRCompileLayer` does get cleaned up, but I'm pretty sure the target machine doesn't!

src/lib/ImpToLLVM.hs
```diff
@@ -1252,6 +1226,27 @@ typeOf x = case fst $ MB.runModuleBuilder emptyState $ L.typeOf x of
...
+--     -- so we have to find and call them ourselves.
+--     dtorNames = do
+--       let dtorStructs = flip foldMap (LLVM.AST.moduleDefinitions ast) \case
+--             LLVM.AST.GlobalDefinition
+--               LLVM.AST.GlobalVariable{
+--                 name="llvm.global_dtors",
+--                 initializer=Just (C.Array _ elems)} -> elems
+--             _ -> []
+--       -- Sort in the order of decreasing priority!
+--       fmap snd $ sortBy (flip compare) $ flip fmap dtorStructs $
```
apaszke, line 1248: Is this meant to be commented out?

python/dex/api.py
dougalm: Fixed (after discussing with you offline)

src/lib/LLVM/Link.hs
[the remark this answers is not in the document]
dougalm: Yep, you're right. I'll track it separately and destroy it explicitly.

PR #1109, opened 2022-10-06
Add a new effect: `InitEffect`

normanrink:
The `InitEffect` is intended to express writing to a memory reference. This is more specific than a general IO effect. 

With the `InitEffect`, we can allow vectorization of non-pure loops, i.e. also of loops that carry an `InitEffect` only.

Note that the only place where an `InitEffect` occurs at the moment is on a `Seq` loop that is the result of lowering a `For` loop.


src/lib/Lower.hs
```diff
@@ -559,7 +559,7 @@ traverseFor maybeDest dir ixDict lam@(LamExpr (LamBinder ib ty arr eff) body) =
   ixDict' <- substM ixDict
   ty' <- substM ty
   eff' <- substM $ ignoreHoistFailure $ hoist ib eff
-  let eff'' = extendEffRow (S.singleton IOEffect) eff'
+  let eff'' = extendEffRow (S.singleton InitEffect) eff'
```
apaszke, line 562: There is one more reference to IOEffect in this file that you will need to replace, in the code that handles Ix dicts. Actually, we might need to add `RunInit`, as a handler counterpart to `RunIO` that I used in there

apaszke (commented):
Mostly LGTM, but there is one more place that needs to be modified.

Also note that the PR description is not exactly correct, as the Init effect actually has nothing to do with Seq loops. It comes from the `Place` operations, and we usually only generate `Place` when we `Seq`uentialize for loops, the `Init` on `Seq` comes from the embedded `Place`s. Similarly, if the `Seq` is nested inside another higher order function (e.g. a state effect handler) then the body of that higher order function has an `Init` effect too.

apaszke (commented):
Ah! You also need to modify [QueryType](https://github.com/google-research/dex-lang/blob/main/src/lib/QueryType.hs#L796) and [CheckType](https://github.com/google-research/dex-lang/blob/main/src/lib/CheckType.hs#L756) so that the `Place` op now has the `InitEffect` instead of `IOEffect`!

And [this](https://github.com/google-research/dex-lang/blob/main/src/lib/Lower.hs#L543) is the other use of `IOEffect` that has to be removed. The context is that we are modifying the implementations of `Ix` methods, which might make them `Init`-effectful. But that doesn't preserve their type signature (they have to be pure!). So, if we detect that the new body becomes effectful, we wrap it in a `RunInit` (well, `RunIO` right now) to discharge the effect. It's safe, since all initializations made in the body of the method don't influence any dests from outside of the method.

apaszke (approved):

PR #1113, opened 2022-10-07
Add an ADT to represent the different kinds of projections

apaszke:
Our current approach (using integers) was simple, but it was erasing lots of information. For example 0 could mean that we're taking a first element of a product type, but it could also mean that we're unwrapping a newtype. This change adds an ADT to start distinguishing between a few different kinds of projections, namely: product projections and two flavors of newtype projections: base and compound newtype unwraps. The difference between them is pretty arbitrary and corresponds to what the newtype erasure pass will eliminate (only compound newtypes). The set of base newtype is small and currently is only `Nat` and `Fin`.

### DIFF
--- a/src/lib/Types/Core.hs
+++ b/src/lib/Types/Core.hs
@@ -25,6 +25,7 @@ import Control.Monad.Writer.Strict (Writer, execWriter, tell)
 import Data.Word
 import Data.Maybe
 import Data.Functor
+import Data.Hashable
 import Data.List.NonEmpty (NonEmpty (..))
 import qualified Data.List.NonEmpty    as NE
 import qualified Data.Map.Strict       as M
@@ -73,7 +74,7 @@ data Atom (n::S) =
  -- a statically-known atom. This is because the variable name used
  -- here may also appear in the type of the atom. (We maintain this
  -- invariant during substitution and in Builder.hs.)
- | ProjectElt (NE.NonEmpty Int) (AtomName n)
+ | ProjectElt (NE.NonEmpty Projection) (AtomName n)
    deriving (Show, Generic)
 
 data Expr n =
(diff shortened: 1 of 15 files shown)


src/lib/Types/Core.hs
```diff
@@ -211,6 +212,12 @@ data PiType  (n::S) where
 data DepPairType (n::S) where
   DepPairType :: Binder n l -> Type  l -> DepPairType n
 
+data Projection
+  = UnwrapCompoundNewtype  -- Unwrap TypeCon, record or variant
+  | UnwrapBaseNewtype      -- Unwrap Fin or Nat
+  | ProjectProduct Int
```
dougalm, line 218: I assume this last case includes user-defined "newtypes", i.e. ADTs with a single unary constructor?

dougalm (approved):
Looks good! This should help with debugging too.

src/lib/Types/Core.hs
apaszke: No! The only projection valid for such an ADT is `UnwrapCompoundNewtype`, which returns a value of its representation type. For a unary ADT its representation is the payload, so that's the only thing you need to do.

dougalm: I see. So to project out the `Int` from a `data MyPair = MkMyPair Int Float` you'd use `[UnwrapCompoundNewType, ProjectProduct 0]`, and to project the `Int` from `MyNewtype = MkMyNewtype Int` you'd just use `[UnwrapCompoundNewtype]`? At some point in the past, we would have used `[0]` in both cases, but I suppose this change already happened with your newtype patch, just with integer indices instead of `Projection`.

All sounds good. Maybe add a comment to remind me that `UnwrapCompoundNewtype` unwraps both newtype-like ADTs and ordinary ADTs.

apaszke: The lengths of projection lists are the same as they were before! While `[0]` extracted an `Int` from `MyNewtype` (and now is `[UnwrapCompoundNewtype]`), you needed `[0, 0]` to get an `Int` out of `MkMyPair` (which now is `[ProjectProduct 0, UnwrapCompoundNewtype]`). This patch basically distinguishes the `0`s used for newtype unwrapping from the zeros used for product projections, but doesn't change the number of zeros you need.

PR #1116, opened 2022-10-10
Treat pointer literals as link-time constants

dougalm:
This allows standalone functions to close over array data.


src/lib/TopLevel.hs
```diff
@@ -693,17 +696,17 @@ evalLLVM :: (Topper m, Mut n) => IxDestBlock n -> m n (Atom n)
 evalLLVM block = do
   backend <- backendName <$> getConfig
   logger  <- getFilteredLogger
-  (blockAbs, ptrVals) <- abstractPtrLiterals block
   let (cc, _needsSync) = case backend of LLVMCUDA -> (EntryFun CUDARequired   , True )
                                          _        -> (EntryFun CUDANotRequired, False)
   ImpFunctionWithRecon impFun reconAtom <- checkPass ImpPass $
-                                             toImpFunction backend cc blockAbs
+    toImpFunction backend cc $ Abs Empty block
```
apaszke, line 702: Perhaps remove the `Abs` too, since it's always empty?

apaszke (approved):
Nice! LGTM, but it would be nice to get rid of the pointer abs in `toImpFunction`. I don't think we use it

src/lib/Util.hs
```diff
@@ -141,6 +141,15 @@ mapMaybe f (x:xs) = let rest = mapMaybe f xs
...
                         Nothing -> rest
 
+forMFilter :: Monad m => [a] -> (a -> m (Maybe b)) -> m [b]
+forMFilter [] _ = return []
+forMFilter (x:xs) f = do
+  ans <- f x
+  rest <- forMFilter xs f
+  return case ans of
+    Just y  -> y : rest
+    Nothing -> rest
```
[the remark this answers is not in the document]
dougalm: Good idea

src/lib/TopLevel.hs
dougalm: Yes, good point. I had thought that this was used in the standalone function path too but I was wrong.

PR #1120, opened 2022-10-12
Add a mechanism for dynamically scoped runtime variables.

dougalm:
The initial use case is to store the current output stream. Previously, we passed it as an explicit argument to the top-level entry function, but that meant we couldn't throw errors from non-inlined functions. Fixing that using explicit arguments would mean adding the output stream as an argument to *every* function. Why can't we just treat the output stream as a link-time constant? Well, although it's a constant for a particular top-level block's execution, different top-level blocks need to use different output streams, even though they may enter exactly the same chunk of executable code when they call a non-inlined function.

Instead, we treat the output stream as a dynamically scoped variable. We implement it using thread-local storage (TLS) via the pthreads API.

### DIFF
--- a/src/lib/ImpToLLVM.hs
+++ b/src/lib/ImpToLLVM.hs
@@ -26,7 +26,6 @@ import qualified LLVM.AST.ParameterAttribute as L
 import qualified LLVM.AST.FunctionAttribute as FA
 import qualified LLVM.IRBuilder.Module as MB
 
-
 import System.IO.Unsafe
 import qualified System.Environment as E
 import Control.Monad
@@ -118,7 +117,7 @@ impToLLVM logger fNameHint (ClosedImpFunction funBinders ptrBinders impFun) = do
   let dtorDef = defineGlobalDtors globalDtors
   let resultModule = L.defaultModule
         { L.moduleName = "dexModule"
-        , L.moduleDefinitions = dtorDef : funDefns ++ ptrDefns ++ defns ++ externDefns }
+        , L.moduleDefinitions = dtorDef : funDefns ++ ptrDefns ++ defns ++ externDefns ++ dyvarDefs }
   let dtorNames = map (\(L.Name dname) -> C8BS.unpack $ SBS.fromShort dname) globalDtors
   return $ WithCNameInterface resultModule fName funNames ptrNames dtorNames
   where
@@ -195,7 +194,7 @@ compileFunction logger fName env fun@(ImpFunction (IFunType cc argTys retTys)
     void $ extendSubst (bs @@> map opSubstVal argOperands) $ compileBlock body
     mainFun <- makeFunction fName argParams (Just $ i64Lit 0)
     extraSpecs <- gets funSpecs
-    return ([L.GlobalDefinition mainFun, outputStreamPtrDef], extraSpecs, [])
+    return ([L.GlobalDefinition mainFun], extraSpecs, [])
   CEntryFun -> liftCompile CPU env $ do
     (argParams, argOperands) <- unzip <$> traverse (freshParamOpPair [] . scalarTy) argTys
     unless (null retTys) $ error "CEntryFun doesn't support returning values"
@@ -203,7 +202,7 @@ compileFunction logger fName env fun@(ImpFunction (IFunType cc argTys retTys)
     void $ extendSubst (bs @@> map opSubstVal argOperands) $ compileBlock body
     mainFun <- makeFunction fName argParams (Just $ i64Lit 0)
     extraSpecs <- gets funSpecs
-    return ([L.GlobalDefinition mainFun, outputStreamPtrDef], extraSpecs, [])
+    return ([L.GlobalDefinition mainFun], extraSpecs, [])
   EntryFun requiresCUDA -> liftCompile CPU env $ do
     (streamFDParam , streamFDOperand ) <- freshParamOpPair attrs $ i32
     (argPtrParam   , argPtrOperand   ) <- freshParamOpPair attrs $ hostPtrTy i64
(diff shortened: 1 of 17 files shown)


src/lib/dexrt.cpp
```diff
@@ -51,6 +51,12 @@ int64_t dex_allocation_size (char* ptr) {
   return *(reinterpret_cast<int64_t*>(ptr - alignment));
 }
 
+void* dex_pthread_key_create () {
+  pthread_key_t* key_ptr = (pthread_key_t*) malloc(sizeof(pthread_key_t));
+  pthread_key_create(key_ptr, NULL);
```
apaszke, line 56: You could use the destructor to free the key pointer. It leaks memory right now.

src/lib/Types/Core.hs
```diff
@@ -318,7 +318,8 @@ data TopEnv (n::S) = TopEnv
   , envCustomRules :: CustomRules n
   , envCache :: Cache n
   , envLoadedModules :: LoadedModules n
-  , envLoadedObjects :: LoadedObjects n }
+  , envLoadedObjects :: LoadedObjects n
+  , envDynamicVarStores :: DynamicVarStores }
```
apaszke, line 322: The fact that this made its was all the way to the `Env` is very surprising and feels like a mistake to me. I thought we're just going to manage that within the topper monad and nothing else in Dex should ever have to think about this. Access to dynamically scoped values is realized through the special primitive ops, so you should never have to carry this here.

src/lib/Runtime.hs
```diff
@@ -160,3 +160,16 @@ readStream h action = go
...
+-- We use TLS to implement some dynamically scoped variables, like the file
+-- descriptor for the current output stream. We allocate the thread-local memory
+-- here, and we generate code for `pthread_setspecific` at entry functions and
+-- `pthread_getspecific` at use sites.
+
+foreign import ccall "dex_pthread_key_create"
+  pthreadKeyCreate :: IO (Ptr ())
+
+allocateTLS :: IO (Ptr (Ptr ()))
+allocateTLS = castPtr <$> pthreadKeyCreate
```
apaszke, line 175: This is a bad name, since it doesn't actually allocate the TLS. It only allocates a key that lets you reach the TLS.

apaszke (changes requested):
I really don't like how we end up having to carry those vars in the env even though 95% of the compiler really doesn't care. Can't we limit modeling the TLS to just the neighborhood of LLVM?

src/lib/dexrt.cpp
dougalm: As discussed, it's a tiny, fixed amount of memory once per process, so we don't need to worry about destructors too much and it's unclear when we'd call them anyway, except for at program termination. I'll add a comment to remind us that we're being a bit naughty.

src/lib/Types/Core.hs
dougalm: Yes, I think I agree. I'll move things around so that we store the dynamic vars (and maybe even the other pointers) somewhere separate.

src/lib/ImpToLLVM.hs
```diff
@@ -1096,10 +1077,48 @@ pointerType = const L.PointerType
...
+  key <- getDyvarKey v
+  emitExternCall pthread_getspecific [key]
+
+getDyvarKey :: LLVMBuilder m => DynamicVar -> m Operand
+getDyvarKey v = do
+  let ref = globalReference (hostPtrTy hostVoidp) $ dynamicVarLName v
+  load hostVoidp (L.ConstantOperand ref)
+
+pthread_setspecific :: ExternFunSpec
+pthread_setspecific = ExternFunSpec "pthread_setspecific" i32 [] [] [hostVoidp, hostVoidp]
```
apaszke, line 1118: This code makes me a little nervous. Those pthread functions take the key as a value, and you've declared it as a `void*` here. But it's often not defined as a `void*`! if you look [here](https://elixir.bootlin.com/glibc/latest/source/sysdeps/htl/bits/types/__pthread_key.h#L22) it's actually an `int`!

One reasonable way to handle the vagueness of the pthread standard would be to assume that it's an int here, and then include a `static_assert` in `dexrt` that verifies that it truly is an int. So, in this file:
```hs
-- This is actually implementation-dependent, but we verify that it's true in dexrt.cpp
pthreadKeyTy :: L.Type
pthreadKeyTy = i32
```
and then in `dexrt.cpp`:
```cpp
#include <type_traits>
static_assert(std::is_same<pthread_key_t, int>::value, "Expected pthread_key_t to be an int!");
static_assert(sizeof(int) == 4, "Expected int to be 4B!");
```

src/lib/ImpToLLVM.hs
```diff
@@ -1096,10 +1077,48 @@ pointerType = const L.PointerType
...
+  setDyvar OutStreamDyvar streamPtr
+
+-- === Dynamically-scoped vars via thread-local pthreads vars ===
+
+dyvarDefs :: [L.Definition]
+dyvarDefs  = (:[]) $ L.GlobalDefinition $ L.globalVariableDefaults
+  { L.name = dynamicVarLName OutStreamDyvar
+  , L.type' = hostVoidp
+  , L.linkage = L.External
+  , L.initializer = Nothing }
```
apaszke, line 1097: You could just make it a list comprehension over the `DynamicVar` enum, right?

apaszke (commented):
Looks much better! Still, there are a few other things we should do to harden this implementation.

src/lib/Runtime.hs
```diff
@@ -160,3 +160,16 @@ readStream h action = go
...
+-- We use TLS to implement some dynamically scoped variables, like the file
+-- descriptor for the current output stream. We allocate the thread-local memory
+-- here, and we generate code for `pthread_setspecific` at entry functions and
+-- `pthread_getspecific` at use sites.
+
+foreign import ccall "dex_pthread_key_create"
+  pthreadKeyCreate :: IO (Ptr ())
+
+createTLS :: IO (Ptr (Ptr ()))
+createTLS = castPtr <$> pthreadKeyCreate
```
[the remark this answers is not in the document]
dougalm: Good idea. I like that zero-constructor ADT trick.

src/lib/ImpToLLVM.hs
dougalm: Nice. Good idea

src/lib/ImpToLLVM.hs
dougalm: Great catch. Thanks for figuring that out. I'll do as you suggest.

apaszke (approved):
Looks great!

PR #1124, opened 2022-10-14
Ix dicts as top-level specializations

dougalm:
Handle Ix dict hoisting in the same way we handle function specialization.

The problems of Ix dict hoisting and function specialization are very similar.
In both cases, we want to emit named chunks of code to avoid the code blowups
we'd get if we inlined them instead. We want to de-dup where possible by using
the same names in several places. We need a way to run the rest of the compiler
passes on the chunks. And we don't learn which chunks we need until
simplification, where we finally have concrete type and dictionary arguments.

Previously, we handled Ix dict hoisting and function specialization in different
ways. For function specialization, we'd gather the required specializations
during simplification. On returning to the top level, we'd lower the specialized
functions all the way to object code, before proceeding with the rest of the
passes on the original code. For Ix dict hoisting, we'd gather the required Ix
dicts in a separate pass, and then keep these dicts alongside the original code
through the subsequent passes, until Imp where they get inlined and translated.
This meant that each pass needed to handle Ix dicts specially. It also meant
that we didn't handle Ix dicts at all for standalone functions.

This patch treats Ix dicts just like function specialization, except that
instead of going all the way to object code, we only go as far as Imp, at which
point we inline them (so that LLVM can do further optimizations). This makes the
code simpler (less 300 loc) and it makes it possible to merge the separate block
evaluation and function compilation top-level paths, at which point we'll be
able to do lowering and optimization for standalone functions.


src/lib/Types/Core.hs
```diff
@@ -295,6 +299,13 @@ data DictExpr (n::S) =
  | SuperclassProj (Atom n) Int  -- (could instantiate here too, but we don't need it for now)
    -- Special case for `Ix (Fin n)`  (TODO: a more general mechanism for built-in classes and instances)
  | IxFin (Atom n)
+   -- Used for dicts defined by top-level functions, which may take extra data parameters
+   -- TODO: consider bundling `(DictType n, [AtomName n])` as a top-level
+   -- binding for some `Name DictNameC` to make the IR smaller.
+   -- TODO: the function atoms should be names. We could enforce that syntactically but then
+   -- we can't derive `SubstE AtomSubstVal`. Maybe we should have a separate name color for
+   -- top function names.
+ | ExplicitMethods (DictType n) [Atom n] [Atom n] -- dict type, names of parameterized method functions, parameters
```
apaszke, line 308: Don't we have another way to create an explicit dict? perhaps we should unify those two?

apaszke: Also, just to clarify: names of parameterized method functions here refer to the fact that those `Atom`s are really `AtomName`s?

src/lib/Name.hs
```diff
@@ -3114,6 +3128,10 @@ unsafeCoerceVC :: forall c' (v::V) c o. v c o -> v c' o
 unsafeCoerceVC = TrulyUnsafe.unsafeCoerce
 {-# NOINLINE [1] unsafeCoerceVC #-}
 
+unsafeCoerceE1 :: forall (m::S -> * -> *) (n1::S) (n2::S) (a:: *). m n1 a -> m n2 a
```
apaszke, line 3131: nit: perhaps `M1` instead of `E1`?

src/lib/TopLevel.hs
```diff
@@ -909,7 +936,57 @@ getFilteredLogger = do
...
+
+hoistIxDicts
+  :: ( Topper m, Mut n, GenericallyTraversableE e
+     , SinkableE e, SubstE Name e, HoistableE e)
+  => e n -> m n (e n)
+hoistIxDicts e = do
+  Abs frag (PairE e' (IxHoistingState UnitE)) <-
+    liftGenericTraverserMTopEmissions (IxHoistingState UnitE) $ traverseGenericE (sink e)
+  e'' <- emitEnv (Abs frag e')
+  evalRequiredSpecializations e''  -- track the needed sepcializations rather than traversing again to find them
```
apaszke, line 949: Is that comment a TODO, or am I missing something?

src/lib/TopLevel.hs
```diff
@@ -909,7 +936,57 @@ getFilteredLogger = do
...
+    DictCon d -> do
+      d' <- substM d
+      dTy <- getType d'
+      DictCon <$> case dTy of
+        DictTy dTy'@(DictType "Ix" _ _) -> emitIxDictSpecialization dTy' d'
+        _ -> return d'
+    atom -> traverseAtomDefault atom
+
+emitIxDictSpecialization :: HoistingTopBuilder m => DictType n -> DictExpr n -> m n (DictExpr n)
+emitIxDictSpecialization _ d@(ExplicitMethods _ _ _) = return d -- Nothing to do if we already have explicit methods
```
apaszke, line 966: Does that ever happen? Perhaps for top-level results that got inlined? We might consider adding a TODO to actually emit the dicts that top-level bound atoms refer to into the top-level too.

src/lib/TopLevel.hs
```diff
@@ -909,7 +936,57 @@ getFilteredLogger = do
   logger  <- getLogger
   return $ FilteredLogger shouldLog logger
 
--- -- === instances ===
+-- === hoisting Ix dicts ===
+
+hoistIxDicts
```
apaszke, line 941: Perhaps move this to Lower.hs?

src/lib/TopLevel.hs
```diff
@@ -624,12 +632,23 @@ execUDecl mname decl = do
     UDeclResultDone sourceMap' -> emitSourceMap sourceMap'
 {-# SCC execUDecl #-}
 
+lowerTopLevelFun :: (Topper m, Mut n) => AtomName n -> m n ()
+lowerTopLevelFun fname = do
+  fPreSimp <- specializedFunPreSimpDefinition fname
+  fSimp <- simplifyTopFunction fPreSimp
+  evalRequiredSpecializations fSimp
+  fSimp' <- hoistIxDicts fSimp
+  -- TODO: actually do the lowering (lowering only handles blocks, not functions, right now)
```
apaszke, line 641: So if we plug in the destination passing etc. here, then it will also apply to Ix dicts?

src/lib/TopLevel.hs
```diff
@@ -588,10 +589,17 @@ evalRequiredSpecializations
...
   forM_ (freeAtomVarsList e) \v -> do
     lookupAtomName v >>= \case
-      TopFunBound _ (SpecializedTopFun _) ->
+      TopFunBound _ (SpecializedTopFun (AppSpecialization _ _)) ->
         queryImpCache v >>= \case
           Nothing -> compileTopLevelFun v
           Just _ -> return ()
+      -- XXX: we might need to toposort these based on dependencies. Or else
+      -- maybe we can be careful about the order in which we request them in the
+      -- first place (rather than querying free vars).
```
apaszke, line 598: Can you elaborate?

src/lib/Types/Core.hs
```diff
@@ -556,18 +572,26 @@ data AtomBinding (n::S) =
...
- | SpecializedTopFun (SpecializationSpec n) -- Original (unspecialized) function, specialization args
- | SimpTopFun          (NaryLamExpr n)
- | FFITopFun           (ImpFunName n)
+   -- This is for functions marked `@noinline`, before we've seen their use
+   -- sites and discovered what arguments we need to specialize on.
+   AwaitingSpecializationArgsTopFun Int (Atom n)
+   -- Specification of a specialized function. We still need to simplify, lower,
+   -- and translate-to-Imp this function. When we do that we'll store the result
+   -- in the `impCache`. Or, if we're dealing with ix method specializations, we
+   -- won't go all the way to Imp and we'll store the result in
```
apaszke, line 581: This seems contradictory with the fact that you do extend the imp cache in `evalRequiredSpecializations`?

src/lib/TopLevel.hs
```diff
@@ -569,15 +570,15 @@ evalBlock typed = do
...
+  simp' <- hoistIxDicts simp
+  opt <- whenOpt simp' $ checkPass OptPass . optimize
   result <- case opt of
     AtomicBlock result -> return result
     _ -> do
-      explicitIx <- emitIx opt
-      instIx <- simplifyIx explicitIx
-      lowered <- checkPass LowerPass $ lowerFullySequential instIx
+      lowered <- checkPass LowerPass $ lowerFullySequential opt
+      evalRequiredSpecializations lowered
```
apaszke, line 579: I'm not loving the fact that this function is called from so many places. And even _twice_ in this function! I can't prove to myself that this set of calls is necessary or sufficient. What's the right way to think about this?

src/lib/Lower.hs
```diff
@@ -587,7 +139,7 @@ traverseTabCon maybeDest tabTy elems = do
     Just  d -> return d
     Nothing -> emitOp $ AllocDest tabTy'
   Abs bord ufoBlock <- buildAbs noHint IdxRepTy \ord -> do
-    buildBlockSimplified $ unsafeFromOrdinal (sink ixTy') $ Var $ sink ord
+    buildBlock $ unsafeFromOrdinal (sink ixTy') $ Var $ sink ord
```
apaszke, line 142: How come this works now? Can we handle method projections in Imp?

src/lib/GenericTraversal.hs
```diff
@@ -22,28 +22,46 @@ import MTL1
...
+  :: (EnvReader m, SinkableE e, SubstE Name e, SinkableE s, SubstE Name s)  => s n
+  -> (forall l. DExt n l => GenericTraverserM TopEnvFrag s l l (e l))
+  -> m n (Abs TopEnvFrag (PairE e s) n)
+liftGenericTraverserMTopEmissions s m =
+  liftM runHardFail $ liftDoubleBuilderT do
+    (e, s') <- runStateT1 (runSubstReaderT idSubst $ runGenericTraverserM' m) (sink s)
+    return $ PairE e s'
+{-# INLINE liftGenericTraverserMTopEmissions #-}
+
+newtype GenericTraverserM f s i o a =
```
apaszke, line 41: I'm not sure how I feel about this extra parameter. I am a little afraid that it will inhibit GHC optimizations and I want to double check that before we merge this. It also seems slightly overkill just for this one Ix hoisting pass. I thought we were going to do it inside simplification, where we don't use the generic traversal, so we wouldn't need this.

src/lib/CheckType.hs
```diff
@@ -421,6 +418,9 @@ dictExprType e = case e of
   IxFin n -> do
     n' <- checkTypeE NatTy n
     liftM DictTy $ ixDictType $ TC $ Fin n'
+  ExplicitMethods ty _ _ ->
+    -- TODO: check
```
apaszke, line 422: Please implement the check. You probably even have the code in the diff below

src/lib/Builder.hs
```diff
@@ -135,6 +135,7 @@ emitAtomToName x = emit (Atom x)
 class (EnvReader m, MonadFail1 m) => HoistingTopBuilder m where
   emitHoistedEnv :: (SinkableE e, SubstE Name e, HoistableE e)
                  => Abs TopEnvFrag e n -> m n (Maybe (e n))
+  willItHoist :: HoistableE e => e n -> m n Bool
```
apaszke, line 138: `canHoistToTop`? It's much more precise, since `willItHoist` can always return `True`: `e n` always hoists to `n` :)

src/lib/TopLevel.hs
```diff
@@ -624,12 +632,23 @@ execUDecl mname decl = do
...
 
+lowerTopLevelFun :: (Topper m, Mut n) => AtomName n -> m n ()
+lowerTopLevelFun fname = do
+  fPreSimp <- specializedFunPreSimpDefinition fname
+  fSimp <- simplifyTopFunction fPreSimp
+  evalRequiredSpecializations fSimp
+  fSimp' <- hoistIxDicts fSimp
+  -- TODO: actually do the lowering (lowering only handles blocks, not functions, right now)
+  fSimpName <- emitLoweredFun (getNameHint fname) fSimp'
+  extendIxLoweredCache fname fSimpName
```
apaszke, line 643: Why does this always extend the `IxLowered` cache when the function is called `lowerTopLevelFun`??

src/lib/Types/Core.hs
```diff
@@ -400,8 +411,13 @@ emptyImportStatus = ImportStatus mempty mempty
 
 -- TODO: figure out the additional top-level context we need -- backend, other
 -- compiler flags etc. We can have a map from those to this.
+
 data Cache (n::S) = Cache
   { specializationCache :: EMap SpecializationSpec AtomName n
+  , ixLoweredCache :: EMap AtomName AtomName n
```
apaszke, line 417: Can you add a comment that describes what this map is supposed to represent?

src/lib/Builder.hs
```diff
@@ -1273,49 +1320,67 @@ unsafePtrLoad x = do
...
-projMethod methodName dict = do
-  getType dict >>= \case
-    DictTy (DictType _ className _) -> do
-      methodIdx <- getMethodIndex className methodName
-      emitOp $ ProjMethod dict methodIdx
-    _ -> error $ "Not a dict: " ++ pprint dict
+-- XXX: it's important that we do the reduction here, because we need the
+-- property that if we call this with a simplified dict, we only produce
+-- simplified decls.
+applyIxMethod :: (Builder m, Emits n) => Atom n -> IxMethod -> [Atom n] -> m n (Atom n)
```
apaszke, line 1326: TODO for me: review this once I understand the `IxLoweredCache`

src/lib/Imp.hs
```diff
@@ -1310,7 +1308,7 @@ computeElemCount (EmptyAbs Empty) =
   return $ IdxRepVal 1
 computeElemCount idxNest' = do
   let (idxList, idxNest) = indexStructureSplit idxNest'
-  sizes <- forM idxList indexSetSizeFromSimp
+  sizes <- forM idxList indexSetSize
```
apaszke, line 1311: How does that work? `indexSetSize` doesn't look into the top level bindings and I thought that at this point the Ix dicts should be represented by names?

src/lib/Types/Core.hs
dougalm: > Don't we have another way to create an explicit dict? perhaps we should unify those two?

Yes, we previously had dicts representable by a newtype-cast of a product of methods. This change removes that version completely. Keeping them non-inlined like this should make the IR smaller.

> Also, just to clarify: names of parameterized method functions here refer to the fact that those Atoms are really AtomNames?

Correct.

src/lib/Name.hs
dougalm: Nice. Good idea.

src/lib/TopLevel.hs
dougalm: Yep, you're right. It was meant to be a TODO

src/lib/TopLevel.hs
dougalm: Good question. I'll check. To be clear, `ExplicitMethods` doesn't inline the methods exactly. It refers versions of them by name, and supplies the local data they need.

src/lib/TopLevel.hs
dougalm: Yes, sure. I might even try moving it to simplification as we'd initially discussed and as you suggest below. The reason I didn't put it in simplification was that I realized that simplification calls `substM` on types rather than traversing them, so we wouldn't find all the ix dicts. But maybe I should just fix that.

src/lib/TopLevel.hs
dougalm: Yes, if we want it to (or we can make it take a different path if we prefer).

src/lib/Types/Core.hs
dougalm: I think it's correct? The `IxMethodSpecialization` path of `evalRequiredSpecializations` calls `lowerTopLevelFun` which doesn't go all the way to imp.

src/lib/Lower.hs
dougalm: It's handled in Builder, in `applyIxMethod`. I wanted to avoid having Lower produce specialization requests, so I didn't want it to call back into simplification.

src/lib/GenericTraversal.hs
dougalm: I can try moving it to simplification. As I mentioned above the only reason was that I didn't want to turn all the `substM ty` lines in Simplify to `simplifyType ty` or `simplifyAtom ty` or something. I agree that we should check whether it inhibits optimizations. Even if we don't need it for ix hoisting (say, because we move that to simplification), I suspect we'll want to have the freedom to emit top-level emissions in other places too. But we can delete it until we need it if we don't need it now.

src/lib/TopLevel.hs
dougalm: You're right. This is confusing. For now, I should rename `lowerTipLevelFun` to `lowerIxMethodFun` or something.

src/lib/Imp.hs
dougalm: `indexSetSize` calls `applyIxMethod`, which does look at the top level bindings in the `ExplicitMethods` case.

src/lib/TopLevel.hs
dougalm: Looks like that code path does get hit in eval_tests. Is there a problem with it? I don't think I understand your suggestion about adding a TODO.

src/lib/TopLevel.hs
dougalm: Follow-up: yes, this particular call to `evalRequiredSpecializations` isn't needed.

src/lib/Types/Core.hs
apaszke: Right. I know that you removed the newtype, but we still have `Con (ExplicitDict ...)` which seems equivalent to what you have here.

src/lib/TopLevel.hs
apaszke: Right, if you are 1000% sure that we never inline into `ExplicitMethods` then this code might be ok. But given that this is not checked statically, I am not sure. Since Ix emissions are temporary, what happens when they're using in a result that we'll have to put in the top level env? You won't be able to hoist the result atom above the Ix.

src/lib/TopLevel.hs
apaszke: Right, I just want to avoid having TopLevel grow beyond the minimum. It's already a huge file and this really does not need to be there.

src/lib/Types/Core.hs
```diff
@@ -174,6 +174,10 @@ data LamExpr (n::S) where
 
 type IxDict = Atom
 
+-- TODO: use this in projectIxFinMethod too
```
apaszke, line 177: You already do, so please remove the TODO

src/lib/CheckType.hs
```diff
@@ -421,6 +418,19 @@ dictExprType e = case e of
...
+    ClassDef _ _ pbs (SuperclassBinders Empty _) methodTys <- lookupClassDef className
+    forMZipped_ methodTys methodFunNames \(MethodType _ methodTy) methodFunName -> do
+      reqTy <- checkedApplyNaryAbs (Abs pbs methodTy) params
+      let args' = case reqTy of
+            Pi _ -> args
+            _ -> args ++ [UnitVal]
+      methodFunTy <- getType =<< substM methodFunName
+      actualTy  <- checkApp methodFunTy args'
+      checkAlphaEq reqTy actualTy
+      -- TODO: something special with thunks/nullary methods?
```
apaszke, line 432: You already do something special about them? Please clarify the TODO or remove it

apaszke (commented):
This mostly LGTM, but I think there are still a few points I'd like to clarify. I resolved all the issues I felt were solved, but please do check the ones still open.

src/lib/Types/Core.hs
dougalm: I see. It's definitely not equivalent. `Con (ExplicitDict ...)` takes the method as an atom, and only allows single methods, whereas `ExplicitMethods` gives the data-parameterized methods as names, and supplies parameter. But we only use the `Con (ExplicitDict ...)` for a single purpose, which is to supply the `accumBaseMonoid` in `run_accum`, and maybe we can do the same thing by exposing surface syntax for single-method `ExplicitMethods`. I can try it.

src/lib/TopLevel.hs
dougalm: Ok, I'll change it to be an error, and I'll also make `ExplicitMethods` take names instead of atoms to make the no-inlining guarantee more explicit. I think it should just require a custom (crashing) substitution rule.

src/lib/TopLevel.hs
dougalm: Sure. I'm planning to move it to simplification in a follow-up patch. So let's leave it here until then.

src/lib/TopLevel.hs
dougalm: > I don't think the Ix typeclass structure is guaranteed to not form cycles?

Typeclasses with cycles are a problem for us in any case! I haven't tried, but I bet we can make the compiler go into an infinite loop during synthesis if we supply some adversarial instances. I don't think the problem here is any worse than what we already have.

src/lib/Builder.hs
```diff
@@ -145,43 +146,69 @@ liftTopBuilderHoisted cont = do
...
+  :: (EnvReader m, Fallible m') => DoubleBuilderT UnitB m' n a -> m n (m' a)
+liftDoubleBuilderTNoEmissions cont = do
+  env <- unsafeGetEnv
+  Distinct <- getDistinct
+  return do
+    Abs UnitB (DoubleInplaceTResult REmpty (LiftE ans)) <-
+      -- XXX: it's safe to use `unsafeCoerceM1` here. We don't need the rank-2
+      -- trick because we've specialized on the `UnitB` case, so there can't be
+      -- any top emissions.
+      runDoubleInplaceT env $ unsafeCoerceM1 $ runDoubleBuilderT' $ LiftE <$> cont
```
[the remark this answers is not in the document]
dougalm: It's just avoiding the rank-2 trick (i.e. `(forall l. DExt n l => ....)` , as the coment says

src/lib/Types/Core.hs
dougalm: On second thoughts, I'm not convinced this will be a net win, rather than just churn. `Con (ExplicitDict ...)` is a one-of hack to handle `run_accum`. But doing the same thing using `ExplicitMethods` is no less hacky. We'll have to add concrete syntax (it's not a regular primop), a custom inference rule, and we'll have to deal with `ExplicitMethods` appearing in pre-simplification IR, which we don't currently. So I think it's just trading one hack for another. If we're worried about the confusion of the name similarity, we could rename `Con (ExplicitDict ..)` to `Con (ExplicitAccumBaseMonoidDict _)`.

src/lib/TopLevel.hs
apaszke: Oh, so what's you're saying is that this patch emits all of them permanently, which is not good, but it will be ok once the other patch lands?

src/lib/Types/Core.hs
apaszke: Ok, fine.

src/lib/TopLevel.hs
apaszke: Yeah I think you're right. In that case I would suggest removing the comment, because IMO it is only more confusing than helpful (we've established that when things are a DAG, it should all work out, and when they're not we're screwed in many other ways too)

src/lib/Builder.hs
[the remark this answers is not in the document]
apaszke: I am asking, because I don't understand the comment :P

apaszke (approved):

src/lib/TopLevel.hs
dougalm: Yes, that's right.

PR #1130, opened 2022-10-18
Update isomorphisms.dx comments to snake_case

niklasschmitz:
This PR updates the isomorphisms.dx example comments for snake_case (`appIso` -> `app_iso` etc) to match the code, as introduced in #814.


axch (approved):
Thanks!  Looks great!

PR #1134, opened 2022-10-20
Format TopLevel.hs

axch:
Added line breaks to keep TopLevel.hs to at most 80 characters per line, because (i) that's Google style across all languages, and (ii) 80 characters is about the width I can get without wrapping when trying to show two files with vertical split-screen on video calls and have the code be even remotely legible on the other end.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,5 @@
+# Run this command to always ignore these in local `git blame`:
+# git config blame.ignoreRevsFile .git-blame-ignore-revs
+
+# Formatted TopLevel.hs to 80 character width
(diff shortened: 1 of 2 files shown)


apaszke (changes requested):
FWIW I don't particularly care about Google style, since that's just a bunch of arbitrary choices. And while lowering the column width makes it easier to stack more windows horizontally, it means that you do see significantly less code in every single window, since there are more line breaks! If you're really concerned, why not enable wrapping lines in your editor?

axch:
Fixed commit history issue by adding the formatting commit to an ignore list (that Github apparently respects by default!)

Issue #1123, opened 2022-10-13
Can't interleave class constraints in custom datatype declarations


duvenaud:
Right now I can't find a way to include tables in custom datatypes, because I can't express the `Ix` constraint before the end of the type declaration.

Here is the closest I got:
```
data Permuted n:Type [Ix n] (perm:n=>n) =
  MkPermuted n
```
which gives
```
MonadFail error (internal error)Invalid class constraint list; expecting one or more bracketed groups

data Permuted n:Type [Ix n] (perm:n=>n) =
                            ^^^^^^^^^^^^
```

PR #1135, opened 2022-10-24
Interleave classes in data definitions

### LINKED ISSUE #1015: Syntax for array-kinded type parameters in `data`
In #733, @duvenaud constructed a data definition one of whose type-level arguments itself had an array as a type:
```
data TableInType n a table:(n=>a) [Ix n]   =
  MkTableInType              -- Doesn't store any data except in the type!
```

Based on comments on that issue, it seems like this worked at some point, but now fails with
```
> Type error:Couldn't synthesize a class dictionary for: (Ix n)
>
> data TableInType n a table:(n=>a) [Ix n]   =
>                             ^^^^
```

However, moving the `[Ix n]` constraint before the `table` binder also fails, with
```
data TableInType n a [Ix n] table:(n=>a) =
  MkTableInType              -- Doesn't store any data except in the type!
> Parse error:15:29:
>    |
> 15 | data TableInType n a [Ix n] table:(n=>a) =
>    |                             ^
> unexpected 't'
> expecting '='
```
because the parser hardcodes class constraints being at the end.  We should probably fix this (unless there's another way to achieve the same effect?).

axch:
Fixes #1015.

This might unblock some of @duvenaud 's ideas about carrying sets or tables in types.


duvenaud:
Wonderful!  This also fixes #1123, and accidental duplicate I made of #1015.

axch:
On advice from @dougalm, replaced `(EitherB Binder Binder)` with `PiBinder` (and likewise in the pre-Inference data structure).  This reduces the number of defined types and simplifies the code in places.  For better or worse, it also creates a channel for the parser to create data binders with implicit or linear arrow arguments, even though status quo ante only allowed plain arrow and class arrow.  Inference does not currently have cases for handling implicit arrow and linear arrow.

A question remains: what actually cares about whether elements of `DataDefParams` refer to ordinary arguments or class arguments?  Before this refactor they were in separate lists, but now they are intermingled.  The only places I can find are `unifyZonked`, `alphaEqE`, and `hashWithSaltE`, all three of which ignore the class parameters with the comment "We ignore the dictionary parameters because we assume coherence".  Now that not ignoring is easier than ignoring, should we stop ignoring them?

src/lib/Types/Core.hs
```diff
@@ -131,22 +131,17 @@ data FieldRowElem (n::S)
...
-data DataDefBinders n l where
-  DataDefBinders
-    :: Nest Binder n h  -- ordinary params
-    -> Nest Binder h l  -- dict params
-    -> DataDefBinders n l
-
-data DataDefParams n = DataDefParams [Atom n] [Atom n]  -- ordinary params, dict params
-                       deriving (Show, Generic)
+-- Left is ordinary binders, Right is class binders
+newtype DataDefParams n = DataDefParams [EitherE Atom Atom n]
```
apaszke, line 143: This `EitherE` is a good candidate for a new ADT.

src/lib/Inference.hs
```diff
@@ -2473,7 +2505,8 @@ instance Unifiable IxType where
 
 instance Unifiable DataDefParams where
   -- We ignore the dictionaries because we assume coherence
-  unifyZonked (DataDefParams xs _) (DataDefParams xs' _) = zipWithM_ unify xs xs'
+  unifyZonked (DataDefParams xs) (DataDefParams xs')
+    = zipWithM_ unify (leftsE xs) (leftsE xs')
```
apaszke, line 2509: I am a little surprised that this still works, given that it actually does something different (see the comment). We should think hard and try to construct a case where this matters.

src/lib/Inference.hs
```diff
@@ -1639,6 +1647,30 @@ withSuperclassBindersRec (ty:rest) cont = do
...
+  Nest (UAnnBinderArrow b ty arr) rest -> do
+    Abs (b':>ty') (Abs rest' body) <-
+      case arr of
+        PlainArrow -> withUBinder (UAnnBinder b ty) id inner_c
+        ClassArrow -> withUBinder (UAnnBinder b ty)
+          (LamBound . LamBinding ClassArrow) inner_c
+        ImplicitArrow -> withUBinder (UAnnBinder b ty)
+          (LamBound . LamBinding ImplicitArrow) inner_c
+        LinArrow -> withUBinder (UAnnBinder b ty) id inner_c
+    return $ Abs (Nest (PiBinder b' ty' arr) rest') body
```
apaszke, line 1666: Wouldn't it be simpler to do
```hs
let binding = case arr of PlainArrow -> id; ClassArrow -> ...
Abs (b':>ty') (Abs rest' body) <- withUBinder (UAnnBinder b ty) binding do
  cont =<< ((:names) <$> sinkM name)
```

apaszke (approved):
LGTM, though the unification changes make me a little uneasy

src/lib/Types/Core.hs
axch: Replacing with [(Arrow, Atom n)]

src/lib/Inference.hs
axch: I believe this actually does the same thing it used to do, i.e., still ignores the dictionaries.  The `leftsE` calls filter the dictionaries out.

src/lib/Inference.hs
axch: Thanks!  Using `toBinding` instead of `id` (as @apaszke suggested offline) made the type-checker happy.

PR #1136, opened 2022-10-24
Propagate more name hints.

axch:
This change tries to increase the distance that user-supplied names make it through compiler passes.  It's far from complete, but suffices to get simplification to preserve the name given in the checked-in test case, which it didn't before.

The place where user-supplied names come from is decls (that were present in the source program).  Such a name names the value returned from the RHS of that decl.  So, this PR collects the name hint from a decl undergoing processing, passes it to the function processing the RHS, and tweaks said function to, if it emits a decl representing the value of that RHS, propagate said name hint to that emission.

Why do this now?  Because I want to write unit tests for the upcoming occurrence analysis that query the occurrence-analyzed IR for the annotation that goes with a variable of interest, and I want to name that variable using the surface name that appears in the string in the unit test.

### DIFF
--- a/src/lib/Algebra.hs
+++ b/src/lib/Algebra.hs
@@ -50,7 +50,7 @@ sumUsingPolys lim (Abs i body) = do
       Just poly' -> return $ Abs i' poly'
       Nothing -> throw NotImplementedErr $
         "Algebraic simplification failed to model index computations: " ++ pprint body'
-  limName <- emitAtomToName lim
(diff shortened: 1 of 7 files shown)


apaszke (commented):
While the change LGTM, I'm just not convinced that this is the right way to test those things. So far we didn't need it, and when I wanted to refer to a particular variable, I basically made a specific enough regex to match it, whatever its name was.

axch:
LGTM confirmed offline.

PR #1138, opened 2022-10-25
Add a role system and use it to guide generalization

dougalm:
The idea is that each parameter to a type constructor, class, or instance, must have one of three roles:
  1. type parameter role        (parameters of type `Type`)
  2. data parameter role        (parameters that are runtime-representable)
  3. dictionary parameter role  (parameters of type `DictTy _`)

When we generalize the arguments for a function specialization, we handle parameters according to their role. Data parameters become additional runtime arguments. Type parameters are traversed recursively. And dictionary parameters are generalized according to the required more general type.

For example, imagine we want to generalize the application of `sum : {a} [Ix a] (a=>Float) -> Float` applied to `(Fin 10)` and `IxFinInstance 10`. We traverse the type parameter, `Fin 10`, and find that it has a data parameter, `n:Nat`, which is instantiated at `10` here. We generalize it to `n`, so we have `\n:Nat. sum (Fin n)`. Then we consider the dictionary, `IxFinInstance 10`. With `sum` specialized at `Fin n` instead of `Fin 10`, we now require that the dictionary has type `Ix (Fin n)` instead of `Ix (Fin (10))`. We assume (and require) that instances are parametric in data parameters, so we know that we can use the same synthesis tree and we just need to update the parameters. We swap out data parameters in the instance for inference variables, `IxFinInstance ?1`, and unify the type, `Ix (Fin ?1)`, with the required `Ix (Fin n)`, which is solved by `?1 = n`. We end up generating a generalized-specialized top-level function defined by `\n. sum (Fin n) (IxFinInstance n)` as we hoped.

This change generalizes the function `generalizeType` which previously just special-cased `Fin n` and `Ix (Fin n)` and didn't handle other types and classes at all. It also lets us hoist more general ix dictionaries (e.g. `\n. Ix (Fin n)` instead of `Ix (Fin 10)`, `Ix (Fin 13)`, etc.), which should produce more cache hits, and makes it more reasonable to emit them to the top level.

With this change, we're formalizing two restrictions which were previously just unimplemented/crashy holes in the compiler:
  1. instances must be parametric over all data arguments. You can't define an instance like `Ix (Fin 10)`. It has to be `{n:Nat} (Ix (Fin n))` instead.
  2. type constructor parameters must be one of type/data/dict, so we can't have, e.g., functions as type constructor parameters.

This commit doesn't actually enforce these invariants - either the parametricity or the data/dict/type constraint on type parameters. We should do that as a follow-up.

### DIFF
--- a/src/lib/Core.hs
+++ b/src/lib/Core.hs
@@ -27,6 +27,7 @@ import Control.Monad.Identity
 import Control.Monad.Reader
 import Control.Monad.Writer.Strict hiding (Alt)
 import Control.Monad.State
+import qualified Control.Monad.State.Strict as SS
 import qualified Data.Map.Strict       as M
 
 import Name
@@ -269,6 +270,20 @@ instance (Monad m, ExtOutMap Env decls, OutFrag decls)
           return (ans, catOutFrags (toScope env') decls d, env')
   {-# INLINE refreshAbs #-}
 
+instance ( Monad m, ExtOutMap Env d1, ExtOutMap Env d2
+         , OutFrag d1, OutFrag d2, SubstB Name d1, HoistableB d1)
+         => EnvExtender (DoubleInplaceT Env d1 d2 m) where
+  refreshAbs ab cont = do
+    (ans, decls) <- UnsafeMakeDoubleInplaceT do
+      SS.StateT \s@(topScope, _) -> do
+        (ans, (_, decls)) <- refreshAbs ab \b e -> do
+          flip SS.runStateT (topScope, emptyOutFrag) $
+            unsafeRunDoubleInplaceT $ cont b e
+        return ((ans, decls), s)
+    unsafeEmitDoubleInplaceTHoisted decls
+    return ans
+  {-# INLINE refreshAbs #-}
+
 -- === Typeclasses for syntax ===
 
 -- TODO: unify this with `HasNames` by parameterizing by the thing you bind,
@@ -319,6 +334,14 @@ instance BindsEnv TopEnvFrag where
 
 instance BindsEnv EnvFrag where
   toEnvFrag frag = frag
+  {-# INLINE toEnvFrag #-}
+
+instance BindsEnv RoleBinder where
+  toEnvFrag (RoleBinder b ty _) = toEnvFrag (PiBinder b ty PlainArrow)
+  {-# INLINE toEnvFrag #-}
+
+instance BindsEnv RolePiBinder where
+  toEnvFrag (RolePiBinder b ty arr _) = toEnvFrag (PiBinder b ty arr)
   {-# INLINE toEnvFrag #-}
 
 instance BindsEnv (RecSubstFrag Binding) where
--- a/src/lib/Generalize.hs
+++ b/src/lib/Generalize.hs
@@ -1,0 +1,231 @@
+-- Copyright 2022 Google LLC
+--
+-- Use of this source code is governed by a BSD-style
+-- license that can be found in the LICENSE file or at
+-- https://developers.google.com/open-source/licenses/bsd
+
+module Generalize (generalizeArgs, generalizeIxDict, Generalized) where
+
+import Control.Monad
+import Control.Category ((>>>))
+
+import Syntax
+import Inference
+import QueryType
+import Name
+import Err
+import MTL1
+import LabeledItems
+
+type Generalized (e::E) (n::S) = (Abs (Nest Binder) e n, [Atom n])
+
+generalizeIxDict :: EnvReader m => Dict n -> m n (Generalized Dict n)
+generalizeIxDict dict = liftGeneralizerM do
+  dict' <- sinkM dict
+  dictTy <- getType dict'
+  dictTyGeneralized <- generalizeType dictTy
+  liftEnvReaderM $ generalizeDict dictTyGeneralized dict'
+{-# INLINE generalizeIxDict #-}
+
+generalizeArgs ::EnvReader m => Nest PiBinder n l -> [Atom n] -> m n (Generalized (ListE Atom) n)
+generalizeArgs reqTys allArgs =  liftGeneralizerM $ runSubstReaderT idSubst do
+  PairE (Abs reqTys' UnitE) (ListE allArgs') <- sinkM $ PairE (Abs reqTys UnitE) (ListE allArgs)
+  ListE <$> go reqTys' allArgs'
+  where
+    go ::Nest PiBinder i i' -> [Atom n] -> SubstReaderT AtomSubstVal GeneralizerM i n [Atom n]
+    go Empty [] = return []
+    go (Nest (PiBinder b ty _) bs) (arg:args) = do
+      ty' <- substM ty
+      arg' <- case ty' of
+        TyKind   -> liftSubstReaderT $ generalizeType arg
+        DictTy _ -> generalizeDict ty' arg
+        _ -> return arg
+      args'' <- extendSubst (b@>SubstVal arg') $ go bs args
+      return $ arg' : args''
+    go _ _ = error "zip error"
+{-# INLINE generalizeArgs #-}
+
+-- === generalizer monad plumbing ===
+
+data GeneralizationEmission n l = GeneralizationEmission (Binder n l) (Atom n)
+type GeneralizationEmissions = RNest GeneralizationEmission
+
+newtype GeneralizerM n a = GeneralizerM {
+  runGeneralizerM' :: DoubleInplaceT Env GeneralizationEmissions UnitB HardFailM n a }
+  deriving (Functor, Applicative, Monad, MonadFail, Fallible, ScopeReader, EnvReader, EnvExtender)
+
+liftGeneralizerM
+  :: EnvReader m
+  => (forall l. DExt n l => GeneralizerM l (e l))
+  -> m n (Generalized e n)
+liftGeneralizerM cont = do
+  env <- unsafeGetEnv
+  Distinct <- getDistinct
+  Abs emissions (DoubleInplaceTResult UnitB e) <- return $ runHardFail $
+    runDoubleInplaceT env $ runGeneralizerM' cont
+  let (bs, vals) = hoistGeneralizationVals (unRNest emissions)
+  return (Abs bs e, vals)
+  where
+    -- TODO: something not O(N^2)
+    hoistGeneralizationVals :: Nest GeneralizationEmission n l -> (Nest Binder n l, [Atom n])
+    hoistGeneralizationVals Empty = (Empty, [])
+    hoistGeneralizationVals (Nest (GeneralizationEmission b val) bs) = do
+      let (bs', vals) = hoistGeneralizationVals bs
+      case hoist b (ListE vals) of
+        HoistSuccess (ListE vals') -> (Nest b bs', val:vals')
+        HoistFailure _ -> error "should't happen"
+{-# INLINE liftGeneralizerM #-}
+
+-- XXX: the supplied type may be more general than the type of the atom!
+emitGeneralizationParameter :: Type n -> Atom n -> GeneralizerM n (AtomName n)
+emitGeneralizationParameter ty val = GeneralizerM do
+  Abs b v <- return $ newName noHint
+  let emission = Abs (RNest REmpty (GeneralizationEmission (b:>ty) val)) v
+  emitDoubleInplaceTHoisted emission >>= \case
+    Nothing -> error $ "expected atom to be hoistable " ++ pprint val
+    Just v' -> return v'
+
+-- === actual generalization traversal ===
+
+-- Given a type (an Atom of type `Type`), abstracts over all data components
+generalizeType :: Type n -> GeneralizerM n (Type n)
+generalizeType ty = traverseTyParams ty \paramRole paramReqTy param -> case paramRole of
+  TypeParam -> generalizeType param
+  DictParam -> generalizeDict paramReqTy param
+  DataParam -> do
+    paramTyGeneralized <- getType param >>= generalizeType
+    Var <$> emitGeneralizationParameter paramTyGeneralized param
+
+-- === role-aware type traversal ===
+
+-- This traverses the type parameters, with knowledge of their roles and
+-- expected types. It's used here for generalization, but it may also be useful
+-- for other operations on types, like newtype stripping.
+
+traverseTyParams
+  :: (EnvReader m, EnvExtender m)
+  => Atom n
+  -> (forall l . DExt n l => ParamRole -> Type l -> Atom l -> m l (Atom l))
+  -> m n (Atom n)
+traverseTyParams ty f = getDistinct >>= \Distinct -> case ty of
+  TypeCon sn def (DataDefParams ordinaryParams dictParams) -> do
+    Abs roleBinders UnitE <- getDataDefRoleBinders def
+    params' <- traverseRoleBinders f roleBinders (ordinaryParams ++ dictParams)
+    let (ordinaryParams', dictParams') = splitAt (length ordinaryParams) params'
+    return $ TypeCon sn def $ DataDefParams ordinaryParams' dictParams'
+  DictTy (DictType sn name params) -> do
+    Abs paramRoles UnitE <- getClassRoleBinders name
+    params' <- traverseRoleBinders f paramRoles params
+    return $ DictTy $ DictType sn name params'
+  TabPi (TabPiType (b:>(IxType iTy d)) resultTy) -> do
+    iTy' <- f TypeParam TyKind iTy
+    dictTy <- ignoreFallibleT1 $ DictTy <$> ixDictType iTy'
+    d'   <- f DictParam dictTy d
+    withFreshBinder (getNameHint b) (toBinding iTy') \b' -> do
+      resultTy' <- applySubst (b@>binderName b') resultTy >>= f TypeParam TyKind
+      return $ TabTy (b':>IxType iTy' d') resultTy'
+  RecordTy  elems -> RecordTy  <$> traverserseFieldRowElemTypes (f TypeParam TyKind) elems
+  VariantTy (Ext elems Nothing) -> do
+    elems' <- traverse (f TypeParam TyKind) elems
+    return $ VariantTy $ Ext elems' Nothing
+  TC tc -> TC <$> case tc of
+    BaseType b -> return $ BaseType b
+    ProdType tys -> ProdType <$> forM tys \t -> f TypeParam TyKind t
+    SumType  tys -> SumType  <$> forM tys \t -> f TypeParam TyKind t
+    Nat -> return Nat
+    Fin n -> Fin <$> f DataParam NatTy n
+    RefType _ _ -> error "not implemented" -- how should we handle the ParamRole for the heap parameter?
+    TypeKind         -> return TypeKind
+    EffectRowKind    -> return EffectRowKind
+    LabeledRowKindTC -> return LabeledRowKindTC
+    LabelType        -> return LabelType
+  _ -> error $ "Not implemented: " ++ pprint ty
+{-# INLINE traverseTyParams #-}
+
+traverseRoleBinders
+  :: forall m n n'. EnvReader m
+  => (forall l . DExt n l => ParamRole -> Type l -> Atom l -> m l (Atom l))
+  ->  Nest RoleBinder n n' -> [Atom n] -> m n [Atom n]
+traverseRoleBinders f allBinders allParams =
+  runSubstReaderT idSubst $ go allBinders allParams
+  where
+    go :: forall i i'. Nest RoleBinder i i' -> [Atom n]
+       -> SubstReaderT AtomSubstVal m i n [Atom n]
+    go Empty [] = return []
+    go (Nest (RoleBinder b ty role) bs) (param:params) = do
+      ty' <- substM ty
+      Distinct <- getDistinct
+      param' <- liftSubstReaderT $ f role ty' param
+      params'' <- extendSubst (b@>SubstVal param') $ go bs params
+      return $ param' : params''
+    go _ _ = error "zip error"
+{-# INLINE traverseRoleBinders #-}
+
+
+traverserseFieldRowElemTypes :: Monad m => (Type n -> m (Type n)) -> FieldRowElems n -> m (FieldRowElems n)
+traverserseFieldRowElemTypes f els = fieldRowElemsFromList <$> traverse checkElem elemList
+  where
+    elemList = fromFieldRowElems els
+    checkElem = \case
+      StaticFields items -> StaticFields <$> traverse f items
+      DynField _ _ -> error "shouldn't have dynamic fields post-simplification"
+      DynFields _  -> error "shouldn't have dynamic fields post-simplification"
+
+-- traverseLabeledRow
+--   :: Monad m
+--   => (Type n -> m (Type n))
+--   -> LabeledItems (Type n)
+--   -> m (LabeledItems (Type n))
+-- traverseLabeledRow f (items rest) = undefined
+-- {-# INLINE traverseLabeledRow #-}
+
+-- traverseLabeledRow (Ext items rest) = do
+--   mapM_ (|: TyKind) items
+--   forM_ rest \name -> do
+--     name' <- lookupSubstM name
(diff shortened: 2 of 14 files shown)


src/lib/Types/Core.hs
```diff
@@ -264,16 +268,19 @@ data ClassDef (n::S) where
   ClassDef
     :: SourceName
     -> [SourceName]              -- method source names
-    -> Nest Binder n1 n2         -- parameters
+    -> Nest RoleBinder n1 n2     -- parameters
```
apaszke, line 271: Why do typeclasses have role binders too? I thought it's just a datacon thing?

src/lib/Types/Core.hs
```diff
@@ -730,6 +737,19 @@ instance BindsOneAtomName PiBinder where
...
   binderType (_ :> IxType ty _) = ty
 
+instance BindsOneAtomName RoleBinder where
+  binderType (RoleBinder _ ty _) = ty
+
+instance BindsOneAtomName RolePiBinder where
+  binderType (RolePiBinder _ ty _ _) = ty
+
+toBinder :: BindsOneAtomName b => b n l -> Binder n l
+toBinder b = asNameBinder b :> binderType b
```
apaszke, line 747: Inline?

src/lib/Types/Core.hs
```diff
@@ -1663,6 +1683,58 @@ instance SubstE AtomSubstVal PiType
 deriving instance Show (PiType n)
 deriving via WrapE PiType n instance Generic (PiType n)
 
+instance GenericB RolePiBinder where
+  type RepB RolePiBinder = BinderP AtomNameC (PairE Type (LiftE (Arrow, ParamRole)))
+  fromB (RolePiBinder b ty arr role) = b :> PairE ty (LiftE (arr, role))
+  toB   (b :> PairE ty (LiftE (arr, role))) = RolePiBinder b ty arr role
```
apaszke, line 1689: Please always mark generic methods INLINE

src/lib/Types/Core.hs
```diff
@@ -1663,6 +1683,58 @@ instance SubstE AtomSubstVal PiType
...
+instance HoistableB  RolePiBinder
+instance SubstB Name RolePiBinder
+instance SubstB AtomSubstVal RolePiBinder
+instance AlphaEqB RolePiBinder
+instance AlphaHashableB RolePiBinder
+
+instance GenericB RoleBinder where
+  type RepB RoleBinder = BinderP AtomNameC (PairE Type (LiftE ParamRole))
+  fromB (RoleBinder b ty role) = b :> PairE ty (LiftE role)
+  toB   (b :> PairE ty (LiftE role)) = RoleBinder b ty role
```
apaszke, line 1715: Like above, please INLINE generic methods

src/lib/MTL1.hs
```diff
@@ -290,6 +290,13 @@ runFallibleT1 m =
     Left errs -> return $ Failure errs
 {-# INLINE runFallibleT1 #-}
 
+ignoreFallibleT1 :: Monad1 m => FallibleT1 m n a -> m n a
+ignoreFallibleT1 m =
```
apaszke, line 294: Isn't this just `liftM ignoreExcept $ runFallibleT1`? I don't think we need a helper for that

src/lib/Core.hs
```diff
@@ -269,6 +270,20 @@ instance (Monad m, ExtOutMap Env decls, OutFrag decls)
           return (ans, catOutFrags (toScope env') decls d, env')
   {-# INLINE refreshAbs #-}
 
+instance ( Monad m, ExtOutMap Env d1, ExtOutMap Env d2
+         , OutFrag d1, OutFrag d2, SubstB Name d1, HoistableB d1)
+         => EnvExtender (DoubleInplaceT Env d1 d2 m) where
```
apaszke, line 275: How come we didn't need this before?

src/lib/Generalize.hs
```diff
@@ -0,0 +1,231 @@
...
+    go _ _ = error "zip error"
+{-# INLINE generalizeArgs #-}
+
+-- === generalizer monad plumbing ===
+
+data GeneralizationEmission n l = GeneralizationEmission (Binder n l) (Atom n)
+type GeneralizationEmissions = RNest GeneralizationEmission
+
+newtype GeneralizerM n a = GeneralizerM {
+  runGeneralizerM' :: DoubleInplaceT Env GeneralizationEmissions UnitB HardFailM n a }
```
apaszke, line 54: Why use `DoubleInplaceT` when one of the emissions are `UnitB`? Isn't that equivalent to the regular `InplaceT`?

src/lib/Generalize.hs
```diff
@@ -0,0 +1,231 @@
...
+  -> m n (Generalized e n)
+liftGeneralizerM cont = do
+  env <- unsafeGetEnv
+  Distinct <- getDistinct
+  Abs emissions (DoubleInplaceTResult UnitB e) <- return $ runHardFail $
+    runDoubleInplaceT env $ runGeneralizerM' cont
+  let (bs, vals) = hoistGeneralizationVals (unRNest emissions)
+  return (Abs bs e, vals)
+  where
+    -- TODO: something not O(N^2)
```
apaszke, line 69: I've recently started using `-- OPTIMIZE: ...` to mark TODOs related to optimization. Can we use that convention?

src/lib/Generalize.hs
```diff
@@ -0,0 +1,231 @@
...
+  ClassDef _ _ bs _ _ <- lookupClassDef def
+  return $ Abs bs UnitE
+{-# INLINE getClassRoleBinders #-}
+
+-- === instances ===
+
+instance GenericB GeneralizationEmission where
+  type RepB GeneralizationEmission = BinderP AtomNameC (PairE Type Atom)
+  fromB (GeneralizationEmission (b:>ty) x) = b :> PairE ty x
+  toB   (b :> PairE ty x) = GeneralizationEmission (b:>ty) x
```
apaszke, line 216: Please mark both INLINE

src/lib/Generalize.hs
```diff
@@ -0,0 +1,231 @@
...
+--   forM_ rest \name -> do
+--     name' <- lookupSubstM name
+--     ty <- atomBindingType <$> lookupEnv name'
+--     checkAlphaEq LabeledRowKind ty
+
+getDataDefRoleBinders :: EnvReader m => DataDefName n -> m n (Abs (Nest RoleBinder) UnitE n)
+getDataDefRoleBinders def = do
+  DataDef _ (DataDefBinders bs dictBs) _ <- lookupDataDef def
+  let bs' = fmapNest (\(b:>ty) -> RoleBinder b ty DictParam) dictBs
+  return $ Abs (bs >>> bs') UnitE
```
apaszke, line 202: This will likely clash badly with Alexey's patch

src/lib/Generalize.hs
```diff
@@ -0,0 +1,231 @@
...
+--     checkAlphaEq LabeledRowKind ty
+
+
+-- checkLabeledRow :: Typer m => ExtLabeledItems (Type i) (AtomName i) -> m i o ()
+-- checkLabeledRow (Ext items rest) = do
+--   mapM_ (|: TyKind) items
+--   forM_ rest \name -> do
+--     name' <- lookupSubstM name
+--     ty <- atomBindingType <$> lookupEnv name'
+--     checkAlphaEq LabeledRowKind ty
```
apaszke, line 196: Why are those commented out? Should they be implemented? Or removed?

src/lib/Generalize.hs
```diff
@@ -0,0 +1,231 @@
...
+  VariantTy (Ext elems Nothing) -> do
+    elems' <- traverse (f TypeParam TyKind) elems
+    return $ VariantTy $ Ext elems' Nothing
+  TC tc -> TC <$> case tc of
+    BaseType b -> return $ BaseType b
+    ProdType tys -> ProdType <$> forM tys \t -> f TypeParam TyKind t
+    SumType  tys -> SumType  <$> forM tys \t -> f TypeParam TyKind t
+    Nat -> return Nat
+    Fin n -> Fin <$> f DataParam NatTy n
+    RefType _ _ -> error "not implemented" -- how should we handle the ParamRole for the heap parameter?
```
apaszke, line 137: answering the comment: like a regular type? just abstract over it fully

src/lib/Inference.hs
```diff
@@ -2629,6 +2654,53 @@ synthTopBlock block = do
   (liftExcept =<<) $ liftDictSynthTraverserM $ traverseGenericE block
 {-# SCC synthTopBlock #-}
 
+-- Given a simplified dict (an Atom of type `DictTy _` in the
+-- post-simplification IR), and a requested, more general, dict type, generalize
+-- the dict to match the more general type. This is only possible because we
+-- require that instances are polymorphic in data-role parameters.
+generalizeDict :: (EnvReader m) => Type n -> Dict n-> m n (Dict n)
```
apaszke, line 2661: Why does this live in Inference? Is it because you need unification? Perhaps we should factor unification into a separate file instead, because this is weird?

src/lib/Generalize.hs
```diff
@@ -0,0 +1,231 @@
...
+
+-- === actual generalization traversal ===
+
+-- Given a type (an Atom of type `Type`), abstracts over all data components
+generalizeType :: Type n -> GeneralizerM n (Type n)
+generalizeType ty = traverseTyParams ty \paramRole paramReqTy param -> case paramRole of
+  TypeParam -> generalizeType param
+  DictParam -> generalizeDict paramReqTy param
+  DataParam -> do
+    paramTyGeneralized <- getType param >>= generalizeType
```
apaszke, line 96: Why do you call `getType param` when you already have `paramReqTy`?

src/lib/Generalize.hs
```diff
@@ -0,0 +1,231 @@
...
+{-# INLINE liftGeneralizerM #-}
+
+-- XXX: the supplied type may be more general than the type of the atom!
+emitGeneralizationParameter :: Type n -> Atom n -> GeneralizerM n (AtomName n)
+emitGeneralizationParameter ty val = GeneralizerM do
+  Abs b v <- return $ newName noHint
+  let emission = Abs (RNest REmpty (GeneralizationEmission (b:>ty) val)) v
+  emitDoubleInplaceTHoisted emission >>= \case
+    Nothing -> error $ "expected atom to be hoistable " ++ pprint val
+    Just v' -> return v'
```
apaszke, line 86: This can't fail because the bottom emissions are UnitB, right?

src/lib/Generalize.hs
```diff
@@ -0,0 +1,231 @@
...
+  return (Abs bs e, vals)
+  where
+    -- TODO: something not O(N^2)
+    hoistGeneralizationVals :: Nest GeneralizationEmission n l -> (Nest Binder n l, [Atom n])
+    hoistGeneralizationVals Empty = (Empty, [])
+    hoistGeneralizationVals (Nest (GeneralizationEmission b val) bs) = do
+      let (bs', vals) = hoistGeneralizationVals bs
+      case hoist b (ListE vals) of
+        HoistSuccess (ListE vals') -> (Nest b bs', val:vals')
+        HoistFailure _ -> error "should't happen"
```
apaszke, line 76: Why can't this happen? It's not immediately obvious. Perhaps add a short comment too?

src/lib/Inference.hs
```diff
@@ -1531,6 +1532,24 @@ binderNestAsPiNest arr = \case
...
+inferRolePiBinderNest Empty = return Empty
+inferRolePiBinderNest (Nest (PiBinder b ty arr) bs) =
+  Nest (RolePiBinder b ty arr (typeToRole ty)) <$> inferRolePiBinderNest bs
+{-# INLINE inferRolePiBinderNest #-}
+
+typeToRole :: Type n -> ParamRole
+typeToRole ty = case ty of
+  TyKind -> TypeParam
+  DictTy _ -> DictParam
+  _ -> DataParam -- TODO(dougalm): check that it's actually "data" and throw a useful error if not
```
apaszke, line 1551: Since this is entirely type-driven, why do we have to keep those role annotations in binders? We could just resynthesize them whenever needed, right? It would save us the new `Role*` types and a bunch of casts between PiBinders and those

src/lib/Inference.hs
```diff
@@ -2629,6 +2654,53 @@ synthTopBlock block = do
...
+    SuperclassProj _ _    -> notSimplifiedDict
+    where notSimplifiedDict = error $ "Not a simplified dict: " ++ pprint dict
+
+generalizeInstanceArgs :: EmitsInf n => Nest RolePiBinder n l -> [Atom n] -> SolverM n [Atom n]
+generalizeInstanceArgs Empty [] = return []
+generalizeInstanceArgs (Nest (RolePiBinder b ty _ role) bs) (arg:args) = do
+  arg' <- case role of
+    TypeParam -> Var <$> freshInferenceName TyKind
+    DictParam -> generalizeDictAndUnify ty arg
+    DataParam -> Var <$> freshInferenceName ty
```
apaszke, line 2698: I am a little surprised by how uniform the treatment of type and data parameters ended up, because they should act quite differently here. It is definitely ok to have instances that are not completely polymorphic over the type parameters. Or is it perhaps an instance of what I mentioned before, which is that we could try to detect how polymorphic a derivation was by looking at which terms are applied as instance args? I'd like to understand this point a bit better, because it does make me uneasy.

src/lib/Generalize.hs
```diff
@@ -0,0 +1,231 @@
...
+  ListE <$> go reqTys' allArgs'
+  where
+    go ::Nest PiBinder i i' -> [Atom n] -> SubstReaderT AtomSubstVal GeneralizerM i n [Atom n]
+    go Empty [] = return []
+    go (Nest (PiBinder b ty _) bs) (arg:args) = do
+      ty' <- substM ty
+      arg' <- case ty' of
+        TyKind   -> liftSubstReaderT $ generalizeType arg
+        DictTy _ -> generalizeDict ty' arg
+        _ -> return arg
```
apaszke, line 42: For example, here you do resynthesize the roles based on the types, instead of using the annotated binders. So why do we even need the binder annotations?

src/lib/Types/Core.hs
dougalm: Thanks for catching these. I'm trying to build the habit but I'm not there yet.

src/lib/MTL1.hs
dougalm: Sure. I'll delete.

src/lib/Core.hs
dougalm: Before we only used `DoubleInplaceT` in `DoubleBuilder`, and we implemented the `EnvExtender` instance directly.

src/lib/Generalize.hs
dougalm: Actually no! DoubleInplaceT has two sets of emissions, but they behave differently. The `d1` emissions are emitted to the "top" (really, the level at which `runDoubleINplaceT` was called), and the emission can fail because it requires hoisting. Ideally, we'd have `InplaceT` and `HoistingInplaceT` and somehow `DoubleInplaceT` would be the composition of those two things. But I haven't figured out how to make that work.

src/lib/Generalize.hs
dougalm: nice. Will do

src/lib/Generalize.hs
dougalm: yup...

src/lib/Generalize.hs
dougalm: Removed. Thanks for catching that.

src/lib/Generalize.hs
dougalm: Did you mean "like regular data". I think that's right. Maybe we could treat the heap parameter as a newtype wrapper around `Unit` or something.

src/lib/Inference.hs
dougalm: I put it here because this is where the rest of dictionary synthesis happens. `generalizeDict` is sort of like a fast-path version of dict synthesis where you don't have do figure out the whole derivation tree from scratch. You just fill in some holes. You could argue that synthesis as a whole should live in a different file though.

src/lib/Generalize.hs
dougalm: No, it can fail. The hoisted emissions don't just hoist about the bottom emissions (which are UnitB, as you said. They also hoist above every local binder. I think that only includes pi binders in this case, since we're just traversing a type.

src/lib/Generalize.hs
dougalm: Great observation. I think that should work. (Initially I didn't even call `generalizeType`, which was an error.) I'll try it.

src/lib/Inference.hs
dougalm: I'm torn on this one. The original intention was to treat role inference separate from rule use. (Just as we like to do with type inference.) That way we can change the inference rules if we like, and we guarantee that we only assign a single role, and we can allow explicit user-annotated roles or whatever. But the implementation turned out to make the role-as-a-function-of-type assumption more explicitly than I'd expected. I'm still inclined to keep it as it is, mainly because it separates "what should the role be" from "how should this role behave".

src/lib/Types/Core.hs
dougalm: I expected it to be just a type constructor thing too, but it turned out that I needed them for classes and instances too. It makes sense for classes because they're a lot like a type constructor (and indeed `Ix n` is a type). Concretely we need them to handle the `DictTy` case of `traverseTyParams`. Instances need the roles for a difference reason, which is that the roles tell you how to generalize a particular dict derivation to a more general type.

src/lib/Generalize.hs
apaszke: Sure, I get that, but if the "top" emissions are `UnitB` then they always successfully hoist and don't give you any extra information at the top. That's why I thought it ends up being equivalent to `InplaceT`? If not, what am I missing?

src/lib/Inference.hs
apaszke: Interesting take, makes sense!

src/lib/Inference.hs
apaszke: Right, I guess a good test for this might be to have a dict that's passed in as an explicit argument somewhere. We do that for `Accum` after all. Even though the type of the dict would indicate that it should have a dict param role, it's an explicit argument, so we should not assume coherence.

So I buy the annotation argument I think. But perhaps the roles should be derived from the arrows _and_ types, not just types.

src/lib/Generalize.hs
apaszke: Right, so you basically do role inference for functions here? What about the point of a dict that's an explicit argument? Perhaps you should take the arrow into account?

src/lib/Generalize.hs
dougalm: Ah, I see the confusion. Here, we're using `UnitB` for the "bottom" emissions. The top emissions are `GeneralizationEmissions`.

src/lib/Inference.hs
```diff
@@ -1567,15 +1571,35 @@ binderNestAsPiNest arr = \case
...
+inferRolePiBinderNest :: Fallible m => Nest PiBinder n l -> m (Nest RolePiBinder n l)
+inferRolePiBinderNest bs = forEachNestItemM bs \(PiBinder b ty arr) -> do
+  role <- inferRole ty arr
+  return $ RolePiBinder b ty arr role
+{-# INLINE inferRolePiBinderNest #-}
+
+inferRole :: Fallible m => Type n -> Arrow -> m ParamRole
+inferRole ty arr = case (ty, arr) of
+  (TyKind  , _         ) -> return TypeParam
+  (DictTy _, ClassArrow) -> return DictParam
```
apaszke, line 1589: Actually isn't anything that uses a ClassArrow a DictParam? The treatment of DictParams is mainly based on the way they are generated (i.e. through synthesis), not about them having a dict type? I don't think the code as written is incorrect, but perhaps a little confusing due to a too strong precondition.

apaszke: Yeah, the more I think about it, the more I am convinced that actually the role inference might be better rephrased as:
```haskell
case arr of
  ClassArrow -> DictParam
  _ -> isData ty >>= \case
    True -> DataParam
    False -> TypeParam
```
WDYT?

apaszke (commented):
I still have to go through the rest of Inference and Generalize, but I had a thought about role inference.

src/lib/Generalize.hs
apaszke: Ok, good point. Can you please add that as a comment?

src/lib/Inference.hs
apaszke: Ok I think I understand why this code might be looking like this, and what's the source of potential issues. One option would be to leave the type unmodified, but this could lead to potential unification issues if you've generalized their data components. Ideally, we'd traverse the type here and only leave the inference vars in place of the data components, since actually we do care about the non-data backbone of this type to be equal (and any inequality should lead to errors!). So it's really an unsound shortcut that might work, but we shouldn't be doing this IMO. Those functions should really live in `Generalize.hs` and then you'd be able to share the type traversal between them.

apaszke (commented):
A few more points that I think we should correct. Can you also please open the issue listing remaining checks we're missing for this system to be sound?

apaszke (commented):
A few more points that I think we should correct. Can you also please open the issue listing remaining checks we're missing for this system to be sound?

src/lib/Inference.hs
```diff
@@ -1567,16 +1570,46 @@ binderNestAsPiNest arr = \case
...
+  _ -> do
+    zonk ty >>= \case
+      TyKind -> return TypeParam
+      ty' -> isData ty' >>= \case
+        True -> return DataParam
+        -- TODO(dougalm): the `False` branch should throw an error but that's
+        -- currently too conservative. e.g. `data RangeFrom q:Type i:q = ...`
+        -- fails because `q` isn't data. We should be able to fix it once we
+        -- have a `Data a` class (see issue #680).
+        False -> return DataParam
```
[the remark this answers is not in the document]
dougalm: I don't think we can raise the error because the example that fails, `data RangeFrom q:Type i:q = ...`, is in the prelude! The point is that we only ever actually instantiate it with `q` satisfying `Data q`. But we don't have a way to express that yet.

src/lib/Inference.hs
```diff
@@ -2687,6 +2736,53 @@ synthTopBlock block = do
...
+  dict' <- generalizeDictRec dict
+  dictTy <- getType dict'
+  unify ty dictTy
+  zonk dict'
+
+generalizeDictRec :: EmitsInf n => Dict n -> SolverM n (Dict n)
+generalizeDictRec dict = do
+  -- TODO: we should be able to avoid the normalization here . We only need it
+  -- because we sometimes end up with superclass projections. But they shouldn't
+  -- really be allowed to occur in the post-simplification IR.
```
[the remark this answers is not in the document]
dougalm: No deeply. I'm hoping we can fix it by making the different IRs more precisely typed.

src/lib/Inference.hs
dougalm: Yes, I agree that would be cleaner and more obviously sound. I'll do that.

src/lib/Generalize.hs
dougalm: In writing the comment, I just realized that this only works if the binders within the type are only used in DataParam roles! This was always the intention (e.g. we assume that sort of thing in creating dests for dependent pairs in Imp). But we should make it more official. I'll add it to the issue about listing the remaining checks needed.

src/lib/Generalize.hs
dougalm: Something strange seems to be happening with github. Right now, I see this line as:
```
DictTy _ -> generalizeDict ty' arg
```
But it should actually be this:
```
DictTy _ | arr == ClassArrow -> generalizeDict ty' arg
```
I changed it in efd06e9f5ef15a8beb3cce9233ff98d66564d8cd, which is part of this PR and is now merged into main. But for some reason it's not showing up.

The reason I don't want to use the helper `inferRole` from `Inference` is that we need a less strict rule here. It's ok for arguments to functions to be non-data things. We just specialize on the actual value in that case.

PR #1142, opened 2022-11-03
Move Ix hoisting to simplification pass

dougalm:


src/lib/Simplify.hs
```diff
@@ -481,6 +517,16 @@ simplifyAtom atom = confuseGHC >>= \_ -> case atom of
   DepPairRef _ _ _ -> error "Should only occur in Imp lowering"
   ProjectElt idxs v -> getProjection (toList idxs) <$> simplifyVar v
 
+-- We use a different name because the implementations might need to diverge.
```
apaszke, line 520: This seems like a bad idea to me, especially that you've already missed a few points where types might appear (e.g. data con params), so it's already misleading. So perhaps just don't do this until we really need to, and then we'll do it properly?

apaszke (approved):
Mostly LGTM, but I left a few comments and suggestions.

src/lib/Builder.hs
```diff
@@ -261,13 +262,24 @@ class (EnvReader m, MonadFail1 m)
   -- `emitBinding` is expressible in terms of `emitEnv` but it can be
   -- implemented more efficiently by avoiding a double substitution
   -- XXX: emitBinding/emitEnv don't extend the synthesis candidates
+  -- TODO: do we actually need `emitBinding`? Top emissions probably aren't hot.
```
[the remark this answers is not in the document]
dougalm: Just that `emitBinding` is just an optimized special case of `emitEnv`. But the optimization is probably unnecessary because we only use it for top-level emissions, which are much less common that local emissions.

src/lib/Simplify.hs
dougalm: sgtm

src/lib/Simplify.hs
```diff
@@ -718,13 +764,19 @@ projectDictMethod d i = do
       case i of
         0 -> return method
         _ -> error "ExplicitDict only supports single-method classes"
+    DictCon (ExplicitMethods methodImplName params) -> do
+      SpecializedDictBinding (SpecializedDictDef _ fs) <- lookupEnv methodImplName
+      let f = fs !! i
+      TopFunBound _ (SpecializedTopFun (IxMethodSpecialization _ (Abs bs dict'))) <- lookupAtomName f
```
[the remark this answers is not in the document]
dougalm: Yeah, this wasn't organized very nicely. A dict is sufficient, but it's definitely not necessary. My latest commit fixes improves things, I think: 69d7d923423f0d03acf8103bf6fb542f66ad6d3d

src/lib/Simplify.hs
```diff
@@ -440,15 +458,33 @@ simplifyAtom atom = confuseGHC >>= \_ -> case atom of
...
+  TypeCon sn dataDefName (DataDefParams arrParams) -> do
+    dataDefName' <- substM dataDefName
+    let (arrs, params) = unzip arrParams
+    params' <- mapM simplifyAtom params
+    return $ TypeCon sn dataDefName' (DataDefParams $ zip arrs params')
+  DictCon d -> do
+    d' <- substM d
+    dTy <- getType d'
+    DictCon <$> case dTy of
+      DictTy (DictType "Ix" _ _) -> emitIxDictSpecialization d'
```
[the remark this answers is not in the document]
dougalm: Agreed

src/lib/Simplify.hs
```diff
@@ -389,6 +389,24 @@ getSpecializedFunction s = do
...
+emitIxDictSpecialization ixDict = do
+  (Abs bs (PairE (DictTy dictTy@(DictType sn _ _)) ixDict'), params) <- generalizeIxDict (DictCon ixDict)
+  ab <- liftTopBuilderHoisted do
+    methodImplNames <- forM [minBound..maxBound] \methodName -> do
+      let s = IxMethodSpecialization methodName (sink (Abs bs ixDict'))
+      querySpecializationCache s >>= \case
+        Just name -> return name
+        _ -> emitSpecialization s
+    emitBinding (getNameHint (sn ++ "_sd")) $ sink $ SpecializedDictBinding $
+      SpecializedDictDef (sink (Abs bs dictTy)) methodImplNames
```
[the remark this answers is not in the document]
dougalm: Great catch! This is exactly the bug I mentioned in chat on Friday. Fixed by 69d7d923423f0d03acf8103bf6fb542f66ad6d3d

src/lib/Types/Core.hs
```diff
@@ -564,16 +565,27 @@ instance SubstE AtomSubstVal EffectOpType
...
 deriving via WrapE EffectOpType n instance Generic (EffectOpType n)
 
+type AbsDict = Abs (Nest Binder) Dict
+
 data SpecializedDictDef n =
-  -- dict type abstracted over local params, method names
-  SpecializedDictDef (Abs (Nest Binder) DictType n) [AtomName n]
-  deriving (Show, Generic)
+   SpecializedDict
+     -- Dict is abstracted over local "data" params. The methods are
```
apaszke, line 572: This comment ends in the middle of a sentence?

apaszke: I also don't like calling them "local" data params. There's nothing local here, this is at the top level. It's just abstracted over some data params we chose at the abstraction site, but they didn't have to be local at all (e.g. the 10 in `Fin 10`)

src/lib/Types/Core.hs
```diff
@@ -2140,6 +2148,12 @@ instance ExtOutMap Env TopEnvFrag where
 
       lookupModulePure v = case lookupEnvPure (Env newTopEnv mempty) v of ModuleBinding m -> m
 
+addMethods :: (SpecDictName n, [NaryLamExpr n]) -> Env n -> Env n
+addMethods (dName, methods) e = do
+  let SpecializedDictBinding (SpecializedDict dAbs _) = lookupEnvPure e dName
```
apaszke, line 2153: Might be a good idea to pattern match on `Nothing` to make sure we don't add them multiple times.

src/lib/TopLevel.hs
```diff
@@ -604,18 +606,42 @@ evalBlock typed = do
...
+evalSpecializations vs sdVs = do
+  forM_ vs \v -> do
+    lookupAtomName v >>= \case
+      TopFunBound _ (SpecializedTopFun (AppSpecialization _ _)) ->
+        queryImpCache v >>= \case
+          Nothing -> compileTopLevelFun v
+          Just _ -> return ()
+      _ -> return ()
+  forM_ sdVs \d -> do
+    SpecializedDictBinding (SpecializedDict absDict@(Abs bs _) _) <- lookupEnv d
```
apaszke, line 619: Might be a good idea to assert `Nothing` here as well, just to make sure we're not wasting work.

src/lib/Simplify.hs
```diff
@@ -764,12 +764,10 @@ projectDictMethod d i = do
...
-    DictCon (ExplicitMethods methodImplName params) -> do
-      SpecializedDictBinding (SpecializedDictDef _ fs) <- lookupEnv methodImplName
-      let f = fs !! i
-      TopFunBound _ (SpecializedTopFun (IxMethodSpecialization _ (Abs bs dict'))) <- lookupAtomName f
-      dict'' <- applySubst (bs @@> map SubstVal params) dict'
-      projectDictMethod dict'' i
+    DictCon (ExplicitMethods sd params) -> do
+      SpecializedDictBinding (SpecializedDict (Abs bs dict) _) <- lookupEnv sd
+      dict' <- applySubst (bs @@> map SubstVal params) dict
+      projectDictMethod dict' i
```
apaszke, line 760: I am a bit surprised here. Isn't the second component of `SpecializedDict` supposed to hold the names of specialized methods? Why can't we just look them up and use them here?

src/lib/QueryType.hs
```diff
@@ -471,9 +458,8 @@ dictExprType e = case e of
     liftM DictTy $ ixDictType $ TC $ Fin n'
   ExplicitMethods v params -> do
     params' <- mapM substM params
-    SpecializedDictBinding (SpecializedDictDef (Abs bs dictTy) _) <- lookupEnv =<< substM v
-    dictTy' <- applySubst (bs @@> map SubstVal params') dictTy
-    return $ DictTy dictTy'
+    SpecializedDictBinding (SpecializedDict (Abs bs dict) _) <- lookupEnv =<< substM v
+    getType =<< applySubst (bs @@> map SubstVal params') dict
```
apaszke, line 462: It would likely be better to use `getTypeSubst` since that avoids doing the full substitution of the dict.

src/lib/Types/Core.hs
dougalm: Agreed. Are we happy with calling them just "data params" then?

dougalm: Or "runtime params"?

src/lib/Simplify.hs
dougalm: Not anymore. The second component is `Maybe [NaryLamExpr n]`. During simplification, we can't be sure that the field will be non-Nothing. Probably deserves a comment.

dougalm:
@apaszke , I made your suggested changes in #1145

PR #1143, opened 2022-11-04
Add occurrence analysis.

axch:
This is not used yet; but now we should be able to write an inliner (which will also do producer-consumer loop fusion) that uses the information occurrence analysis computes.


src/lib/Name.hs
```diff
@@ -3326,6 +3332,78 @@ toListNameMap :: NameMap c a n -> [(Name c n, a)]
...
+mapNameMap :: (a -> b) -> NameMap c a n -> (NameMap c b n)
+mapNameMap f (UnsafeNameMap raw) = UnsafeNameMap $ fmap f raw
+{-# INLINE mapNameMap #-}
+
+newtype NameMapE (c::C) (e:: E) (n::S) = NameMapE (NameMap c (e n) n)
+  deriving (Eq, Semigroup, Monoid)
+
+hoistFilterNameMapE :: (BindsNames b, HoistableE e, ShowE e)
+                    => b n l -> NameMapE c e l -> HoistExcept (NameMapE c e n)
+hoistFilterNameMapE b (NameMapE nmap) =
```
apaszke, line 3354: Well the point of `hoistFilter` was that it never fails, but rather filters out the names that fail to hoist. So please either do that, or just call it `host` without the "filter" part

src/lib/Name.hs
```diff
@@ -3326,6 +3332,78 @@ toListNameMap :: NameMap c a n -> [(Name c n, a)]
...
+  => b n l -> (Abs b e n -> e n) -> NameMapE c e l -> NameMapE c e n
+hoistFilterNameMapEWith b f (NameMapE nmap) =
+  NameMapE $ mapNameMap doit $ hoistFilterNameMap b nmap where
+  doit v = f $ Abs b v
+{-# INLINE hoistFilterNameMapEWith #-}
+
+hoistFilterNameMapEWithM
+  :: (Applicative f, BindsNames b, HoistableE e, ShowE e)
+  => b n l -> (Abs b e n -> f (e n)) -> NameMapE c e l -> f (NameMapE c e n)
+hoistFilterNameMapEWithM b f (NameMapE nmap) =
```
apaszke, line 3368: What does this method do? It's a little hard to follow.

src/lib/Occurrence.hs
```diff
@@ -84,6 +87,14 @@ class MaxPlus a where
   max :: a -> a -> a
   plus :: a -> a -> a
 
+instance (MaxPlus a) => MaxPlus (M.IntMap a) where
```
apaszke, line 90: Well `IntMap` is a bit of an implementation detail of `NameMap`. Perhaps implement this for `NameMap` (since we already have `unionWith`) and then derive for `NameMapE`?

src/lib/Occurrence.hs
```diff
@@ -146,24 +160,40 @@ data Access (n::S) where
...
   -- An index of sum type
   | Inject Int (IxExpr n)
-  -- Unknown function of the given names
-  | Unknown [Name AtomNameC n]
-  -- All indices are accessed.  This constructor only appears after
-  -- `iterate`, below.
+  -- Unknown deterministic function of the given names
+  | Deterministic [Name AtomNameC n]
+  -- All indices are accessed, or we are conservatively assuming they are
+  -- because we don't know which are accessed.
```
apaszke, line 190: Is this form really sufficient? I would expect that we might want to distinguish between "all indices" coming form `for` loops and "any function" since "all indices" implies that every index is used once, while "any function" can do anything, including accessing an index twice?

src/lib/Occurrence.hs
```diff
@@ -637,14 +671,46 @@ instance ApproxConst CollapsedUse where
...
   approxConst (Inj Nothing uses) = approxConst uses
   approxConst (Prod _ use) = approxConst use
 
 instance ApproxConst v => ApproxConst (M.IntMap v) where
   approxConst uses = foldl' max zero $ map approxConst $ M.elems uses
 
+-- === Usage info including static usage ===
+
+-- The `Int` is the count of static occurrences
+data AccessInfo n = AccessInfo Int (Access n)
```
apaszke, line 685: What are "static occurrences" and how do they differ from `Access`?

src/lib/Occurrence.hs
```diff
@@ -637,14 +671,46 @@ instance ApproxConst CollapsedUse where
...
+  deriving (Show)
+
+instance MaxPlus (AccessInfo n) where
+  zero = AccessInfo 0 zero
+  -- Note that, since `max` corresponds to `case`, the static count is
+  -- still added
```
apaszke, line 691: Why? (It might be surprising just because I don't know what "static count" is)

src/lib/Name.hs
```diff
@@ -3326,6 +3332,78 @@ toListNameMap :: NameMap c a n -> [(Name c n, a)]
...
+{-# INLINE unionWithNameMapE #-}
+
+traverseNameMapE :: (Applicative f) => (e1 n -> f (e2 n))
+                 -> NameMapE c e1 n -> f (NameMapE c e2 n)
+traverseNameMapE f (NameMapE nmap) = NameMapE <$> traverseNameMap f nmap
+{-# INLINE traverseNameMapE #-}
+
+mapNameMapE :: (e1 n -> e2 n)
+            -> NameMapE c e1 n -> NameMapE c e2 n
+mapNameMapE f = runIdentity . traverseNameMapE (pure . f)
```
apaszke, line 3401: Perhaps use `mapNameMap` here?

tests/unit/OccAnalysisSpec.hs
```diff
@@ -0,0 +1,336 @@
...
+  go :: Nest Decl n l -> LetAnn
+  go (Nest (Let _ (DeclBinding ann _ (Hof (RunIO _)))) _) = ann
+  go (Nest _ rest) = go rest
+  go Empty = error "RunIO not found"
+
+analyze :: EvalConfig -> TopStateEx -> [Text] -> IO LetAnn
+analyze cfg env code = fst <$> runTopperM cfg env do
+  [block] <- sourceTextToBlocks $ unlines code
+  block' <- analyzeOccurrences block
+  return $ findRunIOAnnotation block';
```
apaszke, line 63: Where is that RunIO in the blocks below?

apaszke (commented):
I didn't review the implementation, because I don't fully understand the new representation. I'll get to it once you reply to the comments.

src/lib/Name.hs
axch: Done.

src/lib/Name.hs
axch: This isn't needed anymore, so I'll just delete it.

src/lib/Occurrence.hs
axch: Done.

src/lib/Occurrence.hs
axch: Yes, it is sufficient.  Explanation added.

src/lib/Occurrence.hs
axch: Occurrences in the code, relevant for code size.  Brief comment added.

src/lib/Name.hs
axch: Done.

src/lib/OccAnalysis.hs
```diff
@@ -0,0 +1,559 @@
...
+  ElimSuccess :: Abs (Nest Decl) Atom n -> ElimResult n
+  ElimFailure :: Decl n l -> UsageInfo -> Abs (Nest Decl) Atom l -> ElimResult n
+
+occNest :: Access n -> Nest Decl n l -> Atom l
+        -> OCCM n (Abs (Nest Decl) Atom n)
+occNest a decls ans = case decls of
+  Empty -> Abs Empty <$> occ a ans
+  Nest d@(Let _ binding) ds -> do
+    isPureDecl <- isPure binding
+    -- Note that we only ever analyze the abs below under this refreshAbs,
```
apaszke, line 275: What is "the abs" here?

src/lib/OccAnalysis.hs
```diff
@@ -0,0 +1,559 @@
...
+        -- occurrence analysis, and each binding is considered for
+        -- inlining separately.
+        DeclBinding _ ty expr <- occ accessOnce binding'
+        let binding'' = DeclBinding (OccInfo usage) ty expr
+        return $ Abs (Nest (Let b' binding'') ds'') ans''
+
+instance HasOCC DeclBinding where
+  occ a (DeclBinding ann ty expr) = do
+    expr' <- occ a expr
+    -- TODO(optimization) Is free variable counting enough for types?
```
apaszke, line 309: How is that related to optimization? Seems more like a correctness concern? And I think the answer is as above.

src/lib/OccAnalysis.hs
```diff
@@ -0,0 +1,559 @@
...
+    (TabApp array (NE.toList -> ixs)) -> do
+      (a', ixs') <- go a ixs
+      array' <- occ a' array
+      return $ TabApp array' $ NE.fromList ixs'
+    (Case scrut alts ty effs) -> do
+      scrut' <- occ accessOnce scrut
+      scrutIx <- summary scrut
+      (alts', innerFVs) <- unzip <$> mapM (isolated . occAlt a scrutIx) alts
+      modify (<> foldl' Occ.max zero innerFVs)
+      -- TODO(optimization) Is free variable counting enough for types?
```
apaszke, line 324: Again, not an optimization

src/lib/OccAnalysis.hs
```diff
@@ -0,0 +1,559 @@
...
+instance HasOCC Expr where
+  occ a expr = case expr of
+    (TabApp array (NE.toList -> ixs)) -> do
+      (a', ixs') <- go a ixs
+      array' <- occ a' array
+      return $ TabApp array' $ NE.fromList ixs'
+    (Case scrut alts ty effs) -> do
+      scrut' <- occ accessOnce scrut
+      scrutIx <- summary scrut
+      (alts', innerFVs) <- unzip <$> mapM (isolated . occAlt a scrutIx) alts
```
apaszke, line 322: I am a little surprised that you pass in the same IxExpr for the scrutinee to every case, and that ends up being bound as the IxExpr for the case binder. The case binder is definitely not that same as the scrutinee, since it has the top constructor eliminated.

src/lib/OccAnalysis.hs
```diff
@@ -0,0 +1,559 @@
...
+    return expr
+  -- I'm pretty sure the others are all strict, and not usefully analyzable
+  -- for what they do to the incoming access pattern.
+  occ _ (ComposeE op) = ComposeE <$> traverse (occ accessOnce) op
+  {-# INLINE occ #-}
+
+-- === The generic instances ===
+
+instance HasOCC e => HasOCC (ComposeE PrimCon e) where
+  -- These too
```
apaszke, line 475: ?? Perhaps remove those comments since they just increase the code size without adding anything?

src/lib/OccAnalysis.hs
```diff
@@ -0,0 +1,559 @@
...
+instance HasOCC TabLamExpr where
+  occ _ view = inlinedLater view
+  {-# INLINE occ #-}
+
+instance HasOCC (ComposeE PrimOp Atom) where
+  occ _ expr@(ComposeE (PrimEffect _ (MExtend (BaseMonoid _ combine) val))) = do
+    valIx <- summary val
+    -- Treat the combining function as inlined here and called once
+    void $ oneShot2 accessOnce (Deterministic []) valIx combine
+    -- Do not change it
```
apaszke, line 465: Is that correct? What if the combining function has some read code that was eliminated, and that was reflected in the accesses?

src/lib/OccAnalysis.hs
```diff
@@ -0,0 +1,559 @@
...
+      bd' <- oneShot2 a (Deterministic []) iniIx bd
+      return $ RunReader ini' bd'
+    RunWriter Nothing (BaseMonoid empty combine) bd -> do
+      -- We will process the combining function when we meet it in MExtend ops
+      -- (but we won't attempt to eliminate dead code in it).
+      empty' <- occ accessOnce empty
+      -- The body is exactly one-shot, and shouldn't index with the reference
+      -- parameter.  (But if it does, Deterministic [] seems reasonable, because
+      -- RunWriter syntehsizes the reference, presumably independently of
+      -- anything else.)  Same for the heap parameter.
```
apaszke, line 382: Could we insert `undefined` in place of the `Deterministic`? While I would be surprised if anybody ever managed to index with a reference, I feel a little uneasy just dumping garbage values in the map.

src/lib/OccAnalysis.hs
```diff
@@ -0,0 +1,559 @@
...
+  _ -> error "Not a lambda expression"
+
+instance HasOCC Atom where
+  occ a atom = case atom of
+    Lam _ -> error "Expecting all Lam Atoms to be covered by their contexts"
+    ACase _ _ _ -> error "Unexpected ACase outside of Simplify"
+    _ -> generic
+    where
+      generic = toE <$> (occ a $ fromE atom)
+  {-# INLINE occ #-}
```
apaszke, line 454: I don't think we should be inlining this one, since the `generic` branch will expand to a huge program. Also, it's probably self-recursive, so no inlining would take place anyway.

src/lib/OccAnalysis.hs
```diff
@@ -0,0 +1,559 @@
...
+instance (BindsEnv b, SubstB Name b, HoistableB b, SubstE Name e, HasOCC e)
+  => HasOCC (Abs b e) where
+  occ access a = do
+    -- The way this and hoistState are written, the pass will crash if any of
+    -- the AccessInfos reference this binder.  We should avoid triggering this
+    -- generic instance in any situation where that can happen.
+    a'@(Abs b' _) <- refreshAbs a \b e -> Abs b <$> occ (sink access) e
+    -- Should this be countFreeVarsAsOccurrences or should we recur
+    -- occurrence analysis?
```
apaszke, line 539: ??

src/lib/OccAnalysis.hs
```diff
@@ -0,0 +1,559 @@
...
+  let freeIterationNames = nub $ concatMap freeAtomVarsList freeIxs
+  return $ Deterministic $ freeIterationNames
+
+-- === The actual occurrence analysis ===
+
+class HasOCC (e::E) where
+  occ :: Access n -> e n -> OCCM n (e n)
+  default occ :: (GenericE e, HasOCC (RepE e)) => Access n -> e n -> OCCM n (e n)
+  occ a e = confuseGHC >>= \_ -> toE <$> occ a (fromE e)
+  {-# INLINE occ #-}
```
apaszke, line 233: Is this INLINE based on Core dumps, or just some expectation? I don't think it's a good idea

tests/unit/OccAnalysisSpec.hs
```diff
@@ -0,0 +1,339 @@
...
+        ]
+      ann `shouldBe` OccInfo (UsageInfo 1 (1,Unbounded))
+    it "detects non-nested single-uses cases despite index arithmetic" do
+      ann <- analyze cfg env
+        [ ":p"
+        , "  xs : (Fin 4) => Float = unreachable ()"
+        , "  xs.(1@_)"
+        ]
+      -- Arguably, should be able to prove that zero levels of exposed indexing
+      -- (not one) suffice for inlining xs to be safe here, but doesn't prove it
```
apaszke, line 147: How come "zero levels"? xs is indexed, right?

tests/unit/OccAnalysisSpec.hs
```diff
@@ -0,0 +1,339 @@
...
+        [ ":p"
+        , "  xs : (Fin 10) => Float = unreachable ()"
+        -- with_state prevents the access from appearing dead
+        , "  with_state 0 \\ref."
+        , "    while \\_."
+        , "      for i."
+        , "        ref := (get ref) + xs.i"
+        , "      False"
+        ]
+      -- TODO Why is this coming up as 0 indexing depth?  Should be 1.
```
apaszke, line 313: Something to fix before merging?

tests/unit/OccAnalysisSpec.hs
```diff
@@ -0,0 +1,339 @@
...
+      ann `shouldBe` OccInfo (UsageInfo 1 (0,Unbounded))
+    it "understands index-defining bindings" do
+      ann <- analyze cfg env
+        [ ":p"
+        , "  xs : (Fin 10 | Fin 3) => Float = unreachable ()"
+        , "  for i."
+        , "    j = Left i"
+        , "    xs.j"
+        ]
+      ann `shouldBe` OccInfo (UsageInfo 1 (1,Bounded 1))
```
apaszke, line 323: This is probably unsurprising, since the `j` binding is an atom binding and will get inlined in simplification?

src/lib/OccAnalysis.hs
```diff
@@ -0,0 +1,559 @@
...
+-- Change all the AccessInfos to assume their objects are _statically_ used many
+-- times, as for expressions that will be inlined unpredictably.
+-- TODO(precision): If this is for a `view`, we could use the occurrence
+-- information of the view itself to predict how much of an amplification
+-- inlining it would cause.
+useManyTimesStatic :: FV n -> FV n
+useManyTimesStatic (FV fvs) = FV $ mapNameMapE update fvs where
+  -- TODO(code cleanliness): Do we want to change the static counter to a Count
+  -- so the Unbounded constructor is available here?
+  update (AccessInfo _ d) = AccessInfo 1000 d
```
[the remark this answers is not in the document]
axch: Done.

src/lib/OccAnalysis.hs
```diff
@@ -0,0 +1,559 @@
...
+-- Run the given action with its own FV state, and return the FVs it
+-- accumulates for post-processing.
+isolated :: OCCM n a -> OCCM n (a, FV n)
+isolated action = do
+  r <- ask
+  lift11 $ lift11 $ runStateT1 (runReaderT1 r action) mempty
+
+-- Run the given action with its own FV state, and merge it afterward.
+-- This way the action can munge its state itself based on local
+-- information and know it's not affecting the surrounding state.
```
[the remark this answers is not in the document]
axch: You're right, this isn't needed any more.  Deleted.

src/lib/OccAnalysis.hs
```diff
@@ -0,0 +1,559 @@
...
+      TabRef _ -> invalid "TabRef"
+      ConRef _ -> invalid "ConRef"
+      ExplicitDict _ _ -> invalid "ExplicitDict"
+      DictHole _ _ -> invalid "DictHole"
+
+unknown :: HoistableE e => e n -> OCCM n (IxExpr n)
+unknown obj = do
+  freeIxs <- mapM ixExpr $ freeAtomVarsList obj
+  let freeIterationNames = nub $ concatMap freeAtomVarsList freeIxs
+  return $ Deterministic $ freeIterationNames
```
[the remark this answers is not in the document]
axch: For RWS I think so, because the reference will appear in the freeAtomVarsList if it is relevant, and I think that means the right thing will happen.  If the effect is IO, then I guess not.

However, this does mean that I have to detect wither and of the `freeIxs` are already `IxAll` and promote to `IxAll` if so.  Fixed that.

src/lib/OccAnalysis.hs
```diff
@@ -0,0 +1,559 @@
...
+      -- The free vars accumulated in the state of OCCM should
+      -- correspond to the free vars of the Abs of the block answer,
+      -- by the decls traversed so far. occNest takes care to uphold
+      -- this invariant, but we temporarily reset the state to an
+      -- empty map, just so that names from the surrounding block
+      -- don't end up influencing elimination decisions here. Note
+      -- that we restore the state (and accumulate free vars of the
+      -- analyzed block into it) right after occNest.
+      Abs decls' ans' <- locally $ occNest a decls ans
+      ty' <- occ accessOnce ty  -- TODO Is free variable counting enough here?
```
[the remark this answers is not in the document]
axch: Say more?  I wasn't sure the right treatment of occurrences in type annotations.  Can a computed array appear in a type?  Do we need to take care to not inline it if it does?  Or is that a different kind of "think twice about inlining this" than occurrences outside of types?

src/lib/OccAnalysis.hs
```diff
@@ -0,0 +1,559 @@
...
+  occ a (Block ann decls ans) = case (ann, decls) of
+    (NoBlockAnn      , Empty) -> Block NoBlockAnn Empty <$> occ a ans
+    (NoBlockAnn      , _    ) -> error "should be unreachable"
+    (BlockAnn ty effs, _    ) -> do
+      -- The free vars accumulated in the state of OCCM should
+      -- correspond to the free vars of the Abs of the block answer,
+      -- by the decls traversed so far. occNest takes care to uphold
+      -- this invariant, but we temporarily reset the state to an
+      -- empty map, just so that names from the surrounding block
+      -- don't end up influencing elimination decisions here. Note
```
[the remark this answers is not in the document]
axch: Copy-pasta from DCE.  The comment is stale now so I got rid of it.

src/lib/OccAnalysis.hs
```diff
@@ -0,0 +1,559 @@
...
+      -- that we restore the state (and accumulate free vars of the
+      -- analyzed block into it) right after occNest.
+      Abs decls' ans' <- locally $ occNest a decls ans
+      ty' <- occ accessOnce ty  -- TODO Is free variable counting enough here?
+      countFreeVarsAsOccurrences effs
+      return $ Block (BlockAnn ty' effs) decls' ans'
+
+-- TODO(optimization) Could reuse the free variable caching from dce here too.
```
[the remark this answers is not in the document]
axch: Agreed; added a comment so the idea is not lost.  Or do you think that should be done now (either before merging or quickly after)?

src/lib/OccAnalysis.hs
axch: I actually don't think this comment matters either.

src/lib/OccAnalysis.hs
axch: I tagged it "optimization" because presumably traversing something to count its free variables is easier than running occurrence analysis on it.  May also affect the precision; but I want to learn more about the right treatment of types.

src/lib/OccAnalysis.hs
axch: It doesn't end up being bound as the IxExpr for the case binder: the case binder is an `unknown` applied to the scrutinee.  This is a conservative approximation, due to not having constructors in `IxExpr` for eliminating the top constructor; and I guess that if I could statically eliminate it then the whole case would have been removed by case-of-known-constructor.

Added a comment to this effect.

src/lib/OccAnalysis.hs
axch: Done.

src/lib/OccAnalysis.hs
axch: Hm.  That sounds like a potential bug.  Noted in a comment.  Let me know if you think we should fix it now.

src/lib/OccAnalysis.hs
axch: This thunk is actually forced, so no.  Wrote out a more complete comment defending `Deterministic []` as sound here, including an example of why `undefined` should not be expected to work.

src/lib/OccAnalysis.hs
axch: Removed.

src/lib/OccAnalysis.hs
axch: I no longer remember what I meant; deleted.

src/lib/OccAnalysis.hs
axch: Flushed.

tests/unit/OccAnalysisSpec.hs
axch: Yes, but inlining `xs` is work-preserving here even if we can't cancel against the indexing operation, so the most precise answer is zero levels.

tests/unit/OccAnalysisSpec.hs
axch: On second thought, this is not a bug (though it is something of an oddity).  Explanation added; I could also fix it if desired.

tests/unit/OccAnalysisSpec.hs
axch: Made the test an expression instead (though that is always "unknown" at the moment).

src/lib/OccAnalysis.hs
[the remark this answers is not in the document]
apaszke: > Can a computed array appear in a type?

Yes. It happens in the sets-as-indices example to name one.

> Do we need to take care to not inline it if it does?

You should preserve the binding I think, but it should not affect the other inlining decisions. We need the binding to be able to decide type equality, but if it's only used in types, then ideally we shouldn't even generate code for it... I guess it might require some more thought even outside of this pass.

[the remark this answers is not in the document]
apaszke: In the meantime, how about creating a helper function that deals with traversing types? I see this pattern appearing multiple times, and in the future it is likely that we will want to do something else.

src/lib/OccAnalysis.hs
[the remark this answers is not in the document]
apaszke: I think it's a good idea to have it, as an extra test. I am ok with merging without it, but I would be more confident if we had it.

src/lib/OccAnalysis.hs
apaszke: In that case I think I don't understand this code. Why are you passing `scrutIx` to `occAlt` then?

src/lib/OccAnalysis.hs
```diff
@@ -100,6 +100,14 @@ useManyTimes (FV fvs) = FV $ mapNameMapE update fvs where
...
   -- which indices of each object are used much, but for now there's
   -- no need, because any Unbounded count will inhibit inlining.
+
+  -- TODO(Oddity) This will also set apparent indexing depth to zero.  The
+  -- indexing depth that occurrence analysis counts is relevant for assessing
+  -- work preservation, in that an inlining might only be work-preserving if it
+  -- leads to canceling against an indexing expression (or several).  However,
+  -- no inlining into a `while` loop is work-preserving in any case, so zero is
+  -- sound.  It does look odd on test cases, though, if the binding is clearly
+  -- indexed inside the while.
```
apaszke, line 110: I feel like this loss of precision might be dangerous? From what I understand, you're arguing here that losing the index-depth doesn't matter, because the binder will shows up as having unbounded use. But, as we discussed yesterday, there might be binders that you might want to inline an arbitrary number of times, for as long as they cancel out against the index: broadcasts! Doing the analysis this way might lead us to inline them eagerly, leaving extra loops in the IR.

src/lib/OccAnalysis.hs
```diff
@@ -221,16 +220,23 @@ summary atom = case atom of
...
+  if any (== IxAll) freeIxs then
+    return IxAll
+  else
+    let freeIterationNames = nub $ concatMap freeAtomVarsList freeIxs
+    -- TODO(correctness) Should we detect whether `obj` is an expression with a
+    -- funny effect like `IO` and fall back from `Deterministic` to `IxAll`?
+    -- (For reader, writer, and state effects such a fallback should not be
+    -- necessary, because the summary of the reference should already have any
+    -- necessary `IxAll` in it.)
+    in return $ Deterministic $ freeIterationNames
```
apaszke, line 232: Right, can't we always return IxAll in that case? Since this is still unsound.

src/lib/OccAnalysis.hs
```diff
@@ -462,7 +471,14 @@ instance HasOCC (ComposeE PrimOp Atom) where
...
     void $ oneShot2 accessOnce (Deterministic []) valIx combine
-    -- Do not change it
+    -- Do not change the expression
+    -- TODO(correctness) There is a potential bug here.  If the combining
+    -- function contains some internally dead code, it both won't be removed by
+    -- this traversal and won't be counted as referencing its free variables.
+    -- This could then lead us to make a bad inlining decision about said free
+    -- variables.  Unless I am mistaken, that requires the monoid instance to
+    -- have been defined nested, which I think is rare enough that I'm inclined
+    -- let the possible bug slide until it occurs.
```
apaszke, line 481: Letting the bug slide until it occurs seems like a bad idea to me. Is there a check we can do that triggers a "not implemented" assertion? That would be an acceptable workaround for me.

src/lib/OccAnalysis.hs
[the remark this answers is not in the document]
axch: Done.

src/lib/OccAnalysis.hs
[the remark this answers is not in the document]
axch: Added.

src/lib/OccAnalysis.hs
axch: Subsumed by the new `occTy` helper.

src/lib/OccAnalysis.hs
axch: Here too.

src/lib/OccAnalysis.hs
axch: Future-proofing for when there will be a constructor for `IxExpr` that represents the elimination that a case expression does.  And even barring that, if we get `unknown` to detect effects and emit `Deterministic` sometimes, then the case binder will also be tagged `Deterministic` rather than `IxAll`, which may help precision at some point.

src/lib/OccAnalysis.hs
axch: OK, fixed it.

src/lib/OccAnalysis.hs
axch: Done.

src/lib/OccAnalysis.hs
axch: Actually, I think I found a reasonable way to fix the bug, and did so.

src/lib/OccAnalysis.hs
```diff
@@ -300,11 +289,23 @@ occNest a decls ans = case decls of
...
+checkAllFreeVariablesMentioned e = do
+#ifdef DEX_DEBUG
+  FV fvs <- get
+  forM_ (nameSetToList (freeVarsE e)) \name ->
+    case lookupNameMapE name fvs of
+      Just _ -> return ()
+      Nothing -> error $ "Free variable map missing free variable " ++ show name
+#else
+  void $ return e  -- Refer to `e` in this branch to avoid a GHC warning
+  return ()
```
apaszke, line 302: Add `INLINE` in the `else` branch to make sure this gets fully eliminated

src/lib/OccAnalysis.hs
apaszke: Right, but so you do assume that `scrutIx` always will be either `Deterministic` or `IxAll`, right? Perhaps worth adding that as a comment, with an explanation that proves why this will be the case?

apaszke (approved):

src/lib/OccAnalysis.hs
axch: No, I don't assume anything about the `scrutIx` that's in scope inside `instance HasOCC SExpr`.  But until we have a representation in `IxExpr` for "case analysis removed constructor foo", the `scrutIx` in scope in `occAlt` will end up being `Deterministic <something>` or `IxAll`, based on how well `unknown` works.

src/lib/OccAnalysis.hs
axch: Done.

PR #1145, opened 2022-11-07
Follow-up to PR 1142

dougalm:


apaszke (approved):
Those changes LGTM, but the comment about `projectDictMethod` in simplification is not addressed.

dougalm:
> the comment about projectDictMethod in simplification is not addressed.

Good catch. Just added it.

PR #1153, opened 2022-11-10
Edit Algebra to use monomial multiplication instead of Map.insert to add powers of a variable to monomials.

### LINKED ISSUE #1152: Memory error involving upper- and lower-triangular arrays.
Minimal-ish working reproduction:
```
def LowerTriMat (n:Type) [Ix n] (v:Type) : Type = (i:n)=>(..i)=>v
def UpperTriMat (n:Type) [Ix n] (v:Type) : Type = (i:n)=>(i..)=>v

def foo (n : Nat) : Float =
  (for k. for j:(..k). 0.0, for k. for j:(k..). 0.0) :: (LowerTriMat (Fin n) Float & UpperTriMat (Fin n) Float)
  0.0

foo $ (f_to_n $ 1.0 + 1.0)
> 0.
free(): invalid next size (fast)
Aborted
```

axch:
Map.insert did the wrong thing when the monomial already referred to the variable being added, as when computing
```
sum_{i=0}^k k * i
```

Fixes #1152.

### DIFF
--- a/src/lib/Algebra.hs
+++ b/src/lib/Algebra.hs
@@ -57,7 +57,9 @@ mul :: Polynomial n-> Polynomial n -> Polynomial n
 mul (Polynomial x) (Polynomial y) =
   poly [ (cx * cy, mulMono mx my)
        | (mx, cx) <- toList x, (my, cy) <- toList y]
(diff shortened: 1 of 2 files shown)


dougalm (approved):
Good work tracking this one down!

Issue #1151, opened 2022-11-08
Precisely typed IR variants


dougalm:
### The problem

We have several different variants of our core IR: pre-simplification, post-simplification, and various special forms used in substitutions, including IRs-within-IRs (e.g. simplified atoms within core atoms). Ideally we'd model each IR as a different Haskell type, but that would involve *lots* of boilerplate and duplicated code because they have so many constructors in common. Instead, we've modeled them all using the same Haskell types, `Atom`/`Expr`/`Block`. But it's easy to get them confused. It also means we're reluctant to add new constructors for particular use cases (e.g. a "canonical table" atom used during Imp translation) because then we'd have to handle them everywhere.

To solve this, we want a system that lets us define IR variants that share most of their constructors, but differ in some details. Here are some features we might want from such a system.
 1. Be able to traverse the IR with native Haskell case expressions
 2. Convenient injections/projections instead of many layers of constructor wrappers
 3. No performance penalty
 4. (nice to have) Open world of constructors we can keep adding to. This would be essential if we wanted to package this system as a library, but for our use case we can put up with having to list all the constructors in a central location, as long as we don't have to pattern-match on them in places where they can't occur.

### The idea

There are several ways we could tackle this. My favorite is an idea proposed by Adam which works like this: we add a phantom `IR` parameter to `Atom`, `Expr`, and other IR ADTs. `IR` is a type-level list of allowable IR fragments, e.g. `RecordIRFragment` or `LambdaIRFragment`. The constructors of `Atom`, `Expr` etc. test type-level list membership of the corresponding fragment in the `IR`. So IRs are defined as unions of IR fragments. This hits desiderata 1-3, and only falls down on 4, which we think is the least important for our use case anyway. Here's a illustration:

```haskell
data IRFragment = PairConIRF | PairProjIRF
type IR = [IRFragment]

type family HasFragment (ir::IR) (irf::IRFragment) :: Bool where
  HasFragment '[] irf = False
  HasFragment (irf  ': rest) irf = True
  HasFragment (irf' ': rest) irf = HasFragment rest irf

type HasPairCon  ir = HasFragment ir PairConIRF  ~ True
type HasPairProj ir = HasFragment ir PairProjIRF ~ True

data Expr (ir::IR) where
  Var  ::                   String             -> Expr ir
  App  ::                   Expr ir -> Expr ir -> Expr ir
  Lam  ::                   String -> Expr ir  -> Expr ir
[...]

PR #1155, opened 2022-11-10
Add an `IR` type parameter to the core/simplified IR

dougalm:
This implements just the first step of #1151: it adds the `IR` parameter to everything but doesn't let it bear any load yet. Should we merge it into main as it stands or should we first demonstrate that we can use it to distinguish e.g. core IR from simplified IR. It's a question of how convinced we are that this is the right path. I'm fairly sure that any precisely typed IR variant system will look basically like this, differing only in the details of what `IR` means. For me, the big unknown is whether the correctness benefits will outweigh the typing burden. Thoughts?

### DIFF
--- a/src/lib/Builder.hs
+++ b/src/lib/Builder.hs
@@ -88,49 +88,49 @@ import Core
 -- === Ordinary (local) builder class ===
 
 class (EnvReader m, EnvExtender m, Fallible1 m)
-      => Builder (m::MonadKind1) where
-  emitDecl :: Emits n => NameHint -> LetAnn -> Expr n -> m n (AtomName n)
+      => Builder (r::IR) (m::MonadKind1) | m -> r where
+  emitDecl :: Emits n => NameHint -> LetAnn -> Expr r n -> m n (AtomName r n)
 
-class Builder m => ScopableBuilder (m::MonadKind1) where
+class Builder r m => ScopableBuilder (r::IR) (m::MonadKind1) where
   buildScoped
     :: SinkableE e
     => (forall l. (Emits l, DExt n l) => m l (e l))
-    -> m n (Abs (Nest Decl) e n)
+    -> m n (Abs (Nest (Decl r)) e n)
 
-type Builder2 (m :: MonadKind2) = forall i. Builder (m i)
-type ScopableBuilder2 (m :: MonadKind2) = forall i. ScopableBuilder (m i)
+type Builder2         (r::IR) (m :: MonadKind2) = forall i. Builder         r (m i)
+type ScopableBuilder2 (r::IR) (m :: MonadKind2) = forall i. ScopableBuilder r (m i)
 
-emit :: (Builder m, Emits n) => Expr n -> m n (AtomName n)
+emit :: (Builder r m, Emits n) => Expr r n -> m n (AtomName r n)
 emit expr = emitDecl noHint PlainLet expr
 {-# INLINE emit #-}
 
-emitHinted :: (Builder m, Emits n) => NameHint -> Expr n -> m n (AtomName n)
+emitHinted :: (Builder r m, Emits n) => NameHint -> Expr r n -> m n (AtomName r n)
 emitHinted hint expr = emitDecl hint PlainLet expr
 {-# INLINE emitHinted #-}
 
-emitOp :: (Builder m, Emits n) => Op n -> m n (Atom n)
+emitOp :: (Builder r m, Emits n) => Op r n -> m n (Atom r n)
 emitOp op = Var <$> emit (Op op)
 {-# INLINE emitOp #-}
 
-emitUnOp :: (Builder m, Emits n) => UnOp -> Atom n -> m n (Atom n)
+emitUnOp :: (Builder r m, Emits n) => UnOp -> Atom r n -> m n (Atom r n)
 emitUnOp op x = emitOp $ UnOp op x
 
-emitBlock :: (Builder m, Emits n) => Block n -> m n (Atom n)
+emitBlock :: (Builder r m, Emits n) => Block r n -> m n (Atom r n)
 emitBlock (Block _ decls result) = emitDecls decls result
 
-emitDecls :: (Builder m, Emits n, SubstE Name e, SinkableE e)
-          => Nest Decl n l -> e l -> m n (e n)
+emitDecls :: (Builder r m, Emits n, SubstE Name e, SinkableE e)
+          => Nest (Decl r) n l -> e l -> m n (e n)
 emitDecls decls result = runSubstReaderT idSubst $ emitDecls' decls result
 
-emitDecls' :: (Builder m, Emits o, SubstE Name e, SinkableE e)
-           => Nest Decl i i' -> e i' -> SubstReaderT Name m i o (e o)
+emitDecls' :: (Builder r m, Emits o, SubstE Name e, SinkableE e)
+           => Nest (Decl r) i i' -> e i' -> SubstReaderT Name m i o (e o)
 emitDecls' Empty e = substM e
 emitDecls' (Nest (Let b (DeclBinding ann _ expr)) rest) e = do
   expr' <- substM expr
   v <- emitDecl (getNameHint b) ann expr'
   extendSubst (b @> v) $ emitDecls' rest e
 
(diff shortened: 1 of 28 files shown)


apaszke (commented):
This is mostly looking good. My main concern would be the few unsafe IR coercions that are in the code. Do they have to be there, or is that just an artifact of an incomplete change?

dougalm:
> This is mostly looking good. My main concern would be the few unsafe IR coercions that are in the code. Do they have to be there, or is that just an artifact of an incomplete change?

Yes, just an artifact of an incomplete change. Fixing them will require changes beyond just changing type signatures, so I wanted to do that in a separate change. For example, we might need to have two different `App` constructs, one for CoreIR and one for SimpIR. The current change is huge in terms of loc but at least it's mostly mechanical. I explained this a bit in the commit message.

axch:
Looks from the tests like you forgot to update the SPECIALIZE pragma at the bottom of Simplify.hs (which is perhaps ignored in the default build which doesn't ask GHC to optimize?).

src/lib/Simplify.hs
```diff
@@ -737,10 +739,10 @@ simplifyOp op = case op of
     Left  a   -> return a
     Right op' -> emitOp op'
 
-pattern IdentityReconAbs :: Abs binder ReconstructAtom n
+pattern IdentityReconAbs :: Abs binder (ReconstructAtom SimpIR) n
```
axch, line 742: All the other reconstructions are CoreIR; why is this SimpIR?

src/lib/Simplify.hs
```diff
@@ -831,17 +833,17 @@ simplifyHof hint hof = case hof of
   Seq _ _ _ _ -> error "Shouldn't ever see a Seq in Simplify"
   RememberDest _ _ -> error "Shouldn't ever see a RememberDest in Simplify"
 
-simplifyBlock :: Emits o => Block i -> SimplifyM i o (Atom o)
+simplifyBlock :: Emits o => SBlock i -> SimplifyM i o (SAtom o)
```
axch, line 836: SBlock or CBlock here?

src/lib/Core.hs
```diff
@@ -652,27 +652,27 @@ infixr 1 ?-->
...
-fromNaryLamExact :: Int -> Atom n -> Maybe (NaryLamExpr n)
+fromNaryLamExact :: Int -> Atom r n -> Maybe (NaryLamExpr r n)
 fromNaryLamExact exactDepth _ | exactDepth <= 0 = error "expected positive number of args"
 fromNaryLamExact exactDepth lam = do
   (realDepth, naryLam) <- fromNaryLam exactDepth lam
   guard $ realDepth == exactDepth
   return naryLam
 
-fromNaryLam :: Int -> Atom n -> Maybe (Int, NaryLamExpr n)
+fromNaryLam :: Int -> Atom r n-> Maybe (Int, NaryLamExpr r n)
```
axch, line 675: Nit: missing space between `n` and `->`

src/lib/Builder.hs
```diff
@@ -296,21 +295,21 @@ addInstanceSynthCandidate :: TopBuilder m => ClassName n -> InstanceName n -> m
...
+emitAtomRules :: TopBuilder m => AtomName r n -> AtomRules n -> m n ()
 emitAtomRules v rules = emitNamelessEnv $
   TopEnvFrag emptyOutFrag $ mempty { fragCustomRules = CustomRules $ M.singleton v rules }
 
-emitTopLet :: (Mut n, TopBuilder m) => NameHint -> LetAnn -> Expr n -> m n (AtomName n)
+emitTopLet :: (Mut n, TopBuilder m) => NameHint -> LetAnn -> Expr r n -> m n (AtomName r n)
 emitTopLet hint letAnn expr = do
   ty <- getType expr
-  emitBinding hint $ AtomNameBinding $ LetBound  (DeclBinding letAnn ty expr)
+  emitBinding hint $ AtomNameBinding $ unsafeCoerceIRE $ LetBound (DeclBinding letAnn ty expr)
```
axch, line 305: Why is this coercion necessary?

axch (approved):
Well, this must have been an immense amount of tedious waiting for GHC to tell you the next kind error!

I second Adam's worry about coercions; but assuming that's addressed satisfactorily, this is good to merge.

Also a few nitpicks below, and the tests are suggesting to check SPECIALIZE pragmas.

axch:
Re: whether this is worthwhile: I actually found the expanded type annotations helpful already (assuming they are correct).  How hard will it be to get this to type-check without the coercions?

src/lib/Builder.hs
dougalm: Because currently the top bindings are all `CoreIR`

src/lib/Types/Core.hs
```diff
@@ -46,55 +47,88 @@ import Types.Primitives
...
+data DeclBinding r n = DeclBinding LetAnn (Type r n) (Expr r n)
      deriving (Show, Generic)
-data Decl n l = Let (NameBinder AtomNameC n l) (DeclBinding n)
+data Decl (r::IR) (n::S) (l::S) = Let (AtomNameBinder n l) (DeclBinding r n)
      deriving (Show, Generic)
-type Decls = Nest Decl
+type Decls r = Nest (Decl r)
+
+-- TODO: make this a newtype with an unsafe constructor
+type AtomName (r::IR) = Name AtomNameC
```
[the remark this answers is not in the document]
dougalm: The idea is that the `IR` parameter will play a role a bit like the `C` parameter. Knowing that you have an `AtomName r n` means that when you look up the name in the environment, you're sure to get a result in the correct IR. To maintain that invariant, we'll need a corresponding IR-parameterized AtomNameBinder too.

src/lib/Types/Core.hs
```diff
@@ -46,55 +47,88 @@ import Types.Primitives
 import Types.Source
 import Types.Imp
 
+-- === IR variants ===
+
+data IR = CoreIR
+type SimpIR = CoreIR  -- TODO: actually distinguish these two
```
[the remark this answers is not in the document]
dougalm: Yes, I meant for `CoreIR` to be the IR before simplification and `SimpIR` to be the IR after simplification.

src/lib/TopLevel.hs
```diff
@@ -720,7 +720,7 @@ linkFunObjCode objCode dyvarStores (LinktimeVals funVals ptrVals) = do
 
 -- Get the definition of a specialized function in the pre-simplification IR.
 specializedFunPreSimpDefinition
-  :: (MonadFail1 m, EnvReader m) => AtomName n -> m n (NaryLamExpr n)
+  :: (MonadFail1 m, EnvReader m) => CAtomName n -> m n (NaryLamExpr SimpIR n)
```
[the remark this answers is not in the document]
dougalm: Yes, you're right. It should be `naryLamExpr CoreIR`

src/lib/Simplify.hs
dougalm: Honestly, it should probably be a different third thing. It's going to have features of both.

src/lib/Simplify.hs
dougalm: Yeah, should be CBlock.

dougalm:
>  I actually found the expanded type annotations helpful already (assuming they are correct). 

I should have said explicitly: they are not at all correct! There are plenty of errors I didn't know about (like the ones you noticed) but also there are also errors I do know about. For example, I used `Atom SimpIR o` for the results of functions like `simplifyExpr` because it was suggestive but it's actually totally wrong. What we're really going to need during the simplification pass is a sort of hybrid atom: an atom that's mostly CoreIR, but whose free variables are all SimpIR variables.

It felt useful to throw around `CoreIR` and `SimpIR` to hint at intent, but please don't interpret them to mean anything more than that yet.

> How hard will it be to get this to type-check without the coercions?

We'll see. We should be able to do it incrementally, so we might have a better idea once we handle a few IRs/constructors.

src/lib/Types/Core.hs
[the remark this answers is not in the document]
axch: Might be good to add a comment to that effect.

axch:
The SPECIALIZE pragma still doesn't compile :(

You can test it locally with `make build-ffis` (`make build-opt` should work too, since it's not in the ffi).

PR #1158, opened 2022-11-14
Start adding predicates to IR variants.

dougalm:
To test it out I restricted the `BoxedRef` constructor to only be allowed during the simp-to-imp translation. It worked! GHC even gave warnings about all the now-inaccessible pattern-matching on `BoxedRef` we can now safely delete.

### DIFF
--- a/src/lib/Imp.hs
+++ b/src/lib/Imp.hs
@@ -38,6 +38,12 @@ import Algebra
 import RawName qualified as R
 
 type AtomRecon = Abs (Nest (NameBinder AtomNameC)) SAtom
+type SIAtom   = Atom  SimpToImpIR
+type SIType   = Type  SimpToImpIR
+type SIBlock  = Block SimpToImpIR
+type SIExpr   = Expr  SimpToImpIR
+type SIDecl   = Decl  SimpToImpIR
+type SIDecls  = Decls SimpToImpIR
 
 type PtrBinder = IBinder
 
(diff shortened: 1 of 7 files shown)


src/lib/Imp.hs
```diff
@@ -314,26 +328,26 @@ buildRecon b x = do
   return (xs, recon)
 
 translateBlock :: forall i o. Emits o
-               => MaybeDest o -> SBlock i -> SubstImpM i o (SAtom o)
+               => MaybeDest o -> Block SimpToImpIR i -> SubstImpM i o (SIAtom o)
```
axch, line 331: Nit: SIBlock?

src/lib/Imp.hs
```diff
@@ -864,34 +878,34 @@ buildAbsHoistingDeclsDest hint binding cont =
...
 instance SinkableE DestPtrInfo
 instance HoistableE  DestPtrInfo
 instance SubstE Name DestPtrInfo
-instance SubstE (AtomSubstVal SimpIR) DestPtrInfo
+instance SubstE (AtomSubstVal SimpToImpIR) DestPtrInfo
 
 -- === Destination builder ===
 
-type Dest = SAtom  -- has type `Ref a` for some a
+type Dest = Atom SimpToImpIR  -- has type `Ref a` for some a
```
axch, line 908: Nit: SIAtom ?

axch (approved):

PR #1159, opened 2022-11-15
Use `ImpName` instead of `AtomName` in ImpIR.

dougalm:
We couldn't do this before because we carry an `Atom` in the Simp->Imp substitution and we didn't want to add an `AtomicImpName` case to `Atom` just for this single use. But with the new IR system that's fine! We add the case and signal that it doesn't have to be handled anywhere else by giving it the `Sat r (Is SimpToImpIR)` constraint.

One awkward part of the change is that now we have a mix of atom names and imp names in the block that we want to hand to Algebra. I resolved this by rebinding the imp names to atom names before calling into Algebra, but it's a bit ugly.


src/lib/Types/Core.hs
```diff
@@ -2432,6 +2445,55 @@ instance Store IxMethod
...
+instance SubstE IAtomSubstVal (FieldRowElems SimpToImpIR)
+instance SubstE IAtomSubstVal (DepPairType SimpToImpIR)
+instance SubstE IAtomSubstVal (DataDefParams SimpToImpIR)
+instance SubstE IAtomSubstVal (PiType SimpToImpIR)
+instance SubstB IAtomSubstVal (PiBinder SimpToImpIR)
+instance SubstE IAtomSubstVal (DictExpr SimpToImpIR)
+instance SubstE IAtomSubstVal (DictType SimpToImpIR)
+instance SubstE IAtomSubstVal (TabLamExpr SimpToImpIR)
+instance SubstE IAtomSubstVal (TabPiType SimpToImpIR)
+instance SubstE IAtomSubstVal (IxType SimpToImpIR)
```
apaszke, line 2495: Oof that's a lot of extra code to generate. Why do we need this?

src/lib/Types/Imp.hs
```diff
@@ -33,16 +33,15 @@ import Util (IsBool (..))
...
 
--- We use AtomName because we convert between atoms and imp
--- expressions without chaning names. Maybe we shouldn't do that.
-type ImpName = Name AtomNameC
+type ImpName = Name ImpNameC
 
 type ImpFunName = Name ImpFunNameC
 data IExpr n = ILit LitVal
              | IVar (ImpName n) BaseType
+             | IPtrLit (Name PtrNameC n) PtrType
```
apaszke, line 41: Well it's a pointer var, not a pointer literal, right?

src/lib/TopLevel.hs
```diff
@@ -4,6 +4,9 @@
 -- license that can be found in the LICENSE file or at
 -- https://developers.google.com/open-source/licenses/bsd
 
+-- This is only for the `SubstE (SubstVal ImpNameC CAtom) CAtom` instance, which
+-- we plan to delete soon.
```
apaszke, line 8: Please add a comment next to that instance, to make sure we remember to do so!

src/lib/TopLevel.hs
```diff
@@ -775,7 +778,13 @@ evalLLVM block = do
   resultVals <-
     liftIO $ callNativeFun nativeFun benchRequired logger [] resultTypes
   resultValsNoPtrs <- mapM litValToPointerlessAtom resultVals
-  applyNaryAbs reconAtom $ (map SubstVal resultValsNoPtrs :: [CAtomSubstVal AtomNameC n])
+  -- TODO: this unsafe coerce won't be needed once we treat everything as a
+  -- function, because then we won't need the reconstruction at all. If we were
```
apaszke, line 782: what does "once we treat everything as a function" mean?

src/lib/Imp.hs
```diff
@@ -1396,7 +1398,20 @@ computeOffset idxNest' idxs = do
...
   ab <- hoistDecls i body
-  sumUsingPolys lim ab
+  PairE lim' ab' <- emitImpVars (PairE lim ab) -- Algebra only works with Atom vars
+  sumUsingPolys lim' ab'
+
+emitImpVars ::
+  (Emits n, HoistableE e, SubstE (SubstVal ImpNameC SIAtom) e, SinkableE e, SubstE Name e)
+  => e n -> SBuilderM n (e n)
+emitImpVars e = do
+  let impVars = nameSetToList (freeVarsE e) :: [ImpName _]
```
apaszke, line 1408: You could consider using type applications in place of partial type signatures

src/lib/Imp.hs
```diff
@@ -1396,7 +1398,20 @@ computeOffset idxNest' idxs = do
...
+  Abs impBs e' <- return $ abstractFreeVarsNoAnn impVars e
+  substVals <- forM impVars \v -> do
+    ty <- impVarType v
+    atomVar <- emitAtomToName (getNameHint v) (AtomicIVar (LeftE v) ty)
+    return $ (SubstVal (Var atomVar) :: SubstVal ImpNameC SIAtom ImpNameC _)
+  applySubst (impBs @@> substVals) e'
```
apaszke, line 1414: Why do we never have to undo this imp->core variable change?

src/lib/Types/Core.hs
dougalm: We use it in two places. (1) the atom reconstruction at the top level. This will go away once we can work with canonical atoms, and (2) we rename imp names to core names before calling into algebra because otherwise we would have a mix. I'd like to delete this soon, but I don't know how to solve (2). Any ideas?

src/lib/Types/Imp.hs
dougalm: Good point(er). I'll rename it.

src/lib/TopLevel.hs
dougalm: Specifically I mean unify the compilation paths around first-order monomorphic ("FOMO") functions, including top blocks, which get treated as nullary functions.

src/lib/Imp.hs
dougalm: This is in the (Simp) builder monad. We bring things back to Imp by translating the resulting Simp block to an Imp block.

src/lib/Types/Core.hs
apaszke: Well the big part that I don't understand is: why do we end up with this mix of IRs in algebra? Is it because some names might refer to top-level values, while some might be defined locally in the computation? Aren't we inlining all top level values anyway?

dougalm: Right, top-level names aren't the issue. The reason we have a mix is that there are Imp names due to ordinary local Imp bindings (e.g. some runtime computed `n`), but there are also Atom names due to index binders in the dependent table types. This is because we do the algebra stuff on the core/simplified IR. If we switched to doing it in Imp IR I think we'd be fine.

PR #1160, opened 2022-11-16
[WIP] Use destination-passing style at the top level, not just for internal calls

dougalm:
This avoids the atom reconstruction we had to do after the imp pass, which was always a pain to maintain. It should let us merge the different compilation paths into one, standardizing on FOMO functions as the core compilation unit.

### DIFF
--- a/src/lib/Runtime.hs
+++ b/src/lib/Runtime.hs
@@ -34,13 +34,13 @@ import CUDA (synchronizeCUDA)
 -- === One-shot evaluation ===
 
 foreign import ccall "dynamic"
-  callFunPtr :: DexExecutable -> Int32 -> Ptr () -> Ptr () -> IO DexExitCode
+  callFunPtr :: DexExecutable -> Int32 -> Ptr () -> IO DexExitCode
 
 foreign import ccall "dynamic"
   callDtor :: DexDestructor -> IO ()
 
 type DexDestructor = FunPtr (IO ())
-type DexExecutable = FunPtr (Int32 -> Ptr () -> Ptr () -> IO DexExitCode)
+type DexExecutable = FunPtr (Int32 -> Ptr () -> IO DexExitCode)
 type DexExitCode = Int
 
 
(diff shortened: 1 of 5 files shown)


src/lib/Runtime.hs
```diff
@@ -105,11 +102,11 @@ runBench run = do
...
 
-loadLitVals :: MonadIO m => Ptr () -> [BaseType] -> m [LitVal]
-loadLitVals p types = zipWithM loadLitVal (ptrArray p) types
+_loadLitVals :: MonadIO m => Ptr () -> [BaseType] -> m [LitVal]
+_loadLitVals p types = zipWithM loadLitVal (ptrArray p) types
 
-freeLitVals :: MonadIO m => Ptr () -> [BaseType] -> m ()
-freeLitVals p types = zipWithM_ freeLitVal (ptrArray p) types
+_freeLitVals :: MonadIO m => Ptr () -> [BaseType] -> m ()
+_freeLitVals p types = zipWithM_ freeLitVal (ptrArray p) types
```
apaszke, line 109: Those probably shouldn't be unused?

src/lib/TopLevel.hs
```diff
@@ -775,18 +775,44 @@ evalLLVM block = do
...
+  dest@(TopDest _ destPtrs) <- makeDestTopLevel destTy
+  liftIO $ callNativeFun nativeFun benchRequired logger (PtrLit <$> destPtrs)
+  loadDestTopLevel dest
+
 {-# SCC evalLLVM #-}
 
+data TopDest (n::S) = TopDest (Atom SimpToImpIR n) [PtrLitVal]
+
+makeDestTopLevel :: (Mut n, Topper m) => SType n -> m n (TopDest n)
+makeDestTopLevel (RawRefTy valTy) = do
```
apaszke, line 787: Can we ensure that this `valTy` is simple? I'm a little afraid of leaking `makeDest` all the way up to here, since the plan is to make `Imp` a lot dumber than it is today. It shouldn't be able to handle most of the types we throw at it today

apaszke (commented):
Just some initial thoughts.

dougalm:
Superseded by #1173

PR #1165, opened 2022-11-19
Replace ReifiedSat with WhenE for IR-dependent generic representations.

axch:
The difference is that `ReifiedSat` was just an object carrying a Sat constraint, whereas `WhenE` carries a payload as well---but if and only if the predicate holds.

The goal is to reduce the instance burden on generically-derived traversals over inhabitants of specific IRs.  To wit, consider defining a class `Foo` that's meant to be implemented generically on `Atom SimpIR`:
```
class Foo e where
  method :: ...
  default method :: (GenericE e, Foo (RepE e)) => ...
  method = ...

instance Foo (Atom SimpIR)
```
We know Atom has constructors that cannot occur in SimpIR (e.g., `BoxedRef`).  Those constructors may (and in this case do) mention types that occur nowhere else in Atom (e.g., `Abs (NonDepNest ...) (Atom r)`). More to the point, the RepE for Atom includes such types, due to that constructor.  Therefore, for GHC to generate the `Foo (Atom SimpIR)` instance generically, it needs an instance of `Foo` that covers `Abs (NonDepNest ...) (Atom r)`, even though such an `Abs` can never appear in an `Atom SimpIR`.

Replacing `ReifiedSat` with `WhenE` solves this problem.  Instead of supplying an instance of `Foo` for every impossible sub-type in Atom, it suffices to add a single pair of boilerplate instances of `Foo` for `WhenE`, of the form

```
instance Foo e => Foo (WhenE True e) where
  method (WhenE e) = method e  -- Adjusting types mutatis mutandis
instance Foo (WhenE False e) where
  method _ = error "Unreachable"
```

This came up now because the inliner contains such a class. Specifically, I'm trying to write a generic `Abs` instance, that suffices to cover a fairly generic `Atom` instance, but I only want to deal with one binder at a time; and `BoxedRef` is the only `Atom` constructor that has an `Abs` with more than one binder.

Two notes:

- `WhenE` imposes some cost (in the form of boilerplate instances) on classes that are parameterically polymorphic in the IR (like `instance AlphaHashableE (Atom r)`), but it's the same as the cost of `PairE` or `UnitE` or similar.

- This mechanism does not, as far as I know, help classes that are meant to be conditionally polymorphic in the IR, such as a hypothetical `instance (Sat r SomePredicate) => Bar (Atom r)`.  Even if `SomePredicate` logically implies that some sub-type cannot appear in `Atom r`, GHC may not be smart enough to figure out that the corresponding instance of `Bar` is not necessary.

### DIFF
--- a/src/lib/Name.hs
+++ b/src/lib/Name.hs
@@ -30,7 +30,7 @@ module Name (
   IsVoidS (..), UnitE (..), VoidE, PairE (..), toPairE, fromPairE,
   ListE (..), ComposeE (..), MapE (..), NonEmptyListE (..),
   EitherE (..), leftsE, rightsE, forgetEitherE,
-  LiftE (..), EqE, EqV, EqB, OrdE, OrdV, OrdB, VoidB,
+  LiftE (..), WhenE (..), EqE, EqV, EqB, OrdE, OrdV, OrdB, VoidB,
   EitherB (..), forgetEitherB, BinderP (..),
   LiftB, pattern LiftB,
   HashMapE (..), HashableE, nestToNames,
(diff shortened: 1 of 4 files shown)


dougalm (approved):
Love it! I'm going to merge it right away because it'll help the patch I'm currently working on.

src/lib/Name.hs
```diff
@@ -628,6 +628,16 @@ newtype ComposeE (f :: * -> *) (e::E) (n::S) =
   ComposeE { fromComposeE :: (f (e n)) }
   deriving (Show, Eq, Generic)
 
+data WhenE (p::Bool) (e::E) (n::S) where
+  WhenE :: (p ~ True) => e n -> WhenE p e n
```
apaszke, line 632: The `(p ~ True)` part seems unnecessary? This would be equivalent, but simpler:
```
WhenE :: e n -> WhenE True e n
```

PR #1168, opened 2022-11-23
Traverse the type of a dict hole to be synthesized when synthesizing.

axch:
Failing to do so created a corner case where an unsynthesized Ix dict could be left in the IR after synthesis.  This was masked because that dict was never accessed down-stream, but a version of the inliner moved it to a place where it _was_ accessed, revealing the existence of a bug.


dougalm (approved):

src/lib/PPrint.hs
```diff
@@ -361,12 +361,18 @@ instance Pretty (PiType r n) where
...
       Pi subpi -> pretty subpi
       _ -> pLowest body
-    in prettyBinder <> (group $ line <> "=>" <+> prettyBody)
+    in prettyBinder <> prettyIxDict dict <> (group $ line <> "=>" <+> prettyBody)
+
+-- A helper to let us turn dict printing on and off.  We mostly want it off to
+-- reduce clutter in prints and error messages, but when debugging synthesis we
+-- want it on.
+prettyIxDict :: Atom r n -> Doc ann
+prettyIxDict dict = if False then " " <> p dict else mempty
```
apaszke, line 375: You could enable it based on an env var. We do that same trick in a few other places I think

PR #1169, opened 2022-11-25
Preliminary work on separating SimpIR from CoreIR.

axch:
Method: Make `SimpIR` a separate (but isomorphic) IR constructor to `CoreIR` and fix compilation errors.

I didn't finish, so I rolled back the impetus change.  So, `SimpIR` is still an alias for `CoreIR`, but the correct one is referred to in more places.

The functional change is that attempting to `substE` at `(AtomSubstVal CoreIR)` into the `LoweredTopFun` constructor of `TopFunBinding` is now a runtime error, on the grounds that `LoweredTopFun` carries `SimpIR` (which I infer from its name and type-checking how it is constructed).  This runtime error does not occur in the test suite.

I encountered three problems that stopped me from making further progress:

**Custom Linearizations**

The content of a custom linearization rule is typed as `Atom CoreIR`, and at runtime is presumably always a lambda.  However, `linearize` just directly constructs an application thereof in `SimpIR`, using `naryApp`.

Is this just a bug?  Should we call `simplifyApp` from `linearizeExpr`?  (Maybe the only reason it isn't currently done that way is the circular module dependency it would cause between Simplify.hs and Linearize.hs.)

On the other hand, that `Atom CoreIR` is just the reconstruction of a definition that was already simplified (when the "create custom linearization" directive was processed).  Is there perhaps some reason to expect that such a reconstruction can just be added to `SimpIR` without trouble?  If so, should we express that reason to GHC?

**Simplification substitutions**

The simplifier's whole job is to change from `CoreIR` to `SimpIR`. However, the substitution it carries is typed `AtomSubstVal CoreIR`, so doesn't have a proof that it did its job!  I am somewhat confused by the comment about top-level function names being an exception; but whatever the exception is, perhaps we can express it to GHC?

**Reconstructions**

What is the IR typology of reconstructions and ReconAbs?  What information are they supposed to carry, and what are they even used for at the end of the day?  In particular, `applyRecon` seems to be used a lot to emit `SimpIR` inside `simplifyHof`, but `simplifyTopBlock` seems to want to be able to use the recon to carry `CoreIR`.

Is there a chance that this problem might be solved by capturing the constraint on where `Lam` atoms can appear after simplification, and what we know about their bodies before and after?  And maybe `Lam`s that float all the way to the top level are allowed to be different from those that stop at a `PrimHof`?

### DIFF
--- a/src/lib/Simplify.hs
+++ b/src/lib/Simplify.hs
@@ -102,12 +102,14 @@ data WillLinearize = WillLinearize | NoLinearize
 -- This is because top-level functions can have special (AD) rules associated with
 -- them, and we try to preserve their identity for as long as we can. Since the only
 -- elimination form for function is application, we do some extra handling in simplifyApp.
-newtype SimplifyM (i::S) (o::S) (a:: *) = SimplifyM
+newtype SimplifyMP (r::IR) (i::S) (o::S) (a:: *) = SimplifyM
   { runSimplifyM'
-    :: SubstReaderT (AtomSubstVal CoreIR) (DoubleBuilderT SimpIR TopEnvFrag (ReaderT WillLinearize HardFailM)) i o a }
+    :: SubstReaderT (AtomSubstVal r) (DoubleBuilderT SimpIR TopEnvFrag (ReaderT WillLinearize HardFailM)) i o a }
   deriving ( Functor, Applicative, Monad, ScopeReader, EnvExtender, Fallible
-           , EnvReader, SubstReader (AtomSubstVal CoreIR), MonadFail
+           , EnvReader, SubstReader (AtomSubstVal r), MonadFail
            , Builder SimpIR, HoistingTopBuilder TopEnvFrag)
+
+type SimplifyM = SimplifyMP CoreIR
 
 liftSimplifyM
   :: (SinkableE e, SubstE Name e, TopBuilder m, Mut n)
@@ -546,7 +548,7 @@ simplifyLam atom = case atom of
     _ -> error "Not a lambda expression"
   where
     doSimpLam :: LamBinder CoreIR i i' -> CBlock i'
-      -> SimplifyM i o (SAtom o, Abs (LamBinder SimpIR) (ReconstructAtom SimpIR) o)
+      -> SimplifyM i o (SAtom o, Abs (LamBinder CoreIR) (ReconstructAtom CoreIR) o)
     doSimpLam b body = do
       (Abs b' body', recon) <- simplifyAbs $ Abs b body
       return $! (Lam $ LamExpr b' body', recon)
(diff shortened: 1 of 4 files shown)


src/lib/Types/Core.hs
```diff
@@ -1981,10 +1984,17 @@ instance GenericE TopFunBinding where
 instance SinkableE TopFunBinding
 instance HoistableE  TopFunBinding
 instance SubstE Name TopFunBinding
-instance SubstE (AtomSubstVal CoreIR) TopFunBinding
 instance AlphaEqE TopFunBinding
 instance AlphaHashableE TopFunBinding
 
+instance SubstE (AtomSubstVal CoreIR) TopFunBinding where
```
dougalm, line 1990: Did you figure out why we even need this instance? I would expect that TopFunBinding would need `SubstE Name` and nothing else.

src/lib/Simplify.hs
```diff
@@ -102,13 +102,15 @@ data WillLinearize = WillLinearize | NoLinearize
 -- This is because top-level functions can have special (AD) rules associated with
 -- them, and we try to preserve their identity for as long as we can. Since the only
 -- elimination form for function is application, we do some extra handling in simplifyApp.
-newtype SimplifyM (i::S) (o::S) (a:: *) = SimplifyM
+newtype SimplifyMP (r::IR) (i::S) (o::S) (a:: *) = SimplifyM
```
dougalm, line 105: I don't think we should make the simplification monad polymorphic in the IR parameter. I realize you did this to handle the `catch` traversal, but I think we should do it differently. To do IR-typed simplification properly, we'll want to have three IR variants at play: CoreIR for the IR that goes into simplification, SimpIR for the IR that comes out, and CoreToSimpIR for the substitution.  CoreToSimpIR will allow SimpIR atoms embedded within CoreIR atoms. I've been using this pattern for the Imp traversal and it works well: https://github.com/google-research/dex-lang/pull/1173.

dougalm (changes requested):
Thanks for trying this out and figuring out what we need to fix.

> However, linearize just directly constructs an application thereof in SimpIR

Actually, I think linearize should be producing CoreIR. The result of linearize gets subsequently simplified.

> The simplifier's whole job is to change from CoreIR to SimpIR. However, the substitution it carries is typed AtomSubstVal CoreIR, so doesn't have a proof that it did its job!

Right, I think the substitution needs to carry a third type of IR. Let's call it CoreToSimpIR. It allows CoreIR constructs like lambda (which it needs, since it's the mechanism for inlining these). But it contains `SimpIR` atoms as its "leaves". For any atom whose type is "data", it should be possible to translate the `Atom CoreToSimpIR` to `Atom SimpIR` without error.


> Reconstructions

Reconstructions are also a funny mix. They reconstruct Core atoms from Simp data.

axch:
Subsumed by #1179.

PR #1170, opened 2022-11-28
First inliner.

axch:
The inliner follows the architecture of the GHC inliner, as described in [Secrets of the Glasgow Haskell Compiler Inliner](https://www.microsoft.com/en-us/research/wp-content/uploads/2002/07/inline.pdf), with a few notable differences:

- For now, I simply ignore the possibility of inlining a binding at some use sites but not others.  It's only safe to do that in Dex for zero-work bindings, so I would want to think about it more to adapt the Secrets algorithm to our situation.
- Unlike Secrets, this inliner takes pains to try to inline non-zero-work bindings into loops --- that's how we get loop fusion!  This makes the required occurrence analysis significantly more elaborate.
- For now, this inliner does not actually measure code size, effectively assuming that anything more than a variable reference is "large"; and therefore hesitates to inline it into more than one place.  This misses opportunities to inline into the arms of case statements, etc.
- For now, this inliner only does one local optimization during inlining, namely beta reduction of table indexing expressions (which is necessary for the inliner to be work-preserving).  Per Secrets, it's probably a good idea to do more, so that fewer passes of inlining + peephole optimization suffice for convergence.
- For now, Dex only runs the inliner once, immediately after simplification and before any other passes.  We should iterate on that, measuring compile-time and run-time performance.

This inliner also leaves at least one known Dex-specific opportunity on the table, one we might call "pointless-array elimination".  To wit, consider an array whose body is an atom, `xs = for i. <atom>`.  This array is not an atom itself, and will do work if materialized at runtime, allocating and filling the array.  But, if it's inlined into positions where it is indexed, it becomes an atom, which we treat as zero work.  Ergo, such an inlining is profitable in more situations than the current setup can identify, such as `for j k. xs.k`.  (If the body of `xs` did work, inlining `xs` would duplicate that work.  If `xs` wasn't indexed at all, then inlining would duplicate the work or array creation regardless of the body.  And right now the inliner can't detect the situation where a binding does work if not beta-reduced but does no work if beta-reduced, partly because that situation does not arise in GHC core.)

### DIFF
--- a/src/lib/Builder.hs
+++ b/src/lib/Builder.hs
@@ -11,6 +11,7 @@
 module Builder (
   emit, emitHinted, emitOp, emitUnOp,
   buildPureLam, BuilderT (..), Builder (..), ScopableBuilder (..),
+  buildScopedAssumeNoDecls,
   Builder2, BuilderM, ScopableBuilder2,
   liftBuilderT, buildBlock, withType, absToBlock, app, add, mul, sub, neg, div',
   iadd, imul, isub, idiv, ilt, ieq, irem,
@@ -29,7 +30,7 @@ module Builder (
   emitDataDef, emitClassDef, emitInstanceDef, emitDataConName, emitTyConName,
   emitEffectDef, emitHandlerDef, emitEffectOpDef,
   buildCase, emitMaybeCase, buildSplitCase,
-  emitBlock, emitDecls, BuilderEmissions, emitAtomToName,
+  emitBlock, emitDecls, BuilderEmissions, emitExprToAtom, emitAtomToName,
   TopBuilder (..), TopBuilderT (..), liftTopBuilderTWith,
   runTopBuilderT, TopBuilder2, emitBindingDefault,
   emitSourceMap, emitSynthCandidates, addInstanceSynthCandidate,
@@ -113,6 +114,7 @@ emitOp op = Var <$> emit (Op op)
 
 emitUnOp :: (Builder r m, Emits n) => UnOp -> Atom r n -> m n (Atom r n)
 emitUnOp op x = emitOp $ UnOp op x
+{-# INLINE emitUnOp #-}
 
 emitBlock :: (Builder r m, Emits n) => Block r n -> m n (Atom r n)
 emitBlock (Block _ decls result) = emitDecls decls result
@@ -129,9 +131,24 @@ emitDecls' (Nest (Let b (DeclBinding ann _ expr)) rest) e = do
   v <- emitDecl (getNameHint b) ann expr'
   extendSubst (b @> v) $ emitDecls' rest e
 
+emitExprToAtom :: (Builder r m, Emits n) => Expr r n -> m n (Atom r n)
+emitExprToAtom (Atom atom) = return atom
+emitExprToAtom expr = Var <$> emit expr
+{-# INLINE emitExprToAtom #-}
+
 emitAtomToName :: (Builder r m, Emits n) => NameHint -> Atom r n -> m n (AtomName r n)
 emitAtomToName _ (Var v) = return v
 emitAtomToName hint x = emitHinted hint (Atom x)
+{-# INLINE emitAtomToName #-}
+
+buildScopedAssumeNoDecls :: (SinkableE e, ScopableBuilder r m)
+  => (forall l. (Emits l, DExt n l) => m l (e l))
+  -> m n (e n)
+buildScopedAssumeNoDecls cont = do
+  buildScoped cont >>= \case
+    (Abs Empty e) -> return e
+    _ -> error "Expected no decl emissions"
+{-# INLINE buildScopedAssumeNoDecls #-}
 
 -- === "Hoisting" top-level builder class ===
 
--- a/src/lib/Inline.hs
+++ b/src/lib/Inline.hs
@@ -4,7 +4,7 @@
 -- license that can be found in the LICENSE file or at
 -- https://developers.google.com/open-source/licenses/bsd
 
-module Inline where
+module Inline (inlineBindings) where
 
 import Data.List.NonEmpty qualified as NE
 
@@ -61,11 +61,11 @@ liftInlineM act = liftBuilder $ runSubstReaderT idSubst $ runInlineM act
 data SizePreservationInfo =
   -- Explicit noinline, or inlining doesn't preserve work
     NoInline
-  -- Used once, ergo size-preserving to inline.  In Secrets, this corresponds to
-  -- either OnceSafe, or OnceUnsafe with whnfOrBot == True
+  -- Used once statically, ergo size-preserving to inline.  In Secrets, this
+  -- corresponds to either OnceSafe, or OnceUnsafe with whnfOrBot == True
   | UsedOnce
-  -- Used more than once, ergo potentially size-increasing to inline.  In
-  -- Secrets, this corresponds to either MultiSafe, or MultiUnsafe with
+  -- Used more than once statically, ergo potentially size-increasing to inline.
+  -- In Secrets, this corresponds to either MultiSafe, or MultiUnsafe with
   -- whnfOrBot == True
   | UsedMulti
   deriving (Eq, Show)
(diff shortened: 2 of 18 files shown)


src/lib/Inline.hs
```diff
@@ -0,0 +1,475 @@
...
+{-# INLINE liftInlineM #-}
+
+-- === Inliner ===
+
+data SizePreservationInfo =
+  -- Explicit noinline, or inlining doesn't preserve work
+    NoInline
+  -- In Secrets, this is either OnceSafe, or OnceUnsafe with whnfOrBot == True
+  | UsedOnce
+  -- In Secrets, this is either MultiSafe, or MultiUnsafe with whnfOrBot == True
```
apaszke, line 66: Can you write a few words to explain what this represents without needing to dive into the Secrets paper?

src/lib/Inline.hs
```diff
@@ -0,0 +1,475 @@
...
+  OccInfoPure _ -> False
+  OccInfoImpure _ -> False
+
+data Context (from::E) (to::E) (o::S) where
+  Stop :: Context e e o
+  TabAppCtx :: NonEmpty (SAtom i) -> Subst InlineSubstVal i o
+            -> Context SExpr e o -> Context SExpr e o
+  EmitToAtomCtx :: Context SAtom e o -> Context SExpr e o
+  EmitToNameCtx :: Context SAtomName e o -> Context SAtom e o
+  RepWrap :: GenericE e1 => Context e1 e2 o -> Context (RepE e1) e2 o
```
apaszke, line 228: It would be helpful to add some comments here, because I don't fully understand what this does. I feel like you've defunctionalized something, but I don't know what that is.

apaszke (approved):
Mostly LGTM, but I was unable to review anything that uses `Context`, since I don't understand how it's supposed to work.

src/lib/Inline.hs
```diff
@@ -0,0 +1,475 @@
...
+  -- Explicit noinline, or inlining doesn't preserve work
+    NoInline
+  -- In Secrets, this is either OnceSafe, or OnceUnsafe with whnfOrBot == True
+  | UsedOnce
+  -- In Secrets, this is either MultiSafe, or MultiUnsafe with whnfOrBot == True
+  | UsedMulti
+  deriving (Eq, Show)
+
+inlineDecls :: Emits o => Nest SDecl i i' -> InlineM i' o a -> InlineM i o a
+inlineDecls decls cont = case decls of
```
apaszke, line 71: BTW if you really care about the performance of this pass, this function should get refactored into a tail recursive one that doesn't call the continuation and only accumulates the substitution, and another INLINE wrapper around it that runs the helper and then calls the continuation. The latter way makes it clear to GHC that the continuation will be called exactly once. You can find this pattern e.g. [in Simplify](https://github.com/google-research/dex-lang/blob/main/src/lib/Simplify.hs#L174-L192).

src/lib/Types/Primitives.hs
```diff
@@ -261,10 +261,19 @@ instance Pretty Arrow where
...
+    PlainLet
+  -- Binding explicitly tagged "do not inline"
+  | NoInlineLet
+  -- Bound expression is pure, and the binding's occurrences are summarized by
+  -- the UsageInfo
+  | OccInfoPure UsageInfo
+  -- Bound expression is impure, and the binding's occurrences are summarized by
+  -- the UsageInfo.  For now, the inliner does not distinguish different effects,
+  -- so no additional information on effects is needed.
+  | OccInfoImpure UsageInfo
```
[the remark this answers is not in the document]
axch: Unfortunately, the type of `getEffects` specializes in context to `Expr o -> InlineM i o (EffectRow o)`, but I want to know whether the RHS is pure _before_ I traverse it.  I suppose I could call `getEffectsSubst`, but at that point, just stashing the answer from the occurrence traversal seemed more reliable.  (In the sense of more reliably avoids traversing any large things.)  Am I wrong about that, and should I just use `getEffectsSubst`?

tests/unit/OccAnalysisSpec.hs
```diff
@@ -335,4 +335,14 @@ spec = do
...
+      ann `shouldBe` OccInfoPure (UsageInfo Two (1, Two))
+    it "is conservative about ix dicts" do
+      ann <- analyze cfg env
+        [ ":p"
+        , "  n : Nat = unreachable ()"
+        , "  xs : (Fin n) => Nat = iota (Fin n)"
+        , "  sum xs"
+        ]
+      -- Statically unbounded because we assume that an ix dict may be
+      -- inlined uncontrollably later.
```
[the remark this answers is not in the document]
axch: It is quite ordinary.  But IIUC, Ix dicts are effectively inlined all over by the rest of the compiler, so the inliner should not inline into them.  In this case, that means taking care not to inline `n`, and not to suffer integer overflow when computing the bound for how much `n` is used.

src/lib/Builder.hs
```diff
@@ -130,10 +131,22 @@ emitDecls' (Nest (Let b (DeclBinding ann _ expr)) rest) e = do
...
 emitAtomToName _ (Var v) = return v
 emitAtomToName hint x = emitHinted hint (Atom x)
 
+buildScopedAssumeNoDecls :: (SinkableE e, ScopableBuilder r m)
+  => (forall l. (Emits l, DExt n l) => m l (e l))
+  -> m n (e n)
+buildScopedAssumeNoDecls cont = do
+  buildScoped cont >>= \case
+    (Abs Empty e) -> return e
+    _ -> error "Expected no decl emissions"
```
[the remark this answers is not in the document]
axch: Done.

src/lib/Inline.hs
```diff
@@ -0,0 +1,475 @@
...
+
+-- === External API ===
+
+inlineBindings :: (EnvReader m) => SBlock n -> m n (SBlock n)
+inlineBindings blk = liftInlineM $ buildScopedAssumeNoDecls $ inline Stop blk
+{-# SCC inlineBindings #-}
+
+-- === Data Structure ===
+
+data SubstRange (o::S) where
```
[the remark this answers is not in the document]
axch: Copy-pasta from how Secrets of the GHC Inliner presented their inliner.  It's the range of the substitution.  Renamed.

src/lib/Inline.hs
axch: Done.

src/lib/Inline.hs
axch: Made an attempt; PTAL.

src/lib/Inline.hs
```diff
@@ -0,0 +1,497 @@
...
+-- which supports now-local optimizations in `reconstruct`.  Read `reconstruct`
+-- together with this data type: `reconstruct` is an interpreter for it.
+--
+-- Note: This pattern of inserting an object into a context can do local
+-- optimizations that would otherwise be hidden from a peephole optimizer by
+-- intervening bindings.  For example, table indexing only permits an Atom in
+-- the array position, but `reconstruct` can check whether it's trying to insert
+-- a `for` expression into that spot and perform the beta reduction immediately,
+-- instead of emitting the binding.
+data Context (from::E) (to::E) (o::S) where
```
dougalm, line 242: This is a really clever trick! I wouldn't have thought of doing it that way.

src/lib/Inline.hs
```diff
@@ -0,0 +1,497 @@
...
+      let (ixsPref, ixsRest) = NE.splitAt bsCount ixs
+      -- Note: There's a decision here.  Is it ok to inline the atoms in
+      -- `ixsPref` into the body `decls`?  If so, should we pre-process them and
+      -- carry them in `DoneEx`, or suspend them in `SuspEx`?  (If not, we can
+      -- emit fresh bindings and use `Rename`.)  This decision affects the
+      -- runtime of the inliner and the code size of the IR the inliner
+      -- produces; since `ixsPref` are atoms, no decision made here can
+      -- duplicate work.
+      -- - Tried going with `SuspEx`, in the interest of "launch and iterate".
+      -- - This creates a bug: what if I'm beta-reducing a body that
```
dougalm, line 327: This sounds like you've exposed a preëxisting bug rather than created a new one. Beta-reducing a `for` expression with a correctly-typed atom shouldn't crash the compiler. We do it all over the place. In particular, `j:(..<some_atom>)` should be perfectly fine. I'm curious what the error message was, if you can remember it.

dougalm (approved):
Really nice work!

src/lib/Inline.hs
axch: On second thought, this is more subtle.  Each `ixsPref` is an atom, but there may a pending inlining that will try to turn it into an expression (that may do work).  So my original tri-lemma is not actually valid at all; rewrote the comment.

src/lib/Inline.hs
axch: Thank you!  It did take some time to work out.

src/lib/Inline.hs
```diff
@@ -0,0 +1,497 @@
...
+liftInlineM :: (EnvReader m) => InlineM n n a -> m n a
+liftInlineM act = liftBuilder $ runSubstReaderT idSubst $ runInlineM act
+{-# INLINE liftInlineM #-}
+
+-- === Inliner ===
+
+data SizePreservationInfo =
+  -- Explicit noinline, or inlining doesn't preserve work
+    NoInline
+  -- Used once, ergo size-preserving to inline.  In Secrets, this corresponds to
```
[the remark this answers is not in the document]
axch: Yes.  Fixed.

src/lib/Inline.hs
```diff
@@ -0,0 +1,497 @@
+-- Copyright 2022 Google LLC
+--
+-- Use of this source code is governed by a BSD-style
+-- license that can be found in the LICENSE file or at
+-- https://developers.google.com/open-source/licenses/bsd
+
+module Inline where
```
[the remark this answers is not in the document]
axch: Right.  Done.

PR #1172, opened 2022-11-29
Add the molecular dynamics demo as an end-to-end benchmark.

axch:
This required tweaking continuous.py to set the LANG environment variable (to en_US.UTF-8), because `hGetContents` reads it to determine how to decode the bytes.  Molecular dynamics triggers this because it uses some Greek letters as variable names.


apaszke:
Is there something we could do to always interpret Dex sources as UTF-8, no matter what encoding is selected?

axch:
Done.

dougalm (approved):

PR #1181, opened 2022-12-15
Generalize `StringSetIx` to any element type, and rename it `Element`.

axch:
Usage example:

```
people = to_set <a bunch of strings>
Person = Element String people
for i:Person.
  ...
```

At runtime, `Person` is an integer indexing into the array backing the set.  You can compare them for equality and ordering, and use them as indices into other arrays (TODO and hash them) without having to mess with the actual strings; but the strings are there if you ever need them.

Tagline: The code book is in the type.

### DIFF
--- a/lib/set.dx
+++ b/lib/set.dx
@@ -79,37 +79,33 @@ def set_intersect {a}
   UnsafeAsSet _ intersection
 
 
-'## Index set for sets of strings
+'## Sets as a type, whose inhabitants can index arrays
 
--- Todo: Make polymorphic in type.  Waiting on a bugfix.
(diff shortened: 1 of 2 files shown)


axch:
@duvenaud, you might be interested in this.

dougalm (approved):
Nice! IIRC, David originally tried to parameterize it by an arbitrary type rather than just `String` but he ran into Dex bugs. I wonder if the bugs have been fixed now or if he was just trying a different approach to yours.

duvenaud:
Yep, I was trying the same approach before, and we were looking at it again now that #1123 was fixed.

PR #1182, opened 2022-12-16
Split up primops into groups so they can be allowed/denied at the group level

dougalm:
I initially tried giving IR predicates to each primop individually. Unfortunately, that means we have to pattern-match on every allowable primop when we do an IR-changing transformation, so that GHC knows that you've handled all the required cases. Previously we could have a fallback case, but that no longer typechecks. With the grouped design, you still have to pattern-match on each top-level case, but you can handle entire subtrees generically.

### DIFF
--- a/src/lib/Lower.hs
+++ b/src/lib/Lower.hs
@@ -27,6 +27,7 @@ import Types.Primitives
 
 import Err
 import Name
+import IRVariants
 import MTL1
 import Core
 import Builder
@@ -89,7 +90,7 @@ instance HoistableState LFS where
 -- annotations manually... The only interesting cases below are really For and TabCon.
 instance GenericTraverser SimpIR UnitB LFS where
   traverseExpr expr = case expr of
-    Op (TabCon ty els) -> traverseTabCon Nothing ty els
+    TabCon ty els -> traverseTabCon Nothing ty els
     Hof (For dir ixDict body) -> traverseFor Nothing dir ixDict body
     Case _ _ _ _ -> do
       Case e alts ty _ <- traverseExprDefault expr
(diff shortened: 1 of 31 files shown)


src/lib/Lower.hs
```diff
@@ -426,22 +412,37 @@ vectorizeOp op = do
...
+    --   VVal Contiguous <$> emitOp (VectorSubref ref i vty)
+    -- Place (VVal vref ref) sval@(VVal vval val) -> do
+    --   VVal Uniform <$> case (vref, vval) of
+    --     (Uniform   , Uniform   ) -> emitOp $ Place ref val
+    --     (Uniform   , _         ) -> error "Write conflict? This should never happen!"
+    --     (Varying   , _         ) -> error "Vector scatter not implemented"
+    --     (Contiguous, Varying   ) -> emitOp $ Place ref val
+    --     (Contiguous, Contiguous) -> emitOp . Place ref =<< ensureVarying sval
+    --     _ -> error "Not implemented yet"
+
```
axch, line 445: Why commented out?

axch (approved):

src/lib/Simplify.hs
```diff
@@ -194,13 +195,22 @@ simplifyExpr hint expr = confuseGHC >>= \_ -> case expr of
...
     simplifyTabApp hint f xs'
   Atom x -> simplifyAtom x
-  -- Handle PrimEffect here, to avoid traversing cb multiple times
-  Op  op  -> (inline traversePrimOp) simplifyAtom op >>= simplifyOp
-  PrimEffect ref eff -> do
+  -- Handle RefOp here, to avoid traversing cb multiple times
+  PrimOp  op  -> (inline traversePrimOp) simplifyAtom op >>= simplifyOp
+  ProjMethod dict i -> bindM2 projectDictMethod (simplifyAtom dict) (pure i)
+  -- Those are not no-ops! Builder methods do algebraic simplification!
+  -- TODO: Move peephole optimizations to peepholeOp in Optimize
```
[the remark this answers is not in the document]
dougalm: yep, soggy pasta

src/lib/Lower.hs
dougalm: Good catch. It's a refactoring artifact, but I didn't notice it because vectorize is dead (until Norman's patch revives it).

PR #1194, opened 2022-12-25
Implement constant folding for casts to floating-point types.

axch:
Caveat: Did not check that they agree bit-for-bit with the runtime behavior, but any discrepancies should be small.

Benefit:

- This ends up constant-folding the cast from the literal zero to the relevant floating-point type in the `Add` instance for floats.

- Therefore, that zero ends up being an `Atom` and not an expression when we build the `BaseMonoid` for adding floats, and so is inlined therein.

- Therefore, that zero is, in effect, rematerialized when doing AD.

- Therefore, an n by m by k tensor of those zeros is not stored by linearize of matmul, so the jvp-matmul benchmark runs some 30% faster.  Of course, it is still storing two other tensors it shouldn't store, but that's for another time.


apaszke (commented):
To be fair, not verifying that this matches the LLVM implementation is asking for very subtle bugs in the future IMO. I know that doing full testing on FP (especially 64-bit) is hard, but perhaps we could test the simple cases and limit constant folding to those? For example, you could disallow any numbers unrepresentable in the destination type.

axch:
Added some measure of consistency testing between run-time and compile-time casts.  Unfortunately, it's hard to test many values, because compiling and running a million 1-instruction LLVM programs is much more expensive than executing a million LLVM instructions.  That said, the tests did reveal two bugs (one already present and one introduced by me in the previous commit), which fixed.

tests/unit/ConstantCastingSpec.hs
```diff
@@ -0,0 +1,99 @@
...
+                let op = castOp ty x
+                evaled <- evalClosedExpr $ PrimOp op
+                -- The failure message will list `evaled` as "expected" and `folded` as "got"
+                folded `shouldBe` evaled
+    forM_ [Int64Type, Int32Type, Word8Type, Word32Type, Word64Type, Float64Type, Float32Type] \ty ->
+      -- TODO: We'd really like to run 10,000 or 1,000,000 examples here, but
+      -- status quo is that each one runs through the LLVM compile-and-run
+      -- pipeline separately, and is thus incredibly slow.  Taking 50 as a
+      -- compromise between test suite speed and test coverage.
+      -- I ran this offline with 10,000 before checking in, and it passed.
```
apaszke, line 76: To be fair, we do have everything we need to compile the program that does the rounding once and evaluate it repeatedly on different inputs. But we can leave it as is for now.

tests/unit/ConstantCastingSpec.hs
```diff
@@ -0,0 +1,99 @@
...
+      modifyMaxSuccess (const 50) $ prop ("agrees with runtime when casting to " ++ show ty) $ constant_folding_and_runtime_casts_agree ty
+    it "agrees with runtime on rounding mode" $
+      -- These values are chosen to tickle the difference between different
+      -- rounding modes when rounding to float32, and specifically between
+      -- breaking ties to even, to zero, or to +infinity when rounding to
+      -- nearest.
```
apaszke, line 82: What are those values?

src/lib/ImpToLLVM.hs
```diff
@@ -451,26 +451,42 @@ compileInstr instr = case instr of
...
-       (L.FloatingPointType fpt, L.FloatingPointType fpt') -> case compare fpt fpt' of
-         LT -> emitInstr dt $ L.FPExt x dt []
-         EQ -> return x
-         GT -> emitInstr dt $ L.FPTrunc x dt []
-       (L.FloatingPointType _, L.IntegerType _) -> emitInstr dt $ L.FPToSI x dt []
-       (L.IntegerType _, L.FloatingPointType _) -> emitInstr dt $ L.SIToFP x dt []
+      _ -> case (getIType ix, sdt) of
+        -- if upcasting from unsigned int, use zext instruction
+        (Scalar Word32Type, L.IntegerType bits) | bits > 32 -> x `zeroExtendTo` dt
+        (Scalar Word8Type,  L.IntegerType bits) | bits > 8  -> x `zeroExtendTo` dt
```
apaszke, line 457: Those seem new? Do we have any tests for them?

apaszke (approved):

tests/unit/ConstantCastingSpec.hs
axch: Ooo, how?

tests/unit/ConstantCastingSpec.hs
axch: Improved the comment.

src/lib/ImpToLLVM.hs
axch: Yes; the QuickCheck property will sometimes pick `Word8Lit` inputs to cast to `Int32Type` or `Int64Type`, and thus exercise this code path.

tests/unit/ConstantCastingSpec.hs
apaszke: If you look at the body of [`evalLLVM`](https://github.com/google-research/dex-lang/blob/main/src/lib/TopLevel.hs#L773), the majority of it leads to generating `nativeFun`, which is then only called once and discarded. What you could do, is to compile a `Float -> Float` function and then use `callNativeFun` repeatedly to evaluate it over a range of inputs.

PR #1197, opened 2023-01-02
Allow SubstE to access Env

dougalm:
This allows us to normalize properly as we substitute. Previously, when we encountered redexes like `ProjectElt 0 (ProdVal [...])` we were limited to purely syntactic normalization (e.g. no type querying). That lead to hacks like "deferred repval" which bundled a repval with a bunch of deferred projections. And it was a blocker for my newtype-stripping patch (#1188), because I needed to embed *two* lists of projections in `SimpInCore`.

To check that it works I removed `DRepVal` and changed `ProjectElt` to be unary and take an atom (rather than a var) as its RHS. Note that this means we're no longer syntactically enforcing normalization. But we were always going to have to give that up once we started allowing more reducible-upon-substitution terms in our types, like table indexing and function application. The plan is to maintain normalization by (1) normalizing within substitution and (2) being careful about using `ProjectElt` and similar, preferring the normalizing builders versions instead. At some point I'd like to distinguish terms in type position using the IR variant system. Then we might be able to enforce normalization (and drop ANF!) on those terms specifically.

### DIFF
--- a/src/lib/Builder.hs
+++ b/src/lib/Builder.hs
@@ -1304,7 +1304,7 @@ getUnpacked atom = do
   atom' <- cheapNormalize atom
   ty <- getType atom'
   idxs <- projectionIndices ty
-  return $ idxs <&> flip getProjection atom'
+  forM idxs \is -> normalizeNaryProj is atom'
 {-# SCC getUnpacked #-}
 
 emitUnpacked :: (Builder r m, Emits n) => Atom r n -> m n [AtomName r n]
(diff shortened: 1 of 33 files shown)


src/lib/Builder.hs
```diff
@@ -1304,20 +1304,20 @@ getUnpacked atom = do
   atom' <- cheapNormalize atom
   ty <- getType atom'
   idxs <- projectionIndices ty
-  return $ idxs <&> flip getProjection atom'
+  forM idxs \is -> normalizeNaryProj is atom'
```
axch, line 1307: Nit: Does this invalidate the comment above this function?

dougalm: I think it's still correct because `normalizeNaryProj` assumes that `atom` itself is already normalized.

dougalm:
@axch , thanks for review. Made follow-up PR here: #1200

PR #1198, opened 2023-01-02
Add a HeapType type and HeapVal value for the heap parameter.

dougalm:
Previously we just used `h:Type` (leftover from ST monad inspiration) and `UnitTy` as its "runtime" value. But that meant SimpIR had to contain vars of type Type, which we want to rule out statically.

### DIFF
--- a/src/lib/Inference.hs
+++ b/src/lib/Inference.hs
@@ -99,7 +99,8 @@ inferTopUDecl (UInstance className bs params methods maybeName) result = do
   -- this.
   Abs Empty ab <- liftInfererM $ solveLocal $ buildDeclsInf do
     checkInstanceArgs bs do
-      checkInstanceParams params \params' -> do
+      ClassDef _ _ paramBinders _ _ <- getClassDef (sink className')
+      checkInstanceParams paramBinders params \params' -> do
         className'' <- sinkM className'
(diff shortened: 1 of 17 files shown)


src/lib/Inference.hs
```diff
@@ -1943,21 +1944,28 @@ checkHandlerBodyTyArg b cont = do
...
+    go Empty []    ptys = Abs Empty <$> cont (reverse ptys)
+    go (Nest (RolePiBinder b k _ _) bs) (p:t) ptys = do
+      checkParamWithMissingDicts k p \ds getParamType -> do
+        mergeAbs <$> introDicts (eSetToList ds) do
+          pty <- getParamType
+          bsAbs <- sinkM $ Abs b (EmptyAbs bs)
+          Abs bs' UnitE <- applyAbs bsAbs (SubstVal pty)
+          ListE ptys' <- sinkM $ ListE ptys
+          go bs' t (pty:ptys')
+    go _ _ _ = error "zip error"
```
axch, line 1968: What are these changes changing?  What are we reading from the class definition now and why?

dougalm: They're changing because we can no longer assume that all class parameters have type `Type`. (Specifically, they might have now have type `Heap`.)

PR #1199, opened 2023-01-02
Effectful stack with O(1) push/pop [WIP]

dougalm:


axch (commented):
Looks good!

Can we add a test that there isn't a hidden quadratic somewhere?  Something like do 1,000,000 single-element stack pushes and check (implicitly) that the test doesn't time out?

PR #1202, opened 2023-01-04
Add a built-in "print absolutely anything" function

dougalm:
The goal is to make this the default printing path and remove Interepreter/Serialize to unblock #1188. It'll also let us debug-print any dex value at runtime. There's still work to do in adding all the cases but I think it's worth merging this for now.

### DIFF
--- a/src/lib/Transpose.hs
+++ b/src/lib/Transpose.hs
@@ -256,6 +256,7 @@ transposeMiscOp op _ = case op of
   BitcastOp    _ _      -> notImplemented
   UnsafeCoerce _ _      -> notImplemented
   GarbageVal _          -> notImplemented
+  ShowAny _ -> error "Shouldn't have ShowAny in simplified IR"
   where
     notLinear = error $ "Can't transpose a non-linear operation: " ++ show op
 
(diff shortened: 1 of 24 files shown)


src/lib/Transpose.hs
```diff
@@ -256,6 +256,8 @@ transposeMiscOp op _ = case op of
   BitcastOp    _ _      -> notImplemented
   UnsafeCoerce _ _      -> notImplemented
   GarbageVal _          -> notImplemented
+  ShowAny _ -> error "Shouldn't have ShowAny in simplified IR"
+  ShowScalar _ -> error "Shouldn't have ShowScalar in simplified IR"
```
axch, line 260: We can't use IR tags to prove this doesn't happen?  Seems like an obvious use case for them...

axch (approved):
So, is generating code to print every possible type of value, inline, going to lead to bloated IRs and compile times?

Though, I don't suppose it would be any better to implement printing in Dex with classes, because I suppose it would amount to more or less the same thing.  Though if we marked the show function noinline, at least there would only be one copy of it per type, instead of per occurrence.

dougalm:
> So, is generating code to print every possible type of value, inline, going to lead to bloated IRs and compile times?

We'll see. My hope is that the sorts of values we like to print have relatively small types. If it's an issue, the next step would be to add `@noinline` as you say.

src/lib/Transpose.hs
dougalm: Yes, absolutely. That will happen as part of #1188 .

PR #1204, opened 2023-01-05
Make codegen/runtime printing the default

dougalm:
Unsurprisingly, this changes the quine tests a lot. Here are the main changes:

  * ~Floats no longer print with a trailing decimal point. I couldn't find a
    format string that would print the decimal point without also printing
    trailing zeros. But I actually think it's an improvement this way. The type
    can be queried separately so it's redundant for it to also appear in the
    printed representation. And it's actively distracting when you see the same
    type printed explicitly in every element of a table. Type uniformity means
    our textual representations can be more information-dense.~
  * Floating point printed precision is a bit different. We're using C `g`
    format now rather than whatever Haskell uses.
  * Non-Fin tables no longer print with their type. Again, I think that's fine
    because the type information is redundant. Same story with `0@(Fin 10)`.
    We now just print `0`.
  * No automatic line breaks. I didn't want to implement the layout algorithm
    from `Data.Text.PrettyPrint` in the dex AST builder...
  * We now print lambda terms as just `<function of type : Int -> Float>` etc
    instead of printing their bodies. This is what most languages do anyway.
    Seeing the lambda explicitly is sometimes useful but there's always `:pp`
    for that.
  * Some extra parentheses here and there. I'm not implementing precedence-aware
    printing in the AST builder.
  * I decided to print standalone Word8 using hex. Before we printed as the
    ASCII char itself if it was in the printable range, and some Haskell-defined
    escape code otherwise. We still want that hack for string literals. But it
    seems a bit much for standalone Word8.
  * `unsafe_io` gets DCEd in some of the io tests because it's now an argument
    to `show_any` but it doesn't produce any non-trivial data. I tweaked the
    tests but we should revisit this.


axch (commented):
Lack of line breaks registers as a regression, but perhaps an acceptable one in the interest of reducing implementation cost.

Losing the decimal point feels like a real downer, though, because I'm so used to it being the difference between a civilized programming language and the barbarians at the gates.  I asked GPT how to keep it, but GPT confidently gave me the wrong answer of using `%.1f` as a format string, which prints the trailing zero.  Actually, for my part, I would rather have the trailing zero than nothing, but I'm not sure offhand how to keep it without limiting the display precision.

How feasible would it be to codegen a test that says "if the value of this float was actually integral, add a trailing ."?  This could screw up if the underlying print decided to produce a . in some cases, but maybe it's predictable enough?

dougalm:
Ok, I added the trailing decimal point. I'll merge once the tests pass.

PR #1210, opened 2023-02-05
Rewrite general_integer_power with a `for` because Dex can't differentiate through `while`

### LINKED ISSUE #1195: "Compiler bug, Unsupported newtype: Nat" when taking grad through `intpow`
I came across this when working on a gradient-based optimization example, using a function that called `intpow` as a test case.

```
def f (x:Float) : Float =
  intpow x 2

x = 3.1
f x
> 9.61
grad f x
> Compiler bug!
>
> Unsupported newtype: Nat
>   error, called at src/lib/Linearize.hs:515:10 in dex-0.1.0.0-56tlNzHvZd2Epf2TOu8iN7:Linearize
```

axch:
There is still a `while` in there, to compute the number of iterations of the squaring loop, but
- It doesn't touch the floating-point input, so can't mess up AD, and
- We should replace it with a count-leading-zeros intrinsic anyway.

Differentiating through general `while` is a problem because the tape becomes a recursive ADT (namely List), and Dex doesn't support those yet.

Fixes #1195.

### DIFF
--- a/lib/prelude.dx
+++ b/lib/prelude.dx
@@ -2281,6 +2281,12 @@ def is_power_of_2 (x:Nat) : Bool =
     then False
     else 0 == %and x' (%isub x' (1::NatRep))
 
+-- This computes the integer part of the binary logarithm of the input.
+-- TODO: natlog2 0 should do something other than underflow the answer.
+-- TODO: Use LLVM ctlz intrinsic instead.  It needs a slightly new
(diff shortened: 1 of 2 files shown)


dougalm (approved):
A custom linearization might be appropriate too. But that would probably require the multiplication function to be passed as a class dictionary so we can expect it to follow particular rules. To differentiate `a^b` as `(b * a ^ (b-1))` is it enough to know that `*` is bilinear and it distributes over addition?

PR #1219, opened 2023-02-07
Handle at least some non-data scrutinees in `case`.

### LINKED ISSUE #1217: Can't handle toplevel case with a functional value?
```
@noinline
def id' (x:Bool) : Bool = x

case (id' True) of
  True -> Just f64_to_f
  False -> Nothing
> Compiler bug!
>
> Not a data atom: (acase (ProjectElt 1 v#1) [(\v#4:Unit.(0| () |)), (\v#4:Unit.(1| f64_to_ |))])
>   error, called at src/lib/Simplify.hs:184:14 in dex-0.1.0.0-Gi9B8TxRbicLQnuTnmjlXJ:Simplify
```
Expected
```
> (Just <function of type (Float64 -> Float32)>)
```

(The `id'` is needed to defeat case-of-known-constructor optimization, which masks the bug).

axch:
Fixes #1217 as filed, but there may be variants I didn't test that still sneak by.

### DIFF
--- a/src/lib/Simplify.hs
+++ b/src/lib/Simplify.hs
@@ -337,16 +337,11 @@ simplifyExpr hint expr = confuseGHC >>= \_ -> case expr of
     liftSimpAtom (Just ty') =<< emitExpr (TabCon tySimp xs')
   UserEffectOp _ -> error "not implemented"
   Case scrut alts resultTy _ -> do
-    -- TODO: this can fail! We need to handle the case of a non-data scrutinee!
     scrut' <- simplifyAtom scrut
-    altBinderTys <- caseAltsBinderTys =<< getType scrut'
-    scrutSimp <- toDataAtomIgnoreRecon scrut'
     resultTy' <- substM resultTy
-    defuncCase scrutSimp resultTy' \i x -> do
+    defuncCaseCore scrut' resultTy' \i x -> do
       Abs b body <- return $ alts !! i
-      let xCoreTy = altBinderTys !! i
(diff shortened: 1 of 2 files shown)


src/lib/Simplify.hs
```diff
@@ -368,8 +363,33 @@ caseComputingEffs scrut alts resultTy = do
...
+      Nothing -> case scrut of
+        SimpInCore (ACase scrutSimp alts _) -> do
+          defuncCase scrutSimp resultTy \i x -> do
+            Abs altb altAtom <- return $ alts !! i
+            altAtom' <- applySubst (altb @> SubstVal x) altAtom
+            cont i altAtom'
+        -- TODO(reviewer): Which cases of Atom can actually appear here, in a
+        -- well-typed source program?  I'd like the error to distinguish between
+        -- "I know this is impossible for reason X" from "This could happen,
+        -- but we just didn't program what to do."
```
dougalm, line 389: I think `ACase` might be the only one. It has to be a sum type with at least one non-data case. True CoreIR sum types are represented as either explicit constant constructors, which you handle above, or variables, which we can't have here because we only have SimpIR variables (hmm... maybe top-level CoreIR variables could still exist. I want to remove that possibility separately from this PR so let's ignore that for now). That leaves the embedded `SimpInCore` cases, and of those, `ACase` is the only one that could represent a sum type with non-data cases.

dougalm (approved):

src/lib/Simplify.hs
axch: OK.

PR #1227, opened 2023-02-09
Use flat products instead of nested pairs in non-dependent telescopes.

dougalm:
Partly this is just to make the IR more readable. But it might have some compile-time performance advantages too. Linearization can produce intermediate telescopes with hundreds of components. Indexing these as nested pairs means an "atom" can have hundreds `ProjectElt` layers. I wouldn't be surprised if some operations like type checking end up being quadratic in that depth.

### DIFF
--- a/src/lib/Builder.hs
+++ b/src/lib/Builder.hs
@@ -1528,7 +1528,52 @@ runMaybeWhile body = do
 
 -- === capturing closures with telescopes ===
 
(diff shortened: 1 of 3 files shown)


dougalm:
Running DCE after linearization reduced the amount of code that we produced, but it exposed a strange bug in AD: https://github.com/google-research/dex-lang/tree/flat-telescopes-dce-ad-bug. Reverting for now.

PR #1232, opened 2023-02-11
Remove "allowed effects" from the Env we carry everywhere.

dougalm:
We only actually use the allowd effects in two places: inference and type checking. Tracking them everywhere else hurts performance because we have to consider extending the allowed effects every time we emit a binding, which is a lot. Since we added the IR parameters, the effects carried in the Env are also ill-typed, because they're treated as CoreIR even though we can have SimpIR effects too. We had to use unsafeCoerceIRE to work around that. After this change, the only use of unsafeCoerceIRE is in pretty-printing.


src/lib/TopLevel.hs
```diff
@@ -731,13 +731,7 @@ checkPass name cont = do
...
 #ifdef DEX_DEBUG
   logTop $ MiscLog $ "Running checks"
-  let allowedEffs = case name of
-                      LowerPass    -> OneEffect InitEffect
-                      LowerOptPass -> OneEffect InitEffect
-                      VectPass     -> OneEffect InitEffect
-                      _            -> mempty
-  {-# SCC afterPassTypecheck #-} (liftExcept =<<) $ liftEnvReaderT $
-    withAllowedEffects @CoreIR allowedEffs $ checkTypesM result
+  checkTypesM result
```
axch, line 734: Where do we now check that top-level forms do not have impermissible effects?

axch (changes requested):
Looks good, but could you add some explanation of the API change of checkE?  Why did it used to return the checked E-kinded object, and why is it ok for it to now return unit?

dougalm:
> Looks good, but could you add some explanation of the API change of checkE? Why did it used to return the checked E-kinded object, and why is it ok for it to now return unit?

It was just a performance consideration in the first place. The idea was that since we often have to apply the substitution as we check the term, we should be able to return the substituted term to save the caller from possibly repeating that work. But it turned out that we only rarely used this feature. The main user was in checking blocks, and we have a separate path for that now anyway. The other thing that's changed is that nowadays we only run the checker in debug mode, so performance isn't as critical.

src/lib/TopLevel.hs
dougalm: We don't. Now that we have to pass the effects around explicitly, it's a bit more awkward. But I can add some checks back.

PR #1251, opened 2023-03-03
Use concrete/grouped syntax in SourceBlock instead of abstract syntax

dougalm:
The abstract syntax isn't needed at the top level. Meanwhile, the grouped representation is closer to the actual source text, which makes it the best place to extract the spanning trees, which we will need at the top level.

### DIFF
--- a/src/Dex/Foreign/Context.hs
+++ b/src/Dex/Foreign/Context.hs
@@ -29,6 +29,7 @@ import Data.ByteString    qualified as BS
 import Data.Text.Encoding qualified as T
 
 import AbstractSyntax
+import ConcreteSyntax
 import Builder
 import Core
 import Err
--- a/src/dex.hs
+++ b/src/dex.hs
@@ -26,8 +26,8 @@ import PPrint (toJSONStr, printResult)
 import TopLevel
 import Err
 import Name
-import ConcreteSyntax (keyWordStrs)
-import AbstractSyntax (parseTopDeclRepl, preludeImportBlock)
+import AbstractSyntax (parseTopDeclRepl)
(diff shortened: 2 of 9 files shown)


dan-zheng (commented):
Just checking my understanding of the main changes:

- `SourceBlock` has been simplified. The `sbContents` field now has a fixed currency type `SourceBlock'` instead of either `SourceBlock'` or `CSourceBlock'` (due to being a type variable `a` from `data SourceBlockP a = ...`).

Before:
```hs
-- Types/Source.hs:
data SourceBlockP a = SourceBlockP
  { ...
    , sbContents :: a }

type SourceBlock = SourceBlockP SourceBlock'

-- ConcreteSyntax.hs:
type CSourceBlock = SourceBlockP CSourceBlock'
-- ConcreteSyntax functions operate on `CSourceBlock`.
```

After:
```hs
-- Types/Source.hs:
data SourceBlock = SourceBlock
  { sbLine     :: Int
  , sbOffset   :: Int
  , sbLogLevel :: LogLevel
  , sbText     :: Text
  , sbContents :: SourceBlock' }

-- ConcreteSyntax.hs now operates on `SourceBlock'` directly.
-- TopLevel.hs continues to operates on `SourceBlock'`.
```

- It's possible to use `sbContents :: SourceBlock'` both in ConcreteSyntax.hs and TopLevel.hs, when before this was not possible due to separate `SourceBlock'` and `CSourceBlock'` types.
- This unblocks on my work on Dex tooltips! (I'll document our effort there in the coming month)

dan-zheng (approved):
Excellent, thanks! :)

PR #1252, opened 2023-03-07
[fix] unbreak the nix code and make it slightly nicer

MangoIV:
## Fixes 
- [x] `dex` builds
- [x] `dex-cuda` builds
- [x] `dex` tests pass
- [ ] `dex-cuda` tests pass (see *Warning*, I don't have an nvidia graphics card)

## Additions 
- nix formatter can be run with `nix fmt` 
- replaced old `devShell` and `package` argument with `<output>.<system>.default`
- don't build the `dex` runtime in the `configurePhase` but instead build it as a 
  separate `drv` 
- made a couple of things more idiomatic

--- 

> **Note**
> the packages of `llvm-hs`, `llvm-hs-pure` and the `dex` package should probably
> be materalized, I'll perhaps at this as a PR at a later point 
> 
> at some point we should probably make the legacy nix code properly follow the flake inputs 
> I didn't update this but as `nixpkgs` is not pinned anyways, I don't think it makes sense 
> to half-ass it. 

---

> **Warning** 
> before merging this, someone with an nvidia graphics card should probably check whether 
> this builds by running `nix build github:mangoiv/dex-lang#dex-cuda` on their system with 
> an nvidia graphics card and ld library path properly set
> 
> be aware that `dex` does not build with current `nixpkgs-unstable` as `floating-bits` is 
> broken for a while, similarly to `llvm-hs/*`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,4 @@ julia/deps/build.log
 examples/t10k-images-idx3-ubyte
 examples/t10k-labels-idx1-ubyte
 examples/camera.ppm
+result*
--- a/default.nix
+++ b/default.nix
@@ -1,72 +1,66 @@
-{ pkgs ? import <nixpkgs> {},
-  llvm-hs-src ? pkgs.fetchFromGitHub {
+{ pkgs ? import <nixpkgs> { }
+, llvm-hs-src ? pkgs.fetchFromGitHub {
     owner = "llvm-hs";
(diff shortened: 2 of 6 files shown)


MangoIV:
I don't have a personal google account and I won't create one; I will contribute under usual terms but if that's not possible I perhaps cannot.

PR #1253, opened 2023-03-08
Fast(er) matmul

axch:
Hand-tiling matrix multiply in Dex speeds it up from ~100ms to ~5.5ms on my laptop (for a 500x500x500 dense multiplication).

There are some caveats, though:
- The tile sizes are just arbitrary numbers, and there is no tuning to different hardware.
- The hand-tiled version relies on Writer to construct the output in place, which defeats output fusion.  (But then again, I don't know that the previous implementation would have fused well on the output either.)
- Adding `@noinline` to this nerfs its performance back to ~40ms, presumably because it defeats LLVM's vectorizer (but I'm not sure why it does that).


apaszke (approved):
LGTM, though I would like us to think more about the index sets useful for tiling at some point. You might also want to do two-level tiling, with the innermost one using numbers that are somewhat architecture inspired (see e.g. [this really nice post](https://gist.github.com/nadavrot/5b35d44e8ba3dd718e595e40184d03f0)).

PR #1258, opened 2023-03-29
QuickCheck found where my `fixUlp` wasn't rounding to nearest even correctly.

axch:
More elaboration fixes it.


dougalm (approved):
nice!

PR #1269, opened 2023-04-04
Update the syntax-philosophy document to the new syntax.

axch:
Also tidy up some loose ends (e.g., parsed but undefined operators) left by the migration.


dougalm (approved):
Hooray for deleting docs! (Even better than deleting code.)

PR #1289, opened 2023-04-28
Deprecate GenericTraversal in favor of TraversableTerm

dougalm:


axch:
Wow, this is turning out to be much nicer!

PR #1290, opened 2023-05-01
Edited “an table” -> “a table” in tutorial.html

jagtesh:
Edited: `an table` -> `a table` in tutorial.html file.


axch:
Thanks for the find!  But, that html file is actually auto-generated from examples/tutorial.dx.  I took the liberty of updating the source in PR #1291, so I'm going to close this PR.

PR #1294, opened 2023-05-15
Case of case and other optimizations in the inliner

axch:
Augment the inliner to do several standard optimizations as it goes:
- Case-of-case
- Case-of-known-constructor
- The existing optimizations in `peepholeExpr`

Doing this in-pass in the inliner is generally recommended because these optimizations can chain with each other, and are basically always beneficial, so getting as many of them done per traversal as possible is desirable.


dougalm (approved):
Looks good!

PR #1295, opened 2023-05-16
Actually code the custom linearization for matmul

axch:
so that we don't embarrass ourselves too much with how poorly AD does thereon.


dougalm (approved):
nice!

PR #1323, opened 2023-07-11
Make a `BinderAndDecls` data type, at first just a wrapper around `Binder`.

dougalm:
Next we'll actually make decls part of it. But that will involve substantive changes. This first step just makes the type distinct from `Binder` so we can make all the administrative changes while still passing tests.

### DIFF
--- a/src/lib/Vectorize.hs
+++ b/src/lib/Vectorize.hs
@@ -26,7 +26,6 @@ import Subst
 import PPrint
 import QueryType
 import Types.Core
-import Types.OpNames qualified as P
 import Types.Primitives
 import Util (allM, zipWithZ)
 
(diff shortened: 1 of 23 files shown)


src/lib/Vectorize.hs
```diff
@@ -257,22 +256,12 @@ monoidCommutativity monoid = case isAdditionMonoid monoid of
...
-  unless (op `elem` [P.IAdd, P.FAdd]) Nothing
-  case (binderName b1, atomVarName b1', binderName b2, atomVarName b2') of
-    -- Checking the raw names here because (i) I don't know how to convince the
-    -- name system to let me check the well-typed names (which is because b2
-    -- might shadow b1), and (ii) there are so few patterns that I can just
-    -- enumerate them.
-    (UnsafeMakeName n1, UnsafeMakeName n1', UnsafeMakeName n2, UnsafeMakeName n2') -> do
-      when (n1 == n2) Nothing
-      unless ((n1 == n1' && n2 == n2') || (n1 == n2' && n2 == n1')) Nothing
-      Just ()
```
axch, line 275: That's true, but is there actually a reason to delete it?  Without a check like this, the vectorizer will cheerfully reorder operations in arbitrary monoids, which may in principle emit incorrect code from a correct program.

PR #1335, opened 2023-08-22
Add multisets

duvenaud:
This is another step towards writing a tokenizer demo in Dex.  Multisets let us find all the unique words in a corpus along with their counts in parallel.


duvenaud-anthro:
Just noting that the Dex tests all pass now.  Something is broken with the jax python tests independent of this PR.

PR #1337, opened 2023-09-25
Add Futhark IR module.

athas:
Not used for anything, but compiled as part of the Dex compiler.


athas:
@apaszke Is the failure here related to my PR?

PR #1343, opened 2023-12-31
Fix debug build by adding a missing conditional import

ararslan:
`src/lib/Err.hs` conditionally uses `unsafePerfomIO` when `DEX_DEBUG` is defined, but it doesn't `import System.IO.Unsafe`, so `unsafePerformIO` is not defined. This can be fixed by importing that module when `DEX_DEBUG` is defined.

### DIFF
--- a/src/lib/Err.hs
+++ b/src/lib/Err.hs
@@ -32,6 +32,10 @@ import Data.Store (Store (..))
 import GHC.Stack
 import GHC.Generics
 
+#ifdef DEX_DEBUG
+import System.IO.Unsafe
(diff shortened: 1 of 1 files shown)


ararslan:
I keep getting a server error every time I try to sign the CLA. 🤔 Perhaps it'll work eventually.

ararslan:
@dougalm, I think this PR will fix CI, which is currently red on `main`. I believe CI is failing on `main` because setting `DEX_CI` enables `DEX_DEBUG`, which is broken without this change.

PR #1347, opened 2024-04-28
Got FFT working again with new syntax.

duvenaud:
I like the new web interface!   But it was a little sad adding extra type annotations.  I was lucky to realize that an especially hairy type on line 126 could be elided by using `each`.  But it's not clear to me why `each` is able to infer the type of the index to loop over when `for` isn't.

### DIFF
--- a/lib/fft.dx
+++ b/lib/fft.dx
@@ -11,7 +11,7 @@ import complex
 
 '## Helper functions
 
-def odd_sized_palindrome(mid:a, seq:n=>a) -> ((n `Either` () `Either` n)=>a) given (a, n|Ix) =
+def odd_sized_palindrome(mid:a, seq:n=>a) -> ((n `Either` () `Either` n)=>a) given (a:Type, n|Ix) =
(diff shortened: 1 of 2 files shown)


dougalm:
Nice!

> But it was a little sad adding extra type annotations. 

Sorry about that. Right now I'm trying to simplify the implementation above all else, which means cutting features.

> But it's not clear to me why each is able to infer the type of the index to loop over when for isn't.

I can explain this one. The new type inference is forward-only and very local (there are arguments for this besides simplifying the implementation but that's another matter). `each` takes a table as an argument which carries an index set. By the time we reach the lambda argument to `each` we've seen that table so we know the index set. But we don't learn the index set for the `for` until we see how the index is used. A general principle of the new inference system is that we never infer things about a variable based on how it's used. Type inference follows data flow. It's the UTI principle: type information shouldn't flow backwards.

But please keep the friction reports coming regardless. I want to at least be aware of the cost I'm imposing on users with these choices.

